PR #7, opened 2015-11-02
Make OGGM windows compatible

JohannesUIBK:
Up to one test (test_inversion_nofs) everything should work.

There is a problem with symlinks: the current solution with file copy is suboptimal.


fmaussion:
Thanks! The travis failures seem to be unrelated to the patch - probably an error due to the new version of matplotlib. I'll check it and come back to you

fmaussion:
Yep that seems to be the problem. We should also check if a matplotlib version mishmash is the problem we had with the windows plots

PR #16, opened 2015-12-11
Add the models workflow

fmaussion:
Still a lot to do, but basis is here.


fmaussion:
First shot with the flowline models. Quite unstable but a good start, documentation and testing are poor. Will get back to this after refactoring.

PR #23, opened 2016-02-01
New Flowmodel added

alexjarosch:
I have added several new things to OGGM:

1. My MUSCL-SuperBee model is now functional, based on my 2013 paper: A. H. Jarosch, C. G. Schoof, and F. S. Anslow, “Restoring mass conservation to shallow ice flow models over complex terrain,” The Cryosphere, vol. 7, iss. 1, pp. 229-240, 2013.

2. I have changed the FluxBasedModel to use the modern 2*Aglen/(N+2) instead of the fd parameter
3. I have changed the KarthausModel to use the modern 2*Aglen/(N+2) instead of the fd parameter
4. The test_constant_bed case includes now and compares to the MUSCL-SuperBee model as a benchmark.
5. I have added the test_constant_bed_cliff test case which demonstrates the effect of a bedrock cliff on the three flowline models. This test also highlights the necessity of a proper numerical scheme to deal with such eventualities to avoid spurious mass creation.

### DIFF
--- a/oggm/core/models/flowline.py
+++ b/oggm/core/models/flowline.py
@@ -288,7 +288,7 @@ class MixedFlowline(ModelFlowline):
 class FlowlineModel(object):
     """Interface to the actual model"""
 
-    def __init__(self, flowlines, mb_model=None, y0=0., fs=None, fd=None):
+    def __init__(self, flowlines, mb_model=None, y0=0., Aglen=None, fs=None, fd=None):
         """ Instanciate.
 
         Parameters
@@ -300,6 +300,7 @@ class FlowlineModel(object):
         """
 
         self.mb = mb_model
+        self.Aglen = Aglen
         self.fd = fd
         self.fs = fs
 
@@ -414,7 +415,7 @@ class FlowlineModel(object):
 class FluxBasedModel(FlowlineModel):
     """The actual model"""
 
-    def __init__(self, flowlines, mb_model=None, y0=0., fs=None, fd=None,
+    def __init__(self, flowlines, mb_model=None, y0=0., Aglen=None, fs=None, fd=None,
                        fixed_dt=None, min_dt=SEC_IN_DAY, max_dt=SEC_IN_MONTH):
 
         """ Instanciate.
@@ -427,7 +428,7 @@ class FluxBasedModel(FlowlineModel):
         #TODO: document properties
         """
         super(FluxBasedModel, self).__init__(flowlines, mb_model=mb_model,
-                                             y0=y0, fs=fs, fd=fd)
+                                             y0=y0, Aglen=Aglen, fs=fs, fd=fd)
         self.dt_warning = False
         if fixed_dt is not None:
             min_dt = fixed_dt
@@ -479,7 +480,7 @@ class FluxBasedModel(FlowlineModel):
 
             # Staggered velocity (Deformation + Sliding)
             rhogh = (RHO*G*slope_stag)**N
-            u_stag = (thick_stag**(N+1)) * self.fd * rhogh + \
+            u_stag = (thick_stag**(N+1)) * 2.*self.Aglen/(N+2.) * rhogh + \
                      (thick_stag**(N-1)) * self.fs * rhogh
 
             # Staggered section
@@ -546,7 +547,7 @@ class FluxBasedModel(FlowlineModel):
 class KarthausModel(FlowlineModel):
     """The actual model"""
 
-    def __init__(self, flowlines, mb_model=None, y0=0., fs=None, fd=None,
+    def __init__(self, flowlines, mb_model=None, y0=0., Aglen=None, fs=None, fd=None,
                  fixed_dt=None, min_dt=SEC_IN_DAY, max_dt=SEC_IN_MONTH):
 
         """ Instanciate.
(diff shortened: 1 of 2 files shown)


fmaussion:
Great contribution, thanks a lot! Do not worry about the tests not passing yet, I think I know where it comes from

oggm/core/models/flowline.py
```diff
@@ -300,6 +300,7 @@ def __init__(self, flowlines, mb_model=None, y0=0., fs=None, fd=None):
         """
 
         self.mb = mb_model
+        self.Aglen = Aglen
         self.fd = fd
```
fmaussion: To preserve backwards compatibility (at least for a while), setting `fd` could be equivalent to setting `Aglen`, and the `self.fd` property could be removed. Something like:

```python
if fd is not None:
    Aglen = (N+2.) / 2
```

What do you think?

oggm/core/models/flowline.py
```diff
@@ -623,6 +625,153 @@ def step(self, dt=SEC_IN_MONTH):
...
+            #explicit time stepping scheme
+            div_q = (D_up * (S[kp] - S[k])/dx - D_dn * (S[k] - S[km])/dx)/dx    # Eq. 36
+            S = S[k] + (m_dot + div_q)*dt_use                                   # Eq. 35
+            
+            S = np.maximum(S,B)                                                 # Eq. 7
+        
+        # Done with the loop, prepare output
+        NewIceThickness = S-B
+        
+        fl.section = NewIceThickness * width
```
fmaussion: Here you could set `fl.thick = NewIceThickness`

This would have the advantage that you don't have to assume a U-shaped glacier bed. If you have a look at the code for `ParabolicFlowline`, `VerticalWallFlowline`, and `TrapezoidalFlowline` (https://github.com/OGGM/oggm/blob/master/oggm/core/models/flowline.py#L89), you'll see that setting `fl.section` is equivalent to setting the thickness, which is the actual parameter defining the section. This is how the `FluxBasedModel` works, regardless of which shape the glacier bed has.

Note that in a way the `FluxBasedModel` physics are a bit wrong (in https://github.com/OGGM/oggm/blob/master/oggm/core/models/flowline.py#L476, the thickness is physically the same regardless of the bed shape). How "wrong" do you think that your maths above will be if the bed is e.g. parabolic? 

I hope that I've been clear, otherwise I'll try to elaborate more.

oggm/core/models/flowline.py
```diff
@@ -623,6 +625,153 @@ def step(self, dt=SEC_IN_MONTH):
...
+        dx = fl.dx_meter
+        width = fl.widths_m
+        
+        """ Switch to the notation from the MUSCL_1D example
+            This is useful to ensure that the MUSCL-SuperBee code
+            is working as it has been benchmarked many time"""
+         
+
+        # mass balance
+        m_dot = self.mb.get_mb(fl.surface_h, self.yr)
```
fmaussion: Shouldn't the mb be multiplied by the glacier width here?

oggm/core/models/flowline.py
alexjarosch: That should be fine, but we should clearly note in the code that this is for legacy backwards compatibility and make an effort to "switch" oggm to the modern notation.

oggm/core/models/flowline.py
alexjarosch: Sounds good, I'll set it to fl.thick = NewIceThickness

oggm/core/models/flowline.py
```diff
@@ -623,6 +630,154 @@ def step(self, dt=SEC_IN_MONTH):
...
```
alexjarosch: You asked last time round if mb should not be multiplied by glacier with here. I don't know. I took your Karthaus model as a template and there (line 593 above) you don't do that either. Please elaborate on your thought

oggm/core/models/flowline.py
```diff
@@ -300,8 +300,14 @@ def __init__(self, flowlines, mb_model=None, y0=0., fs=None, fd=None):
...
         self.mb = mb_model
-        self.fd = fd
+        # self.fd = fd
         self.fs = fs
+        # for current backwards compatibility I calculate fd from Aglen if fd is not given
+        # We should however streamline oggm to only use Aglen and get rid of fd
+        if fd is not None and Aglen is None:
+            Aglen = (N+2.) * fd / 2.
+            
+        self.Aglen = Aglen
```
alexjarosch: I have changed this here to calculate Aglen from fd only if fd exists but Aglen not, else we should use Aglen. This works for all the tests I ran, but some of them still return failed. The reason is that your original fd is set to quite a different value. 

In your TestIdealisedCases class you set ``self.fd = 1.9e-24`` and I set ``self.Aglen = 2.4e-24``. Converting the fd into the Aglen world you end up with your fd being equivalent to an Aglen of 4.75e-24, which is way softer ice than usually assumed. The perceived softer ice makes the benchmarks fail I guess.

fmaussion:
Perfect, thanks! The errors seem to be related to numerical issues: on my system I am able to reproduce the cliff plots but the asserts don't pass. This is not a problem, I'll take care of this. The non-responding tests is because of `do_plot=True`, which is also OK.

I'll merge this now, take care of my failing tests in a separate PR and come back to you soon about this width stuff, ok?

fmaussion:
OK, just fixed the tests here: https://github.com/OGGM/oggm/pull/23

I'll come back to you in the next couple of days. Thanks again!

alexjarosch:
Good stuff, looking forward to hear from you.

PR #25, opened 2016-02-09
Added simple script to start and control Amazon AWS EC2 instances

fso42:
Title says it all. Script is meant as a starting point. Documentation is in the file itself...

### DIFF
--- a/deployment/fabfile.py
+++ b/deployment/fabfile.py
@@ -1,0 +1,822 @@
+#!---------written by Felix Oesterle (FSO)-----------------
+#-DESCRIPTION:
+# This is based on fabfile from Raincloud Project (simplified)
+#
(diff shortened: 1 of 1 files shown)


fmaussion:
Thanks a lot! I can't comment much because I don't understand all the details, but I can see that it is a template on which we could build. The installation might be a bit more complicated in our case but @TimoRoth is already working on that [here](https://github.com/OGGM/OGGM-Anaconda).

I am not sure if we should do a new install for each run as you do, or if we should make an OGGM image. This will depend on the benchmarks I guess (if an Anaconda run is as fast as a compiled install, no need to make an image). 

For the rest, I'm afraid this goes beyond my actual competences and I'll have to delegate to Timo ;-)

PR #36, opened 2016-02-26
Worldwide DEM download

fmaussion:
Just to keep track -- not finished yet


fmaussion:
OK. Merging this now. Again, this induces backwards compatibility problems, because the config file has changed. The changes are however not too large:
- cfg.PATHS['srtm_file'] -> cfg.PATHS['dem_file']
- cfg.PATHS['histalp_file'] -> cfg.PATHS['climate_file'] 
- plotting routines now all follow the entity_task signature (one single argument -gdir- and the rest must be keywords)

PR #39, opened 2016-03-01
Check oggm-sample-data github commit for updates

### LINKED ISSUE #37: Re-download sample data when updated
OGGM will download some data files automatically from the oggm-sample-data repository:

https://github.com/OGGM/oggm-sample-data/archive/master.zip

The relevant code is [here](https://github.com/OGGM/oggm/blob/master/oggm/utils.py#L37) and [here](https://github.com/OGGM/oggm/blob/master/oggm/utils.py#L54). 

Currently there's no check to see of the server file has changed: this is bad but I don't know how to ask github if the file has changed.

Any idea on how to do this within python, @TimoRoth , @fso42 ? Any interest in submitting a PR? ;-)

TimoRoth:
I currently don't have a python2 environment to test with.
Works as expected on 3.4 for me.

It only checks for updates once every 30 minutes, to prevent it from doing a http request for each requested sample file.

This Fixes #37

### DIFF
--- a/oggm/utils.py
+++ b/oggm/utils.py
@@ -11,7 +11,8 @@ import six.moves.cPickle as pickle
 
 import pandas as pd
 from salem import lazy_property
-from six.moves.urllib.request import urlretrieve
+from six.moves.urllib.request import urlretrieve, urlopen
+from six.moves.urllib.error import HTTPError
 
 # Builtins
 import os
@@ -19,6 +20,8 @@ import gzip
 import shutil
 import zipfile
 import sys
+import json
+import time
 from collections import OrderedDict
 from functools import partial, wraps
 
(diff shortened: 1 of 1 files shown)


fmaussion:
Thanks a lot! It seems to work well on py3, but py2 is making problems. Quite strangely, conda on appveyor is also failing... Any idea why?

TimoRoth:
It's running into an error 403, which most likely means it hit GitHubs API rate limiting.
That's probably an issue on those Travis/AppVeyor machines, as the limit is 60 requests per IP per hour.
I'll add an additional check, that causes a re-download on those errors, they should rarely be an issue on user-machines, and travis/appveyor have to download it anyway.

fmaussion:
This is perfect thanks a lot!

PR #43, opened 2016-03-06
fabfile now py2/p3 compatible

fmaussion:
I know, fabric is not py3 compatible. But I am being pedantic and want to have parentheses around my prints (also, PyCharm is getting crazy red with py2 code).

Will fabric support python3 one day? 
https://github.com/fabric/fabric/issues/1050


fmaussion:
Before I merge this: any objection @fso42 ? ;-)

PR #45, opened 2016-03-06
Ensure proper download on py2

fmaussion:
There was an issue with the with context manager in py2


fmaussion:
Actually I have decided to wait for https://github.com/OGGM/oggm/pull/36 to be merged. Right now this is a mess because sample-data has changed but the code has not.

Note for the future: make smaller, more consistent PRs...

Issue #41, opened 2016-03-05
Better ssh keynames in deployment


fso42:
Currently the default variable 
keyn 
is set to oggm. Leading to problems if a new user just leaves it as is. If somebody else has oggm already, the keypair is not recognized as being new, so the 2nd user does not get a new keypair. Maybe include the local hostname as default...

PR #48, opened 2016-03-07
Update fabfile

TimoRoth:
Fixes issue #41 by generating a unique identifier for the keyname.

While at it, I added some functions to terminate individual instances/volumes, so we don't have to kill each others instances via cloud_terminate, or leave them running indefinitely.

Also bumped the Ubuntu AMIs to include some of the latest security fixes by default.

After this is merged, connecting to already running instances via fab connect won't work anymore, but manually invoking ssh with the old key file still works(ssh -i ~/.ssh/eu-west-1_oggm.pem ec2-52-50-61-172.eu-west-1.compute.amazonaws.com)

### DIFF
--- a/deployment/fabfile.py
+++ b/deployment/fabfile.py
@@ -288,6 +288,45 @@ def run_workflow(cn=def_cn,avz=def_default_avz):
 
     return target_file
 
+def print_instance(inst):
+    if inst.state != 'terminated':
+        cu_time = datetime.datetime.utcnow()
+        it =  datetime.datetime.strptime(inst.launch_time,'%Y-%m-%dT%H:%M:%S.000Z')
+    else:
+        try:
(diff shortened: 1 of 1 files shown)


fmaussion:
Thanks! Could you add this new info in the doc also? (i.e.: how to connect, and in particular, how to NOT kill each others instances)

fmaussion:
mh. I guess this will have to be done in a separate PR! :D

TimoRoth:
In general. nothing about how to connect changed, it's just that already running instances still use the old key, while the updated script will generate and use new ones. The issue will be gone once all currently running instances have been restarted.

fmaussion:
So will `fab cloud_terminate` still terminate the instances I created?

TimoRoth:
Yes, it will tear down everything except for the user volumes mounted at /work.
I'll add some owner tags which are printed in the instance/volume listing, so it's easier to see which instance belongs to whom.

PR #49, opened 2016-03-07
Update fabfile documentation

TimoRoth:
Updates the documentation at the top of fabfile.py for the new less destructive termination tasks.

To make it easier to identify which instances are your own, I added the ability to add a user identifier, which defaults to the local username is left unset.

Also I just noticed that the AMI listing function was filtering for the wrong type, so depending on the instance type, it might have failed to startup because of the wrong images.


fmaussion:
Perfect, thanks! Will try it out tomorrow

PR #52, opened 2016-03-10
some flowline examples

fmaussion:


fmaussion:
Merging, as the errors are fixed by the parallel PR https://github.com/OGGM/oggm/pull/36

Issue #55, opened 2016-03-13
More wordy fab connect


fmaussion:
It would be nice to have more info on the `fab connect` command? something similar to `fab terminate`, with owner name and stuff, if possible of course!

PR #56, opened 2016-03-13
AWS install packages for OGGM

fmaussion:
`node_install` now has options to install the packages if needed. After install the AWS instance can run oggm without error, in python3. 

pip install takes 10 to 20 minutes, I think we can do better if we use wheels

@fso42 , @TimoRoth for you to review and merge if found useful

### DIFF
--- a/deployment/fabfile.py
+++ b/deployment/fabfile.py
@@ -88,12 +88,11 @@ env.user = 'root'
 def_cn = 'AWS'
 
 # Change to a string identifying yourself
-# Defaults to your system username if set to None
 user_identifier = None
 
 # FSO--- ssh and credentials setup
 # FSO---the name of the amazon keypair (will be created if it does not exist)
-keyn='yourname_oggm'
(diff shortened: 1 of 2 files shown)


fmaussion:
I alsoslightly changed the defaults to force a username. This makes it less easy to do wrong manips (as I just did)

TimoRoth:
First, I would setup the VirtualEnv in /work, if it's available.
That way there is no need to run the 20 minute Virtualenv-Setup every time.

Also, I'd prefer to have the setup sequence in a separate task, so it can be run individually on machines that still need it.

TimoRoth:
If you don't mind, i can do the necessary changes.

fmaussion:
Perfect! Yes, please, modify it and make it better! I'll merge this one so that you can go on from that

fmaussion:
Ha, and also if you have time could you look at this one? https://github.com/OGGM/oggm/issues/55

PR #57, opened 2016-03-13
Finalize ITMIX

fmaussion:
No finished yet - we keep track


fmaussion:
False alarm. we do again later

PR #60, opened 2016-03-14
Fabfile Updates

TimoRoth:
Moves instance setup into separate task(s) and updates doc accordingly.
Also refactors some common tasks into reusable functions.

The install_env script is re-purposed into a generic install script that can be used in any environment where pip is available.


fmaussion:
Great, thanks! Is the install_env script still necessary btw?

BtbN:
I left it in because it could be useful in case someone wants to setup a virtualenv outside of AWS, it's otherwise unused for now.

TimoRoth:
I left it in because it could be useful in case someone wants to setup a virtualenv outside of AWS, it's otherwise unused for now.

fmaussion:
Yes good idea. Merging this, thanks!

PR #62, opened 2016-03-17
Retry downloading srtm files on error 500

TimoRoth:
This keeps failing CI builds, because sometimes the server randomly dies and returns error 503 for a while.
So with this change, downloading will be retried up to 5 times, with a 10 second delay between each attempt.


fmaussion:
Thanks! In an upcoming PR (with even more automated download), I might make the download tests optional. In a second step it might be good to put all this download logic in a separate module and make it more robust. I'll might come back to you for this!

PR #63, opened 2016-03-18
Fix and speed up travis builds

TimoRoth:
The python 2.7 builds were failing because the pip wheel downloaded and compiled a local numpy version, which all the other pip packages were using.
But because of the installed system numpy, pip install never installed it, so the other packages failed because of the incompatible ABI of the system numpy.

In turn I also made python 2.7 required for Travis to pass again, so no python2 incompatible code sneaks in.

This also gets rid of xvfb, and instead switches matplotlib to a headless backend.
Speeds up the build by a few minutes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,8 +33,6 @@ matrix:
   allow_failures:
     - python: 2.7
 before_install:
-  - "export DISPLAY=:99.0"
-  - "sh -e /etc/init.d/xvfb start"
   - pip install -U pip
   - pip install wheel
 install:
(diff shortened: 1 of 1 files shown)


fmaussion:
Thanks! I just restarted the py2.7 build to see if it now completes.

.travis.yml
```diff
@@ -30,16 +30,12 @@ matrix:
...
-    - python: 2.7
 before_install:
-  - "export DISPLAY=:99.0"
-  - "sh -e /etc/init.d/xvfb start"
   - pip install -U pip
   - pip install wheel
 install:
   - travis_wait pip wheel -r ci/requirements.txt
-  - travis_wait pip install -r ci/requirements.txt
+  - travis_wait pip install --upgrade --force-reinstall -r ci/requirements.txt
```
fmaussion: Isn't the `--force-reinstall` always reinstalling the builds, thus not taking advantages of the wheels?

BtbN: It still uses the wheels, just forces installation even if the same version is already installed.

TimoRoth: It still uses the wheels, just forces installation even if the same version is already installed(by the system package manager in this case).

fmaussion:
Perfect, thanks!

PR #75, opened 2016-04-19
Download corrected SRTM files and spelling 

JohannesUIBK:
Automatic filled SRTM download from viewfinderpanoramas.org should work now.

### DIFF
--- a/oggm/utils.py
+++ b/oggm/utils.py
@@ -128,48 +128,6 @@ def _download_oggm_files():
     return out
 
 
-def _download_srtm_file(zone):
-    """Checks if the srtm data is in the directory and if not, download it.
-    """
-
-    odir = os.path.join(cfg.PATHS['topo_dir'], 'srtm')
-    if not os.path.exists(odir):
-        os.makedirs(odir)
-    ofile = os.path.join(odir, 'srtm_' + zone + '.zip')
-    ifile = 'http://srtm.csi.cgiar.org/SRT-ZIP/SRTM_V41/SRTM_Data_GeoTiff' \
-            '/srtm_' + zone + '.zip'
-    if not os.path.exists(ofile):
-        retry_counter = 0
-        retry_max = 5
-        while True:
-            # Try to download
-            try:
-                retry_counter += 1
-                urlretrieve(ifile, ofile)
-                with zipfile.ZipFile(ofile) as zf:
-                    zf.extractall(odir)
-                break
-            except HTTPError as err:
-                # This works well for py3
-                if err.code == 404:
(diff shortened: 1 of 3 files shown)


fmaussion:
Thanks for this PR! 

I think there was a small misunderstanding however. The SRTM files are quite fine, it is mostly the ASTER zones (> 60° N and < 60°S) which we would like to replace with the corrected files. 

In my idea, all existing download functions will remain the same, but we will have a new `_download_dem3`[1] which could jump in in place of `_download_aster` when possible. It would allow to choose between all the possibilities. 

[1]: I don't know what the best name of this new dataset would be?

oggm/utils.py
```diff
@@ -218,12 +245,17 @@ def _download_alternate_topo_file(fname):
...
+        'gimpdem_90m.zip',
+        'iceland.zip',
+        'northcanada.zip',
+        'svalbard.zip',
+        'AN1.zip',
+        'AN2.zip',
+        'AN3.zip',
+        'AN4.zip',
+        'AN5.zip'
     ]:
```
fmaussion: What do AN1 to AN5 stand for?

fmaussion: Maybe we should get rid of this test anyway, because its a maintenance burden (for example, franz joseph is missing and such)

fmaussion:
It would also be easier to split the spell-checking and the DEM stuff into two separate PRs, if you dont mind?

oggm/utils.py
JohannesUIBK: Ah okay!
I thought it was easier to have all SRTM DEMs from one source (as they should be the same anyway!?)
I will change things the way you have suggested now over the next days.
I think a good name for the dataset was `_download_dem3_filled` if you are okay with it.

AN1 to AN5 stand for the Antarctica tiles 1-5.

Last but not least: yes, if you talk about the `fzipname` test, we should definitely check whether we can implement special conditions for special lats/lons in the `_download_dem3_filled` download function. That's what I believe would be the easiest to get the files nevertheless.

JohannesUIBK:
Ah okay, testing failed, because I forgot to delete an "old" line in test_utils.py....I will do it or fall back to the original version and include your suggested changes.

Okay, we can split the PR as the spelling is hopefully correct and you can accept this PR already.

fmaussion:
Yes, lets do the spell check first (because it might also fail, for different reasons)

oggm/utils.py
fmaussion: Yes, maybe it will replace SRTM also, but this requires careful checking. And it's always good to keep all possible options open

PR #79, opened 2016-04-25
new viewfinderpanorama DEM files

JohannesUIBK:
Not finished yet

### DIFF
--- a/oggm/tests/test_utils.py
+++ b/oggm/tests/test_utils.py
@@ -103,7 +103,11 @@ class TestDataFiles(unittest.TestCase):
                         '01-15': [-180., -91., -90, -60.],  # Antarctica tiles as UTM zones, FILES ARE LARGE!!!!!
                         '16-30': [-91., -1., -90., -60.],
                         '31-45': [-1., 89., -90., -60.],
-                        '46-60': [89., 189., -90., -60.]}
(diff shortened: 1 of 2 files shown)


fmaussion:
@JohannesUIBK , is this last commit now taking care of what we discussed last time?

JohannesUIBK:
Indeed it is, I was just not able to test some things due to this GDAL/rasterio issue. But as I have seen now the dependcies should be fine, right? So maybe I can test this evening and then you can merge the PR! I will get back to you soon!

JohannesUIBK:
Okay so I tried now to reinstall all Python packages from the conda forge channel, but I still cannot import rasterio properly somehow...consequently, I'm not really able to test if all the things work.
Does it work for you now?

fmaussion:
OK. If you can reproduce the problem easily in a fresh environment it might be worth to report it to `conda-forge`, but no hurry (windows sucks anyway ;-) ). On a side note: we have no such problems on appveyor, which is also using a windows machine.

I will have to clone your branch to test it, I will have a look by the end of the week. Thanks!

fmaussion:
OK, merging this now since tests are OK, will have a look at the details later today or tomorrow. Thanks @JohannesUIBK !

JohannesUIBK:
Okay and thanks for both the tip concerning the report and for the merge!

PR #82, opened 2016-05-02
Fabfile Updates

TimoRoth:
New Ubuntu AMIs, fixes issues with keypairs not being saves per-region and adds user volume to fstab so it gets mounted after a reboot.


fmaussion:
Great, thanks a lot! Should I merge now or are other changes coming up?

TimoRoth:
That's it for now, unless i forgot an issue you mentioned?

fmaussion:
I don't think so! Merging now

PR #83, opened 2016-05-02
Report download progress

TimoRoth:
Makes long lasting downloads of data file a little less annoying.

### DIFF
--- a/docs/installing-oggm.rst
+++ b/docs/installing-oggm.rst
@@ -61,6 +61,9 @@ Other libraries:
     - `cleo <https://github.com/fmaussion/cleo>`_
     - `motionless (py3) <https://github.com/fmaussion/motionless>`_
 
+Optional:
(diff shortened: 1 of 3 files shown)


fmaussion:
Great! Could you also update the docs to take into account this new dependency?
 https://github.com/OGGM/oggm/blob/master/docs/installing-oggm.rst 
Is the progressbar graphical or text based?

TimoRoth:
It's a text based wget-style progressbar:
53% |####################                                               |  20.4 MiB   2.5 MiB/s ETA:  0:00:07

TimoRoth:
conda does not seem to have a package for progressbar2, only the old python2-only progressbar.
Updated the docs accordingly, optionally installing it via pip if desired.

fmaussion:
Perfect! I am starting to think that `utils.py` would benefit from a little bit of refactoring, we'll keep this in mind for later

PR #84, opened 2016-05-04
Fix ProgressBar exception on unknown length

TimoRoth:
urlretrieve returns -1 as total size if the size is unknown for some reason.
ProgressBar does not expect this and throws an exception, so translate values <=0 to UnknownLength.


fmaussion:
Merging, thanks!

Maybe we should put all this download stuff in a separate module and make a test environment which would run once per build (so that the other tests, which are long enough, are run in paralell)

TimoRoth:
We could also transform the utils.py file to a directory, with separate files for each category of utility functions.

PR #89, opened 2016-05-10
Workaround bug in some progressbar versions

TimoRoth:
Some progressbar versions miscalculate the progress, causing ValueError from math.log on finish.
As errors from progressbar are not fatal for us, just ignore them.


fmaussion:
Thanks!

Issue #87, opened 2016-05-04
Shorten test duration time


fmaussion:
This can be achieved by running different tests in different travis environments. 

The appveyor tests should be concentrated on the very core functions, so that only the usual conda package problems are caught, but not necessarily the heavy numerical computation stuffs...

fmaussion:
partly addressed by https://github.com/OGGM/oggm/pull/107

PR #101, opened 2016-05-18
fabfile: Switch to Ubuntu 16.04

TimoRoth:
Also delay the final reboot on software install a bit, so ssh can cleanly disconnect first.

### DIFF
--- a/deployment/fabfile.py
+++ b/deployment/fabfile.py
@@ -563,7 +563,7 @@ def install_node_pip(nn='', inst=None):
     pip install matplotlib &&
     pip install gdal==1.11.2 --install-option="build_ext" --install-option="--include-dirs=/usr/include/gdal" &&
     pip install fiona --install-option="build_ext" --install-option="--include-dirs=/usr/include/gdal" &&
-    pip install pyproj rasterio Pillow geopandas netcdf4 scikit-image configobj joblib xarray nose &&
+    pip install pyproj rasterio Pillow geopandas netcdf4 scikit-image configobj joblib xarray nose progressbar2 &&
     pip install git+https://github.com/fmaussion/motionless.git &&
     pip install git+https://github.com/fmaussion/salem.git &&
     pip install git+https://github.com/fmaussion/cleo.git &&
--- a/deployment/fabfile.py
+++ b/deployment/fabfile.py
@@ -119,8 +119,8 @@ def_default_requesttype = 'spot'
 
 # FSO--- the AMI to use
 def_ami = dict()
-def_ami['eu-west-1'] = 'ami-6177f712' #eu Ubuntu 14.04 LTS
-def_ami['us-east-1'] = 'ami-2b594f41' #us Ubuntu 14.04 LTS
+def_ami['eu-west-1'] = 'ami-7a138709' #eu Ubuntu 16.04 LTS
+def_ami['us-east-1'] = 'ami-13be557e' #us Ubuntu 16.04 LTS
 
 # Size of the rootfs of created instances
 rootfs_size_gb = 50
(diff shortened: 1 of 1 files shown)


deployment/fabfile.py
```diff
@@ -563,7 +563,7 @@ def install_node_pip(nn='', inst=None):
     pip install matplotlib &&
     pip install gdal==1.10.0 --install-option="build_ext" --install-option="--include-dirs=/usr/include/gdal" &&
```
fmaussion: This update is going to have some implications, as Ubuntu 16.04 ships with GDAL 1.11.3: https://launchpad.net/ubuntu/xenial/+source/gdal 

I guess (but I can't be 100% sure) that you'll have to adapt here

fmaussion:
Could you also copy/paste the output of `nosetests .` here so that I can have a look?

Thanks!

TimoRoth:
The python GDAL package only needs a minimum version of the gdal library:
https://pypi.python.org/pypi/GDAL/1.10.0
For 1.10.0 this is 1.7.0 or greater, so 1.11.3 from 16.04 is fine.

We could aim for a more recent version of the python package on 16.04, but the last time i tried, that failed a lot of tests.

```
(oggm_env) ubuntu@ip-172-31-18-171:/work/ubuntu/oggm$ nosetests . |& tee ~/nosetest-output.txt
Trying `CDLL(libgeos_c.so.1)`
Library path: 'libgeos_c.so.1'
DLL: <CDLL 'libgeos_c.so.1', handle 24f5ed0 at 0x7f5861c49898>
Trying `CDLL(libc.so.6)`
Library path: 'libc.so.6'
DLL: <CDLL 'libc.so.6', handle 7f5863f729b0 at 0x7f58053a55c0>
...............S.SS............../work/ubuntu/oggm/oggm/core/models/massbalance.py:111: DeprecationWarning: BackwardsMassBalanceModel is deprecated
  warnings.warn(DeprecationWarning('BackwardsMassBalanceModel'
/work/ubuntu/oggm/oggm/core/models/massbalance.py:111: DeprecationWarning: BackwardsMassBalanceModel is deprecated
  warnings.warn(DeprecationWarning('BackwardsMassBalanceModel'
.......S......................SSS..SSS.SSSSS.SSS
----------------------------------------------------------------------
Ran 81 tests in 184.499s

OK (SKIP=18)
```

fmaussion:
Yes, because recent versions of GDAL do things a bit differently and some tests fail because of numerical issues. Could you also run the tests after setting the following environment variable:
```
export $OGGM_SLOW_TESTS=True
```
The relevant part of the code is here: https://github.com/OGGM/oggm/blob/master/oggm/tests/__init__.py

TimoRoth:
GDAL from 2.0 onwards fail to build entirely. But with 1.11.2 from pypi things still seem to work fine.

Running the slow tests with GDAL-1.11.2 now, but staying on 1.10.0 should not be an issue, even on 16.04 which has more recent system libraries.

If the tests succeed, should I update the install routines for 1.11.2?

fmaussion:
This is a good question. Currently, some tests are skipped when on GDAL 1.11:

```python
HAS_NEW_GDAL = False
if osgeo.gdal.__version__ >= '1.11':
    HAS_NEW_GDAL = True
```
I think that basically, we should indeed opt for the more recent version. But I'll have to adapt the tests that fail on GDAL 1.11 also. It would be good to have GDAL 1.11 on travis also. Do you think it is possible? Maybe with mixing conda/nonconda test environments?

TimoRoth:
Travis has the option to use Docker-Containers, so we can setup whatever environment we like on there, even a full ubuntu 16.04.

fmaussion:
I you think that it's worth it, why not. But if you don't have the time it's not that urgent (I just checked, and the tests that are concerned are not so many).

It also depends if the docker stuff makes the build even slowier than it is right now. For me 20-25 minutes for the tests to run is already too long, and I am thinking about alternative testing strategies to make the tests run faster: https://github.com/OGGM/oggm/issues/87

TimoRoth:
I'll setup a test for it. I think the Docker-Builds run on slightly faster hardware, so they might even end up being a bit faster.

fmaussion:
Thanks!

PR #104, opened 2016-05-20
Multiprocessing fixes and improvements

TimoRoth:
Create a global mp.Lock() which is then used to guard all download operations, so they don't happen in parallel.

In turn, also passes cfg.PARAMS to all newly created processes.


fmaussion:
Merged. Thanks!

PR #112, opened 2016-06-29
Make execute_entity_task forward kwargs

TimoRoth:
Without this it's impossible to call for example distribute_thickness, which needs the how parameter to be set to do anything.

### DIFF
--- a/oggm/workflow.py
+++ b/oggm/workflow.py
@@ -45,7 +45,20 @@ def _init_pool():
     return mp.Pool(cfg.PARAMS['mp_processes'], initializer=_init_pool_globals, initargs=(download_lock, cfg_contents))
 
 
-def execute_entity_task(task, gdirs):
+class _pickle_copier(object):
+    """Pickleable alternative to functools.partial,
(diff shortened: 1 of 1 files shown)


fmaussion:
:+1: this is great. I thought I had implemented this already but obviously not!

I 'm not sure if you need your own decorator for this though, [functools.partial](https://docs.python.org/3.5/library/functools.html#functools.partial) should do the trick?

The failing tests are unrelated.

TimoRoth:
functools.partial is only pickleable in python3, so in python2 passing it to map fails.

fmaussion:
OK. Can you add a comment line in the `_pickle_copier` class to document this so we can remember in the future?

I am starting to think that we should drop python2 support as a whole (http://python3statement.github.io/ )

TimoRoth:
Comment added.

Yes, dropping python2 support would definitely simplify the code in a lot of places, and speed up builds.

fmaussion:
Thanks! We will drop python 2 support soon (shouldve done it in the first place already :( )

PR #114, opened 2016-07-04
Use Docker for Travis Tests

TimoRoth:
This runs the Travis-Tests via Docker, on Ubuntu 16.04 with Python 3.5 and libgdal 1.11.3

It's also quite a bit faster.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,61 +1,41 @@
-language: python
-sudo: false
+sudo: required
+dist: trusty
+services:
+  - docker
+
 notifications:
   email: false
+
 branches:  # blacklist
   except:
     - appveyor
     - docs
-cache:
-  directories:
-    - ~/.cache/pip
-env:
-  global:
-    - PIP_WHEEL_DIR=$HOME/.cache/pip/wheels
-    - PIP_FIND_LINKS=file://$HOME/.cache/pip/wheels
-addons:
-  apt:
-    packages:
-      # copied those from scipy
-      - libatlas-dev
-      - libatlas-base-dev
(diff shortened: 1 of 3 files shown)


TimoRoth:
It should be noted that this drops Python 2 tests from Travis.
They would be possible to add, but so far I did not do that.

fmaussion:
That's fine for me! I just restarted the appveyor build to see what's going on there.

No need to test for py2 since appveyor is doing it. I will do the transition later this summer, but before that I have to find out which python is used by readthedocs to build our docs. Should be no big deal to go for py3 there also

fmaussion:
aha, maybe we are on py3 already: https://github.com/OGGM/oggm/blob/master/readthedocs.yml#L3 I'll test this

fmaussion:
Something seems wrong in the appveyor python3 test. Do you have an idea why?

TimoRoth:
There seems to be some conflict between the packages we need, but conda isn't exactly verbose about what's going wrong.
I'll test that locally and see if there is a workaround.

TimoRoth:
Ok, the issue is geopandas and some other packages not having windows binaries built for python3.5, only for 3.4.
So using 3.4 for conda should workaround that issue for now.

TimoRoth:
Could you cancel the second latest appveyor build? I forgot the --yes option and it's stuck now.

fmaussion:
Just canceled it

TimoRoth:
I hate Windows batch scripts, but I think I found a solution that works now.

fmaussion:
Great! Merging now

PR #116, opened 2016-07-21
added a decorator for plotting maps

fmaussion:
It reduces boilerplate code a little bit, and allows to change options of all plots at the same time


fmaussion:
Merging this despite of failing tests - it is a chicken and egg problem with docker

PR #118, opened 2016-08-04
Use untested_base for Travis-Tests

TimoRoth:
This avoids strange chicken/egg problems preventing Docker-Updates and Tests on Travis.


fmaussion:
Thanks!

PR #120, opened 2016-09-09
Work on glacier calving (WIP)

bearecinos:


oggm/core/preprocessing/inversion.py
```diff
@@ -171,25 +171,25 @@ def invert_parabolic_bed(gdir, glen_a=cfg.A, fs=0., write=True):
...
+            # pno = np.where(ratio > max_ratio)
+            # if len(pno[0]) > 0:
+            #     ratio[pno] = np.NaN
+            #     ratio = utils.interp_nans(ratio, default=max_ratio)
+            #     out_thick = w * ratio
 
             # TODO: last thicknesses can be noisy sometimes: interpolate?
             if cl['is_last']:
                 out_thick[-4:-1] = np.NaN
                 out_thick = utils.interp_nans(out_thick)
```
fmaussion: I think this should be commented, too. This interpolation is not meaningful for calving glaciers either.

oggm/core/preprocessing/inversion.py
```diff
@@ -171,25 +171,25 @@ def invert_parabolic_bed(gdir, glen_a=cfg.A, fs=0., write=True):
...
-            if len(pno[0]) > 0:
-                out_shape[pno] = np.NaN
-                out_shape = utils.interp_nans(out_shape, default=max_shape)
-                out_thick = (out_shape * w**2) / 4
+            # out_shape = (4*out_thick)/(w**2)
+            # pno = np.where(out_shape > max_shape)
+            # if len(pno[0]) > 0:
+            #     out_shape[pno] = np.NaN
+            #     out_shape = utils.interp_nans(out_shape, default=max_shape)
+            #     out_thick = (out_shape * w**2) / 4
```
fmaussion: In order to keep the things as they were before, we could put all these cosmetic interpolations behind an ``if gdir.glacier_type == 'Land-terminating':``

oggm/sandbox/run_alaska.py
```diff
@@ -127,65 +130,237 @@
...
+    # We take the middle of this area
+    ind_term = ind[np.round(len(ind) / 2.).astype(np.int)]
+
+    # Getting the termminus altitude
+    t_altitude = zoutline[ind_term]
+
+    # We read the model output, of the last pixel of the inversion
     cl = gdir.read_pickle('inversion_output', div_id=1)[-1]
     width = cl['width'][-1]
     thick = cl['volume'][-1] / cl['dx'] / width
```
fmaussion: isn't there a ``'thick'`` value available already in ``cl``?

oggm/sandbox/run_alaska.py
```diff
@@ -127,65 +130,237 @@
...
+    # indices of where in the outline the altitude is lower than the qth
+    # percentile and lower than 100m higher, than the minimum altitude
+    ind = np.where((zoutline < plow) & (zoutline < (mini + deltah)))[0]
+
+    # We take the middle of this area
+    ind_term = ind[np.round(len(ind) / 2.).astype(np.int)]
+
+    # Getting the termminus altitude
+    t_altitude = zoutline[ind_term]
+
```
fmaussion: Here, instead of recomputing what we already did in ``core.centerlines.compute_centerlines``, you could get the altitude from ``cl['hgt']`` I guess. You can check it first, and see if it is the same as what you compute.

oggm/cfg.py
```diff
@@ -182,6 +182,14 @@ def doc_str(self, key):
...
 BASENAMES['past_model'] = ('past_model.nc', _doc)
 
+_doc = 'Calving flux values'
+BASENAMES['calving_fluxes'] = ('calving_fluxes.pkl', _doc)
+
+_doc = 'Water depth'
+BASENAMES['water_depth'] = ('water_depth.pkl', _doc)
+
+_doc = 'Thickness at the calving front'
+BASENAMES['H_ice'] = ('H_ice.pkl', _doc)
```
fmaussion: Maybe it would be easier to put all these data in one single dictionary, leaving only one entry for BASENAMES

oggm/core/preprocessing/inversion.py
```diff
@@ -170,26 +170,27 @@ def invert_parabolic_bed(gdir, glen_a=cfg.A, fs=0., write=True):
...
-
-            # Check for the shape parameter (should not be too large)
-            out_shape = (4*out_thick)/(w**2)
-            pno = np.where(out_shape > max_shape)
-            if len(pno[0]) > 0:
-                out_shape[pno] = np.NaN
-                out_shape = utils.interp_nans(out_shape, default=max_shape)
-                out_thick = (out_shape * w**2) / 4
+            if gdir.glacier_type == 'Land-terminating':
+                # Check for thick to width ratio (should ne be too large)
```
fmaussion: Sorry what you actually want here is is:
```python
if gdir.terminus_type == 'Land-terminating':
```
I think this is the reason for the failing tests

oggm/graphics.py
```diff
@@ -390,7 +390,7 @@ def plot_distributed_thickness(gdir, ax=None, cleomap=None, how=None):
             cleomap.set_geometry(l, crs=crs, color='black', linewidth=0.5)
 
     cleomap.set_cmap(plt.get_cmap('viridis'))
-    cleomap.set_plot_params(nlevels=256)
+    cleomap.set_plot_params(nlevels=256, vmax=1100)
     cleomap.set_data(thick)
```
fmaussion: For this PR it would be good to remove this, but I will make the plotting functions more flexible in a separate PR: you should be able to control plotting levels from outside the function

oggm/cfg.py
fmaussion: In order to avoid to have too many entries in ``BASENAMES``, I suggest to define only one entry here:

```python
+_doc = 'Dict of calving flux values, water depth, and thickness at the calving front'
+BASENAMES['calving'] = ('calving.pkl', _doc)
```
Unless there is a good reason to have three entries (e.g. the data is written and read from two different tasks).

See my comment below for how this translates in your code.

oggm/sandbox/run_alaska.py
```diff
@@ -127,65 +130,201 @@
...
+#We write everything
+gdir.write_pickle(data_calving, 'calving_fluxes', div_id=1)
+gdir.write_pickle(w_depth, 'water_depth', div_id=1)
+gdir.write_pickle(thick, 'H_ice', div_id=1)
+
+#We read everything
+all_calving_data = gdir.read_pickle('calving_fluxes', div_id=1)
+all_data_depth = gdir.read_pickle('water_depth', div_id=1)
+all_data_H_i = gdir.read_pickle('H_ice', div_id=1)
+
```
fmaussion: If you change the entry in `cfg.BASENAMES`, these six lines would translate into (not tested):

```python
#We write everything
calvout = {'calving_fluxes':data_calving, 'water_depth':w_depth, 'h_ice':thick}
gdir.write_pickle(calvout, 'calving')

#We read everything
calin = gdir.read_pickle('calving')
all_calving_data = calin['calving_fluxes']
all_data_depth = calin['water_depth']
all_data_H_i = calin['H_ice']
```
Note that I also removed the ``div_id=1`` keyword, as I hope that calving glaciers won't need any divide.

fmaussion:
Thanks Bea, we are on the right track! My suggestion is that you implement the changes I suggested if they work for you, and then we are ready to merge. 

The current solution with having your code in the `run_alaska.py` file is very good as long as you are trying things out. For your task to be implemented in the core oggm it will need tests. But this is not urgent for now.

Does that sound reasonable?

fmaussion:
Thanks!

Issue #86, opened 2016-05-04
Add test for downloads?


fmaussion:
Most of the tests which require a download of large files have a `test_download` decorator which is defined here: https://github.com/OGGM/oggm/blob/master/oggm/tests/__init__.py#L55

This was done so that one can tests the downloads locally (by setting the `RUN_DOWNLOAD_TESTS` variable to True, or with an environment variable). The tests are not run on travis, because downloads are unstable and because I didn't want to overwhelm the data providers with countless requests from Travis.

Maybe some tests could be run from time to time, maybe at random times or something. Thoughts?

fmaussion:
After https://github.com/OGGM/oggm/pull/146, there is plenty of room for improvement:
- a lot of code is duplicated
- the download logic could be tested by making the methods return a www path instead of actually downloading the file
- the downloads are still not 100% robust. A ``robust_download`` function taking care of this logic would be extremely useful (see for example: https://github.com/OGGM/oggm/blob/master/oggm/utils.py#L312-L339)

@TimoRoth let me know if you have interest and time to have a look at these in the upcoming weeks

TimoRoth:
The downloaded data could just be cached on Travis, if it's not too large, that should also make it fairly reliable.
Will look into it.

fmaussion:
Thanks a lot, this is very helpful! But again, I still think that all get_* routines should return a www path, and that one function should take care of the actual download, to improve the testing and to avoid duplicated code.

If you have time it would be very cool to add also this option that all data downloaded externally could be stored on Amazon S3 for later use.  (that is, we download SRTM data, CRU data, and DEM3 data once and only once from their - slower - servers )

fmaussion:
This has been partly addressed by https://github.com/OGGM/oggm/pull/284

Issue #69, opened 2016-03-29
Add more alternate DEM files


fmaussion:
ASTER data can be crap, and people did the valuable effort to correct them for many regions of the globe: http://viewfinderpanoramas.org/dem3.html

OGGM already has three region specific DEMs here: https://github.com/OGGM/oggm/blob/master/oggm/utils.py#L756

It would be cool to have more! @JohannesUIBK , if you still have a few hours left on your contract this would be a possible very useful task.

JohannesUIBK:
Sure, I can have a look at it!
Thanks to your efforts there will enough time over the next weeks!

fmaussion:
> there will enough time over the next weeks!

Cool! then maybe it will be worth it ti have a look at the always growing list of [issues](https://github.com/OGGM/oggm/issues) ;-)

fmaussion:
Partly addressed by https://github.com/OGGM/oggm/pull/79 but there is still a bug so I had to revert part of it here: https://github.com/OGGM/oggm/pull/96

I should get back to this soon

fmaussion:
@all : this task is still not finished. The relevant code is here: https://github.com/OGGM/oggm/blob/master/oggm/utils.py#L1133-L1155 (commented because not working...)

fmaussion:
closed via https://github.com/OGGM/oggm/pull/146

Issue #58, opened 2016-03-13
Improve data download


fmaussion:
Since ITMIX, OGGM can theoretically work everywhere on the globe without effort. 

Some of the data (e.g. WGMS, GlaThiDa, CRU Clim) is stored on the [oggm-data](https://github.com/OGGM/oggm-sample-data) repository. This is downloaded at run time, and this is fast.

Some larger files are downloaded from their original repo, when public:
- [CRU](https://github.com/OGGM/oggm/blob/master/oggm/utils.py#L677)
- [SRTM](https://github.com/OGGM/oggm/blob/master/oggm/utils.py#L131)
- [RGI](https://github.com/OGGM/oggm/blob/master/oggm/utils.py#L635)

Some large files are not really public (ASTER -- TODO: what is their data policy by the way?) and are stored on S3:
- [ASTER](https://github.com/OGGM/oggm/blob/master/oggm/utils.py#L173)
- [alternate topo files](https://github.com/OGGM/oggm/blob/master/oggm/utils.py#L208): some files I got from http://viewfinderpanoramas.org/dem3.html, but which I had to pre process with GIS (mosaics + convert to geotiff) 

Note that SRTM and ASTER tiles are downloaded only if necessary, i.e. just for the region of interest.

It would be great to formalize all this logic in a separate module (maybe `download.py`?). When a file needs to be downloaded (eg CRU, SRTM...), it can take a long time. It would be good to inform the user with a log message that the download is happening.

I also have the feeling that S3 download is faster than the other repos. A possible (but clumpsy) solution would be to download SRTM from the public server once and only once, and make a local copy on S3 for faster download for the next run. 

For the very long term, we should find a solution so that external users can get access to our S3 drive.

TimoRoth:
The s3 drive can be configured to be public, so anyone can download from it via http or boto s3 api calls.
I'm just unsure if this is acceptable for the data on there, as we would be redistributing it which might not be ok with the owners.

fmaussion:
Yes, exactly. I'll ask Ben about that

fmaussion:
But it doesnt have to be entirely public: would it be possible to have guest read-only credentials?

TimoRoth:
Yes, it's possible to craft credentials to behave like that.
Only Ben can do so via the S3 Webinterface.

fmaussion:
OK, perfect. For now, it is not very important, but good to know that its possible

fmaussion:
[...]

PR #123, opened 2016-10-12
Add MPI support

TimoRoth:
Adds support for distributing tasks via MPI.
MPI allows for distributing the tasks across multiple hosts.
Primarily used on clusters and similar environments.

Usage:
`mpiexec python3 ./run_rgi_region.py --mpi-master --mpi-count=16`

One limitation of this is the lack of a common lock for safe downloads.
And it seems like nothing like this exists, so preventing accidental parallel downloads of the same file is not possible with the current approach.

I'll look into implementing a fetch-only run, that doesn't do any calculation and just downloads all the required data, if that's possible at all.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -33,7 +33,7 @@ install:
 
   # install dependencies
   - "python -m pip install --upgrade pip"
-  - "conda install --yes --quiet -c conda-forge geopandas matplotlib Pillow joblib netCDF4 scikit-image configobj nose pyproj numpy rasterio krb5 xarray"
+  - "conda install --yes --quiet -c conda-forge geopandas matplotlib Pillow joblib netCDF4 scikit-image configobj nose pyproj numpy rasterio krb5 xarray filelock"
   - "pip install git+https://github.com/fmaussion/motionless.git"
   - "pip install git+https://github.com/fmaussion/salem.git"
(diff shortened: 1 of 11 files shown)


fmaussion:
This is brilliant, thanks! I can't comment much on the code, but looks good to me.

As for the downloads:
- some downloads are going to happen all the time (i.e. the oggm-sample-data files, CRU, RGI) and could be downloaded before anything else. It's  really just the topography data which is dependent on which region you wanna do. 
- for now, a quick solution could be to use lock files in order to let the threads now that something is being downloaded
- other things with downloads need to be fixed (see https://github.com/OGGM/oggm/issues/86, https://github.com/OGGM/oggm/issues/69, https://github.com/OGGM/oggm/issues/58). I'll try to summarize them in a new issue and then maybe we'll find time to tackle them.

Can I merge this now?

TimoRoth:
I want to do some more tests on an actually clustered environment tomorrow, so far I only tested with multiple processes on the same host.

TimoRoth:
Ok, this should be fine once all the tests succeeded.
Still couldn't test on a clustered environment, due to lack of a clustered environment, but it runs fine locally and I don't see why it wouldn't work.

fmaussion:
Great! And if it doesn't work we wouldn't notice for now, so for me it's fine how it is.

I'll come back to you about this download stuff soon.

PR #124, opened 2016-10-16
Various MPI related improvements

TimoRoth:
Use async recv/send functions in the slave process, so it can receive the next task in the background while the task is calculating, greatly reducing idle times.

Add some helper functions to speed up setting up an MPI-Capable cluster on multiple AWS nodes to our fabfile.

Improve overall parallelization, by making init_glacier_regions parallel as well.
For this to work, execute_entity_task was extended with a mode for per-gdir kwargs.

Add a cfg setting for the mpi recv buffer size. The default size is too small to hold some of the parameters.
So instead of hardcoding a higher guess, I made it available as config parameter, with the default being high enough for the entire rgi region test script.

### DIFF
--- a/oggm/workflow.py
+++ b/oggm/workflow.py
@@ -107,11 +107,14 @@ def init_glacier_regions(rgidf, reset=False, force=False):
             rmtree(fpath)
 
     gdirs = []
+    new_gdirs = []
     for _, entity in rgidf.iterrows():
         gdir = oggm.GlacierDirectory(entity, reset=reset)
         if not os.path.exists(gdir.get_filepath('dem')):
-            tasks.define_glacier_region(gdir, entity=entity)
+            new_gdirs.append((gdir, dict(entity=entity)))
         gdirs.append(gdir)
+
+    execute_entity_task(tasks.define_glacier_region, new_gdirs)
 
     return gdirs
 
--- a/oggm/workflow.py
+++ b/oggm/workflow.py
@@ -5,6 +5,7 @@ from __future__ import division
 import logging
 import os
 from shutil import rmtree
+import collections
 # External libs
 import pandas as pd
 import multiprocessing as mp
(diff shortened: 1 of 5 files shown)


oggm/workflow.py
```diff
@@ -34,20 +34,33 @@ def _init_pool():
...
+            gdir, gdir_kwargs = gdir
+            gdir_kwargs = _merge_dicts(self.out_kwargs, gdir_kwargs)
+            return self.call_func(gdir, **gdir_kwargs)
+        else:
+            return self.call_func(gdir, **self.out_kwargs)
 
 
-def execute_entity_task(task, gdirs, **kwargs):
+def execute_entity_task(task, gdirs, gdir_has_kwargs=False, **kwargs):
     """Execute a task on gdirs.
```
fmaussion: I think it would be cleaner to remove the optional ``**kwargs`` as a whole, and have following function signature:

```python

def execute_entity_task(task, gdirs, kwargs=None):
    """Execute a task on gdirs.

    If you asked for multiprocessing, it will do it.

    Parameters
    ----------
    task: function
        the entity task to apply
    gdirs: list
        the list of oggm.GlacierDirectory to process
    kwargs: None, dict, or list of dict
        the keyword arguments to pass to the task. Set to a dict if the kwargs
        are the same for all instances of the task, and to a list of dict
        to pass different arguments to each instance (in that case, len(kwargs) 
        must be equal to len(gdirs))
    """
```

What do you think? We should be careful to see if this doesn't brake some code elsewhere

```

fmaussion:
This is great! Can't wait to test it.

The actual reason why ``init_glacier_region`` was not parallelized is because I had the naive thought that concurrent read access on a file are going to be problematic. It turns out that the system can handle this well, albeit wil probable performance issues.

oggm/workflow.py
TimoRoth: There are a couple problems with that signature:
It breaks compatibility with existing code, which is currently passing kwargs directly to the function.
With that, it would be impossible to potentially have common and per-gdir arguments at the same time.

Having the gdir and its kwargs in separate objects makes handling it in the function, and i believe in calling code as well, more complicated. You need to maintain two lists, match them up. A simple for gdir in gdirs loop won't work anymore.

I think the additional parameter can be safely dropped though, checking if the gdir is a tupple or not should be enough of a check, and would also enable a gdirs list with mixed normal gdirs and gdirs with kwargs.

TimoRoth:
It should not cause any problems, the system will cache the file rather quickly and read calls won't hit the disk anymore.

fmaussion:
LGTM! Merging now?

TimoRoth:
Yes, should be fine.
I'm running a couple rgi region benchmarks on 4 8 Core AWS instances right now, to see how it performs and if something fails in the long run.

fmaussion:
Thanks a lot!

PR #127, opened 2016-10-22
Salem update and deprecation of cleo

fmaussion:
There has been a few backwards incompatible changes in salem, and cleo has been merged into the larger project. This simplifies things a bit.


fmaussion:
cc @TimoRoth , I guess docker has to be updated for the tests to pass

TimoRoth:
Pushed the Docker update.
Tests will use it once it finished building, which can take an hour or two:
https://hub.docker.com/r/oggm/untested_base/builds/

fmaussion:
Thanks! It was indeed an issue with Salem, I hope it's solved now

PR #132, opened 2016-11-15
Various Updates and improvements

TimoRoth:
First, some minor fabfile Updates.

Then primarily some updates improving the log output when errors happen while running tasks, so one allways knows where the error is coming from, and on which glacier it happened.

Also adds a decorator for self_iterating_entity_task.
This helps, as those tasks can then be passed to execute_entity_tasks, which has a special case for them that just calls them. That way they can be in the same list of tasks as normal tasks and don't need special cases.

Finally a quick update to the installation instructions, making use of the new conda packages.

### DIFF
--- a/deployment/fabfile.py
+++ b/deployment/fabfile.py
@@ -508,6 +508,7 @@ def node_install(cn=def_cn,inst_type_idx=def_inst_type,idn=0,
                       key_name=key_name,
                       placement=avz,
                       subnet_id=subnet_id,
+                      ebs_optimized=True,
                       instance_type=instance_infos[inst_type_idx]['type'],
                       block_device_map=bdm)
         req_ids = [request.id for request in requests]
@@ -522,6 +523,7 @@ def node_install(cn=def_cn,inst_type_idx=def_inst_type,idn=0,
                 placement=avz,
                 subnet_id=subnet_id,
                 security_group_ids=[group.id],
+                ebs_optimized=True,
                 instance_type=instance_infos[inst_type_idx]['type'],
                 block_device_map=bdm)
         node = reservation.instances[0]
--- a/deployment/fabfile.py
+++ b/deployment/fabfile.py
@@ -477,19 +477,10 @@ def node_install(cn=def_cn,inst_type_idx=def_inst_type,idn=0,
         else:
             raise
 
+    # Authorize all Intra-VPC traffic
     if vpc is not None:
         try:
-            group.authorize('tcp', 0, 65535, vpc.cidr_block)
-        except cloud.ResponseError as e:
-            if e.code != 'InvalidPermission.Duplicate':
-                raise
-        try:
-            group.authorize('udp', 0, 65535, vpc.cidr_block)
-        except cloud.ResponseError as e:
-            if e.code != 'InvalidPermission.Duplicate':
-                raise
-        try:
-            group.authorize('icmp', 0, 255, vpc.cidr_block)
+            group.authorize('-1', -1, -1, vpc.cidr_block)
         except cloud.ResponseError as e:
             if e.code != 'InvalidPermission.Duplicate':
                 raise
--- a/docs/installing-oggm.rst
+++ b/docs/installing-oggm.rst
@@ -127,7 +127,12 @@ Packages
 
 Install the packages from the `conda-forge`_ channel::
 
-    conda install -c conda-forge geopandas matplotlib Pillow joblib netCDF4 scikit-image configobj nose pyproj numpy krb5 rasterio xarray filelock
+    conda install -c oggm -c conda-forge oggm-deps
+
+The oggm-deps package is a meta package. It does not contain any actual code,
+but only depends on all packages oggm needs, in order to simplify setting up
+an environment.
+If you want to install the actual oggm, install oggm instead of oggm-deps.
 
 .. warning::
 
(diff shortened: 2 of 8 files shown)


TimoRoth:
Seems like python2 doesn't work anymore on Windows, in line with the other build environments, just disabling it seems the way to go.

deployment/fabfile.py
```diff
@@ -99,7 +99,7 @@
 def_cn = 'AWS'
 
 # Change to a string identifying yourself
-user_identifier = None
+user_identifier = "Timo"
```
fmaussion: Keep None here

docs/installing-oggm.rst
```diff
@@ -127,7 +127,11 @@ Packages
 
 Install the packages from the `conda-forge`_ channel::
 
-    conda install -c conda-forge geopandas matplotlib Pillow joblib netCDF4 scikit-image configobj nose pyproj numpy krb5 rasterio xarray filelock
+    conda install -c oggm -c conda-forge oggm-deps
+
+If you want to install oggm itself to just use it, install oggm instead of oggm-deps.
+oggm-deps is a meta package that contains no actual content, and just depends on all
+dependencies of oggm.
```
fmaussion: I *think* I understand what you mean, but it could be clearer. Most people reaching this page are python newcomers, and need to be taken by the hand a little bit.

Do you mean that a ``conda install -c oggm -c conda-forge oggm`` will install all dependencies + oggm per git (i.e. not in dev mode)?

oggm/sandbox/run_rgi_region.py
```diff
@@ -92,6 +92,8 @@
 assert len(rgi_shp) == 1
 rgidf = gpd.read_file(rgi_shp[0])
 
+rgidf = rgidf[750:1000]
+
```
fmaussion: I think this file should be left unchanged

oggm/utils.py
```diff
@@ -1371,18 +1371,19 @@ def _entity_task(gdir, **kwargs):
...
-            else:
+            try:
                 out = task_func(gdir, **kwargs)
+                gdir.log(task_func)
+            except Exception as err:
+                # Something happened
+                out = None
+                gdir.log(task_func, err=err)
+                pipe_log(gdir, task_func, err=err)
+                self.log.error('%s occured during task %s on %s!', type(err).__name__, task_func.__name__, gdir.rgi_id)
```
fmaussion: Is this line PEP8 compliant?

oggm/utils.py
```diff
@@ -1430,6 +1431,18 @@ def _divide_task(gdir, div_id=None, **kwargs):
         return _divide_task
 
 
+def self_iterating_entity_task(task_func):
+    """
```
fmaussion: I think I'd rather give it some other name. "entity task" is a well defined concept in the (admittedly parse) documentation. In the API doc I called them ``global task``: http://oggm.readthedocs.io/en/latest/api.html#global-tasks

I'm open to other suggestions, though.

fmaussion:
Brilliant, thanks! 

Shouldn't the new decorator imply more changes elsewhere, too? I can think of https://github.com/OGGM/oggm/blob/master/oggm/tests/test_workflow.py , but probably also some code in the sandbox. At least ``run_rgi_region`` (which is the cleanest example) could be updated.

docs/installing-oggm.rst
TimoRoth: There are two packages for oggm on conda.
oggm and oggm-deps.
oggm contains the actual oggm code, as you get if you run pip install on oggm.
oggm-deps does not contain any code, and only pulls in the required dependencies to run oggm, so you can use your own fork of it.

oggm/utils.py
TimoRoth: Yes, the name is indeed a bit clunky.
Didn't know about them being called global_tasks, calling the decorator global_task seems way better.

TimoRoth:
Yes, most of the examples can be simplified now.

docs/installing-oggm.rst
fmaussion: Yes ok, that's how I understood it.

oggm/utils.py
fmaussion: yes ok, then let's go with ``global_task`` then

TimoRoth:
This should be it. I didn't update the sandbox runs for now, as at least bea is working on them and it would mess things up quite a bit if I were to overhaul them now.

docs/installing-oggm.rst
```diff
@@ -137,27 +142,18 @@ Install the packages from the `conda-forge`_ channel::
...
-    pip install progressbar2
-
 No scientific python installation is complete without installing
 `ipython`_ and `jupyter`_::
 
     conda install -c conda-forge ipython jupyter
 
-After success, install the following packages from Fabien's github::
-
-    pip install git+https://github.com/fmaussion/motionless.git
```
fmaussion: It's unrelated to this PR, but did you update your builds to use the newly released package on PyPi? It's no longer necessary to use my fork for install

fmaussion:
Looks good to me, thanks! I like the idea of the global task decorator.

docs/installing-oggm.rst
TimoRoth: So far everything is still using the fork.
I'll update docker and conda to use the latest official version.

PR #145, opened 2016-12-20
DOC: add a flowline page and mass-balance improvements

fmaussion:


fmaussion:
Ignoring appveyor fails which are unrelated

PR #146, opened 2017-01-01
Get DEM downloads right

fmaussion:
This finalizes the great work by @JohannesUIBK . The error was a simple win/linux compatibility issue.

The ``get_topo_file`` is a bit cleverer now, with possibility for the user to select the DEM source manually for each RGI glacier.

DEM3 is the default before ASTER.


fmaussion:
Test coverage decreased again, there is plenty of room for a nice refactoring of the download code.

PR #152, opened 2017-01-17
WIP: Move testing framework to pytest

fmaussion:
This PR contains the following changes:
- use pytest instead of nose (which is deprecated)
- use pytest-mpl for the image comparisons (https://github.com/matplotlib/pytest-mpl)
- removes the baseline (reference) images from the code repository (they are now in [oggm-sample-data](https://github.com/OGGM/oggm-sample-data/tree/master/baseline_images/2.0.x))
- the new graphic tests run with MPL 2+ only
- the plots are no longer compared per default, which is a good thing (pytest run without ``--mpl`` option just checks that the code runs without performing image comparisons)
- motionless is no longer installed from my github

CC @TimoRoth for  review + docker updates (we need the ``pytest``, ``pytest-mpl`` and ``pytest-cov`` packages)

There is still one reference to nose here: https://github.com/OGGM/oggm/blob/master/deployment/docker/docker-compose.test.yml . Not sure exactly what to do with that though.


TimoRoth:
The command in docker-compose.test.yml is the test command run by Docker Hub before publishing a new build, it needs to be updated as well.
I never managed to get a plain nosetests invocation to work, that's why I came up with that slightly odd invocation via python. Will investigate how it should look for pytest

TimoRoth:
Oh, and the base image is updated, travis is giving meaningful test errors now.

TimoRoth:
Ok, just replace the command: in docker-compose.test.yml with `command: /root/test.sh`

fmaussion:
OK, I artificially pushed up the image tests tolerance in order to get the tests to pass, but it would be great if we had a chance to look at the images.

@TimoRoth , would you have a bit of time to investigate on how do do this? I think it will involve an ``after_failure`` script on Travis (see https://github.com/matplotlib/pytest-mpl/issues/40)
Matplotlib does something which looks like what we need here: https://github.com/matplotlib/matplotlib/blob/master/.travis.yml#L191

TimoRoth:
I'll just run the tests manually and upload the images.

fmaussion:
> I'll just run the tests manually and upload the images.

Thanks! You'll have to push down the tolerance to 2 then. On the long term I wonder if it woudn't be a good idea to have this travis mechanism though.

Issue #151, opened 2017-01-16
Bug in untils.py?


anoukvlug:
The 5th of January is started using an updated version of oggm. When I run oggm the following error occured:

```
> /vega1/avlug/Programs/miniconda3/envs/oggm_env/bin/python3.5 /vega1/avlug/Programs/oggm/oggm/Anouk/run_rgi_regionSelectionTotalTest.py
> 2017-01-10 17:16:02: oggm.cfg: Parameter file: /vega1/avlug/Programs/oggm/oggm/params.cfg
> Downloading http://www.glims.org/RGI/rgi50_files/rgi50.zip ...
> 100% of 410.4 MiB |#######################| Elapsed Time: 0:01:32 Time: 0:01:32
> OrderedDict([('working_dir', '/vega1/avlug/Programs/oggm/Workspace/A3smalltest'), ('topo_dir', '/vega1/avlug/Programs/oggm/oggm-data/topo'), ('cru_dir', '/vega1/avlug/Programs/oggm/oggm-data/cru'), ('rgi_dir', '/vega1/avlug/Programs/oggm/oggm-data/rgi'), ('dem_file', '~'), ('climate_file', '~'), ('wgms_rgi_links', '~'), ('glathida_rgi_links', '~'), ('leclercq_rgi_links', '~')])
> Delete all glacier directories? [Y/n] y
> 2017-01-10 17:19:05: oggm.core.preprocessing.gis: RGI50-03.00840: define_glacier_region
> Downloading http://viewfinderpanoramas.org/dem3/T15.zip ...
> 100% of  11.2 MiB |#######################| Elapsed Time: 0:00:01 Time: 0:00:01
> 2017-01-10 17:19:07: oggm.core.preprocessing.gis: KeyError occured during task define_glacier_region on RGI50-03.00840!
> Traceback (most recent call last):
>   File "/vega1/avlug/Programs/oggm/oggm/Anouk/run_rgi_regionSelectionTotalTest.py", line 95, in <module>
>     gdirs = workflow.init_glacier_regions(rgidf, reset=True)
>   File "/vega1/avlug/Programs/oggm/oggm/workflow.py", line 120, in init_glacier_regions
>     execute_entity_task(tasks.define_glacier_region, new_gdirs)
>   File "/vega1/avlug/Programs/oggm/oggm/workflow.py", line 94, in execute_entity_task
>     pc(gdir)
[...]

PR #153, opened 2017-01-18
PopChange

anoukvlug:
This pull request is a result of the Key error that was discussed in issue #151 "Bug in utils.py?".

Two times:

`profile.pop('affine')`

in utils.py is changed for 

```
if 'affine' in profile:
    profile.pop('affine')
```


TimoRoth:
Failing tests are unrelated, caused by the imminent update to matplotlib 2.0.
Thanks

PR #156, opened 2017-01-24
Update CRUTS to version 3.24

TimoRoth:
As simple as it gets. If the tests pass, it can be merged right away.


fmaussion:
Thanks! Indeed, this is not tested so I just believe you ;-)

PR #161, opened 2017-02-10
Several changes in run_alaska.py

bearecinos:
-Unecessary comments deleted
-now calving loop runs for all tidewater glaciers inside
gdirs and not just one selected glacier directory
(before: gdir=gdirs[a number])


fmaussion:
Great, thanks!

PR #162, opened 2017-02-14
Uniformize and document the mass-balance models

fmaussion:


fmaussion:
There is a failing test on Appveyor which I don't understand, I'll just ignore it for now

PR #166, opened 2017-02-16
Python 3.5 finally seems to work on Windows Anaconda

TimoRoth:


fmaussion:
mh, there still seems to be a problem...

fmaussion:
OK so the failing test is unrelated. I'll try to have a look at this when I get a windows machine next week.

Thanks @TimoRoth !

PR #174, opened 2017-02-21
Downstream bedshape based on parabolic assumption

anton-ub:


fmaussion:
Going to merge this now (finally) although there is a few things to do on top of it.

Thanks @anton-ub !

PR #181, opened 2017-03-10
WIP: add intersect geometries

fmaussion:
This PR will add the possibility to take into account that glaciers can have ice divides with other glaciers, or that neighbor tributaries can be touching each other and therefore shouldn't assume a parabolic bed.

This is going to imply quite a bit of change. I will update this PR from time to time.

(vision: being largely finished before the workshop)


fmaussion:
Closed in favor of commit towards dev branch

PR #182, opened 2017-03-10
Add a regional plot

fmaussion:
This makes regional plots, for example:

test_plot_region_inversion

cc @bearecinos


bearecinos:
omg FABIEN THIS IS AWESOME!

PR #185, opened 2017-03-13
Bug fix in glacier_characteristics + better log file names

fmaussion:
Solved a bug when glacier_characteristics was called on a gdir which errored in the workflow.

Logfile names now start with date.


fmaussion:
This should solve the problem we had earlier today @bearecinos

bearecinos:
Good thanks I will update the oggm in the AWS again and try it out... I do think that in some of the glaciers a run will still give problems because of the divides image that I showed you yesterday.

Issue #157, opened 2017-01-25
Time limit in process_custom_climate_data


anoukvlug:
When my climate file starts in 1600 (or earlier) the years are not recognized as years any more.

`yrs = nc_ts.time.year`
`AttributeError: 'Index' object has no attribute 'year'`

I'm planning to make some changes in my fork, to solve this problem.
Would it make sense to make a pull request of these chances later on?

fmaussion:
Hi Anouk, 

This is getting quite interesting, I'd like to know more about your plans ;-) But first, to your question:

I expect that solving this problem won't be trivial, as there is an underlying problem in pandas and xarray (see e.g. https://github.com/pydata/xarray/issues/789 for xarray, which relies on pandas: https://github.com/pandas-dev/pandas/issues/7307 ).

OGGM does not rely heavily on pandas datetimes so I _think_ we could solve this problem without getting burned in a hell of changes, but on the top of my head I can't really make a prognostic about how much change this would imply. 

In fact, the dynamical model of OGGM already relies on a very simple time system (https://github.com/OGGM/oggm/blob/master/oggm/utils.py#L675-L718), which could be meliorated and extended (based on other peoples systems, as I hope we don't have to reinvent the wheel here).

I know this is not very helpful, but yes a change that would make OGGM work on longer timescales is necessary (I just didn't have that in mind when I started :-( )

anoukvlug:
Hello Fabien,
Thank you for your quick response. My first idea is to change
```
yrs = nc_ts.time.year
y0, y1 = yrs[0], yrs[-1]
```

into

```
time1 = nc_ts.variables['time']
time1 = netCDF4.num2date(time1[:], time1.units)
y0=int(time1[0].strftime('%Y'))
y1=int(time1[-1].strftime('%Y'))
```
to avoid using pandas. 

I don't know yet what to do with these lines:
`nc_ts.set_period(t0='{}-10-01'.format(y0), t1='{}-09-01'.format(y1))`
`gdir.write_monthly_climate_file(time, iprcp, itemp, igrad, ihgt, ref_pix_lon, ref_pix_lat)`

So far these are the only lines that has give me problems with the starting date. So hopefully this requires just a small change. I will let you know once I know more about what I what to do with those lines.

[...]

PR #188, opened 2017-03-15
Gcm data

anoukvlug:
I made this PR, to make it possible to use CESM-LME simulations as climate forcing in oggm.

The issue related to this PR is: #157

### DIFF
--- a/oggm/core/preprocessing/climate.py
+++ b/oggm/core/preprocessing/climate.py
@@ -136,16 +136,15 @@ def process_custom_climate_data(gdir):
     out = {'climate_source': fpath, 'hydro_yr_0': y0+1, 'hydro_yr_1': y1}
     gdir.write_pickle(out, 'climate_info')
 
-@entity_task(log, writes=['gcm_data'])
-def process_gcm_data(gdir):
+@entity_task(log, writes=['cesm_data'])
+def process_cesm_data(gdir):
     """Processes and writes the climate data for this glacier.
 
-    Interpolates the GCM data to the high-resolution CL2 climatologies
+    This function is made for interpolating the Community
+    Earth System Model Last Millenial Ensemble (CESM-LME) climate simulations,
+    from Otto-Bliesner et al. (2016), to the high-resolution CL2 climatologies
     (provided with OGGM) and writes everything to a NetCDF file.
     """
-    # read the climatology high resolution CL2 climatologies
-    clfile = utils.get_cru_cl_file()
-    ncclim = salem.GeoNetcdf(clfile)
 
     # GCM temperature and precipitation data
     if not (('gcm_temp_file' in cfg.PATHS) and \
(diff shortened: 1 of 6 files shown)


oggm/core/preprocessing/climate.py
```diff
@@ -136,6 +136,230 @@ def process_custom_climate_data(gdir):
     out = {'climate_source': fpath, 'hydro_yr_0': y0+1, 'hydro_yr_1': y1}
     gdir.write_pickle(out, 'climate_info')
 
+@entity_task(log, writes=['gcm_data'])
+def process_gcm_data(gdir):
```
fmaussion: I think it should have a more precise name. What GCM data is this function handling currently? This seems like a special model output, so maybe it could be specified in the name and in the docstring.

oggm/core/preprocessing/climate.py
```diff
@@ -136,6 +136,230 @@ def process_custom_climate_data(gdir):
...
+        raise IOError('GCM precl file not found')
+
+
+    # read the file
+    fpathTemp = cfg.PATHS['gcm_temp_file']
+    fpathPrecc = cfg.PATHS['gcm_precc_file']
+    fpathPrecl = cfg.PATHS['gcm_precl_file']
+    temp = salem.open_xr_dataset(fpathTemp)
+    temp = temp.TREFHT
+    temp.time
```
fmaussion: This line is useless

oggm/core/preprocessing/climate.py
```diff
@@ -136,6 +136,230 @@ def process_custom_climate_data(gdir):
...
+    temp = temp[9:-3, :, :] # from normal years to hydrological years
```
fmaussion: This is ok for now, but good practice would imply that we check if the time series are clean (i.e. here you assume a start in January, which is likely but not guaranteed)

oggm/core/preprocessing/climate.py
```diff
@@ -136,6 +136,230 @@ def process_custom_climate_data(gdir):
...
+@entity_task(log, writes=['gcm_data'])
+def process_gcm_data(gdir):
+    """Processes and writes the climate data for this glacier.
+
+    Interpolates the GCM data to the high-resolution CL2 climatologies
+    (provided with OGGM) and writes everything to a NetCDF file.
+    """
+    # read the climatology high resolution CL2 climatologies
+    clfile = utils.get_cru_cl_file()
+    ncclim = salem.GeoNetcdf(clfile)
```
fmaussion: Move these two lines down, and open the file at the moment it is needed.

oggm/core/preprocessing/climate.py
```diff
@@ -136,6 +136,230 @@ def process_custom_climate_data(gdir):
...
+    temp = salem.open_xr_dataset(fpathTemp)
+    temp = temp.TREFHT
+    temp.time
+    temp = temp[9:-3, :, :] # from normal years to hydrological years
+    precpC = xr.open_dataset(fpathPrecc)
+    precpL = xr.open_dataset(fpathPrecl)
+    precp = precpC.PRECC+precpL.PRECL
+    precp = precp[9:-3, :, :] # from normal years to hydrological years
+
+    precp.time # Is this line needed?
```
fmaussion: Not needed

oggm/core/preprocessing/climate.py
```diff
@@ -136,6 +136,230 @@ def process_custom_climate_data(gdir):
...
+    temp['time'] = pd.period_range('850-10', '2005-9', freq='M')
+    precp['time'] = pd.period_range('850-10', '2005-9', freq='M')
```
fmaussion: Here also, this decision could be based on info obtained from the file. But let's keep it like this for now.

oggm/core/preprocessing/climate.py
```diff
@@ -136,6 +136,230 @@ def process_custom_climate_data(gdir):
...
+
+    # Use the year and month for selection
+    year = np.array([t.year for t in temp.time.values])  # selects for each point in time the year
+    month = np.array([t.month for t in temp.time.values])  # selects for each point in time the month
+
+    year = np.array([t.year for t in precp.time.values])  # selects for each point in time the year
+    month = np.array([t.month for t in precp.time.values])  # selects for each point in time the month
```
fmaussion: These two lines are a repetition from above

oggm/core/preprocessing/climate.py
```diff
@@ -136,6 +136,230 @@ def process_custom_climate_data(gdir):
...
+            loc_tmp_mth = loc_tmp[i, ...][isok].flatten()
+            slope, _, _, p_val, _ = stats.linregress(hgt_f, loc_tmp_mth)
+            ts_grad[i] = slope if (p_val < 0.01) else def_grad
+    # ... but dont exaggerate too much
+    ts_grad = np.clip(ts_grad, g_minmax[0], g_minmax[1])
+    # convert to timeserie and hydroyears
+    ts_grad = ts_grad.tolist()
+    ts_grad = ts_grad[9:] + ts_grad[0:9]
+    ts_grad = np.asarray(ts_grad * ny)
+
```
fmaussion: Instead of doing all these things above, you could actually assume that a climate file is already present in the directory (which it should), and use this data as reference.

This is actually better because it avoids code repetition, and makes sure that we agree upon the actual climate data. 

Also you don't need to take the three closest pixels anymore, just the actual time series

oggm/core/preprocessing/climate.py
```diff
@@ -136,6 +136,230 @@ def process_custom_climate_data(gdir):
...
+    latdif=abs((temp.lat)-lat)
+    latmin=min(latdif)
+    latsel=np.where(latdif==latmin)
+    latsel=np.asarray(latsel)
+    latsel=latsel.squeeze()
+    temp = temp.isel(lat=((latsel-1), latsel, (latsel+1))) 
+    precp = precp.isel(lat=((latsel-1), latsel, (latsel+1)))
+
+    longi = np.hstack((temp.lon[:73], (temp.lon[73:]-360)))
```
fmaussion: This could be done more elegantly with ``np.where()``

oggm/core/preprocessing/climate.py
```diff
@@ -136,6 +136,230 @@ def process_custom_climate_data(gdir):
...
+    lonsel=np.where(londif==lonmin)
+    lonsel=np.asarray(lonsel)
+    lonsel=lonsel.squeeze()
+    temp=temp.isel(lon=((lonsel-1), lonsel, (lonsel+1)))
+    precp = precp.isel(lon=((lonsel-1), lonsel, (lonsel+1)))
+
+    Ndays = np.tile([31, 30, 31, 31, 28, 31, 30, 31, 30, 31, 31, 30], (y1 - y0))
+    for i in range(0,3):
+        for j in range(0,3):
+            precp[: ,i, j] = precp[: ,i, j]* (60 * 60 * 24 * 1000) * Ndays # from [m/s] to [kg/m^2 per month]
```
fmaussion: No need for loops here: xarray (or numpy) could take care of the vectorized operations.

oggm/core/preprocessing/climate.py
```diff
@@ -136,6 +136,230 @@ def process_custom_climate_data(gdir):
...
+
+    temp.salem.grid
+    precp.salem.grid
```
fmaussion: Two lines not needed

oggm/core/preprocessing/climate.py
```diff
@@ -136,6 +136,230 @@ def process_custom_climate_data(gdir):
...
+                    found_it = True
+        if not found_it:
+            msg = '{}: OMG there is no climate data'.format(gdir.rgi_id)
+            raise RuntimeError(msg)
+    elif np.any(~np.isfinite(ts_tmp)):
+        # maybe the side is nan, but we can do nearest
+        ts_tmp = ncclim.grid.map_gridded_data(ts_tmp.values, temp.salem.grid,
+                                              interp='nearest')
+        ts_pre = ncclim.grid.map_gridded_data(ts_pre.values, precp.salem.grid,
+                                              interp='nearest')
```
fmaussion: I also think all this is obsolete if you use the existing climate file.

oggm/core/preprocessing/climate.py
```diff
@@ -136,6 +136,230 @@ def process_custom_climate_data(gdir):
...
+    assert np.all(np.isfinite(ts_grad))
+    gdir.write_gcm_data_file(time2, ts_pre.values, ts_tmp.values,
+                                    ts_grad, loc_hgt, loc_lon, loc_lat)
+
+    ncclim._nc.close()
+    Tindex._nc.close()
+
+    # metadata
+    out = {'climate_source': 'GCM data', 'hydro_yr_0': y0 + 1, 'hydro_yr_1': y1}
+    gdir.write_pickle(out, 'climate_info')
```
fmaussion: This overwrites the existing metadata, which isn't desirable. You should remove these two lines for now.

oggm/utils.py
```diff
@@ -1970,6 +1970,51 @@ def write_monthly_climate_file(self, time, prcp, temp, grad, ref_pix_hgt,
             v.long_name = 'temperature gradient'
             v[:] = grad
 
+    def write_gcm_data_file(self, time, prcp, temp, grad, ref_pix_hgt,
+                                   ref_pix_lon, ref_pix_lat):
```
fmaussion: Is this enteriley new function really needed, or could we make the existing ``write_climate_data`` more flexible in order to handle GCM data, too?

oggm/core/preprocessing/climate.py
anoukvlug: Currently the function is handling CESM data from the Last Millennial Ensemble. 
I could change the function name into process_cesm_data and give more information in the doc string.

oggm/core/preprocessing/climate.py
anoukvlug: I will do that.

oggm/core/preprocessing/climate.py
anoukvlug: I will remove it.

oggm/core/preprocessing/climate.py
anoukvlug: I will remove it.

oggm/core/preprocessing/climate.py
anoukvlug: I will remove those.

oggm/core/preprocessing/climate.py
anoukvlug: I didn't think of that, but it actually makes a lot of sense. So you mean, using the climate_monthly.nc file for that?

oggm/core/preprocessing/climate.py
anoukvlug: I will do that.

oggm/core/preprocessing/climate.py
anoukvlug: I will remove those.

oggm/utils.py
anoukvlug: I think this new function is not needed. It would indeed be better to make `write_climate_data` more flexible.

fmaussion:
Great, thanks! No worries if you need time, we can have a look at the details in Bremen, too

anoukvlug:
Thank you for all your feedback Fabien! I included your suggestions in the commit that I just uploaded. 
It would indeed  be nice to discuss the remaining points once you're here :)

How should the mini-climate file look like? Could it be a 3x3 grid selection of one of the CESM files?

fmaussion:
>  Could it be a 3x3 grid selection of one of the CESM files?

Yes, exactly! But no worries, we can do that in Bremen

fmaussion:
Great addition, thanks!

anoukvlug:
Thank you for helping me! 
I made the 3x3 mini climate files. Shall I upload those via a pull request?

fmaussion:
Please send them to me via slack first, I'd like to have a look at them

PR #192, opened 2017-03-16
fabfile.py updates

TimoRoth:
To speed up instance creation, I created AMIs with the OGGM dependencies already pre-installed.
This should speedup software installation by quite a bit.

Also force the C locale, in an attempt to fix the locale issues which I still haven't been able to reproduce on my end.


fmaussion:
Thanks! Will try when I have time

PR #198, opened 2017-03-23
Make define_glacier_region more flexible

jlandmann:
Up to now, the glacier region resolution was determined as a function of the glacier area.
Actually, it is quite easy to also introduce a fixed resolution by setting d1 in params.cfg to 0 and then ds gives the desired resolution, however, it is more convenient to have also a fixed method.

### DIFF
--- a/oggm/params.cfg
+++ b/oggm/params.cfg
@@ -73,7 +73,7 @@ mpi_recv_buf_size = 131072
 ### CENTERLINE determination
 
 # Decision on grid spatial resolution for each glacier
-grid_dx_method = 'linear'
+grid_dx_method = 'fixed'
 # 'fixed': dx (meters) = fixed_dx
 # 'linear':  dx (meters) = d1 * AREA (km) + d2 ; clipped to dmax (e.g.: 5, 10, 200)
 # 'square':  dx (meters) = d1 * sqrt(AREA) (km) + d2 ;  clipped to dmax (e.g.: 20, 10, 200)
--- a/oggm/params.cfg
+++ b/oggm/params.cfg
@@ -73,12 +73,14 @@ mpi_recv_buf_size = 131072
 ### CENTERLINE determination
 
 # Decision on grid spatial resolution for each glacier
-grid_dx_method = 'linear'
+grid_dx_method = 'fixed'
+# 'fixed': dx (meters) =
 # 'linear':  dx (meters) = d1 * AREA (km) + d2 ; clipped to dmax (e.g.: 5, 10, 200)
 # 'square':  dx (meters) = d1 * sqrt(AREA) (km) + d2 ;  clipped to dmax (e.g.: 20, 10, 200)
 d1 = 5.
 d2 = 10.
 dmax = 100.
+fixed_dx = 50.
 
 # Which algorithm to use for interpolating the topography to the local grid
 # 'bilinear' or 'cubic'
--- a/oggm/params.cfg
+++ b/oggm/params.cfg
@@ -73,7 +73,7 @@ mpi_recv_buf_size = 131072
 ### CENTERLINE determination
 
 # Decision on grid spatial resolution for each glacier
-grid_dx_method = 'fixed'
+grid_dx_method = 'linear'
 # 'fixed': dx (meters) = fixed_dx
 # 'linear':  dx (meters) = d1 * AREA (km) + d2 ; clipped to dmax (e.g.: 5, 10, 200)
 # 'square':  dx (meters) = d1 * sqrt(AREA) (km) + d2 ;  clipped to dmax (e.g.: 20, 10, 200)
--- a/oggm/params.cfg
+++ b/oggm/params.cfg
@@ -73,8 +73,8 @@ mpi_recv_buf_size = 131072
 ### CENTERLINE determination
 
 # Decision on grid spatial resolution for each glacier
-grid_dx_method = 'fixed'
-# 'fixed': dx (meters) =
+grid_dx_method = 'linear'
+# 'fixed': dx (meters) = fixed_dx
 # 'linear':  dx (meters) = d1 * AREA (km) + d2 ; clipped to dmax (e.g.: 5, 10, 200)
 # 'square':  dx (meters) = d1 * sqrt(AREA) (km) + d2 ;  clipped to dmax (e.g.: 20, 10, 200)
 d1 = 5.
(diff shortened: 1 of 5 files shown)


jlandmann:
Oooppssss...I actually wanted to add this to my fork, obviously something went wrong ;-)
You can delete this if you want, @fmaussion.
However, if you think this is useful for whatever reason also for the original version of OGGM, I'll make a further commit now in order not to break your default params.cfg

fmaussion:
Thanks! In general I would avoid adding code that isn't really necessary, but in that case it is straightforward enough so I'm fine with this change.

Could you also add a test for this? I see we don't have a dedicated test for these three ways to define dx, but it would  be quite easy to implement following this model: https://github.com/OGGM/oggm/blob/master/oggm/tests/test_prepro.py#L84, and test for ``dx`` explicitly.

jlandmann:
Thanks a lot @fmaussion...this would make the mistake not so bad at least ;-)

I have two things on my agenda:
1) Understand why 9 test failed after the first commit, when I set the resolution to a fixed value. In my opinion, this shouldn't break the other functions.
2) I was thinking about that test actually, but wouldn't it be nice to make the `dxmethod` a keyword then? This makes testing a lot easier.

fmaussion:
> Understand why 9 test failed after the first commit, when I set the resolution to a fixed value. In my opinion, this shouldn't break the other functions.

In a prefect world I would agree with you, but in this case I can see a couple of reasons why tests could fail:
- my tests are not always general enough
- the graphics tests are very demanding: they expect the ouptut plots to be very close to what is expected
- probably the most important thing: many aspects of the model are extremely dependent on topography (the altitude-area distribution determines the widths, the mass-balance, the slope, all of which will influence the results)

So if there is a possibility to make some tests more flexible I'd be totally happy, but I think this is quite a piece of work. 

> I was thinking about that test actually, but wouldn't it be nice to make the dxmethod a keyword then? This makes testing a lot easier.

Indeed, but if we start to do this then we will have A LOT of keywords (which was the original argument for a ``cfg`` module).

One possibility would be to have both keywords and ``cfg.PARAMS``, and always test at the beginning of the function whether the keyword is set and if not, use the params one. This is a general design question, and I have to think about it. In your case, testing is still quite easy:

```python
def test_dx_methods(self):
        hef_file = get_demo_file('Hintereisferner.shp')
        entity = gpd.GeoDataFrame.from_file(hef_file).iloc[0]
        gdir = oggm.GlacierDirectory(entity, base_dir=self.testdir)
        
        # Test fixed method
        cfg.PARAMS['grid_dx_method'] = 'fixed'
        cfg.PARAMS['fixed_dx'] = 50
        gis.define_glacier_region(gdir, entity=entity)
        np.testing.assert_allclose(gdir.dx, 50)
        
        # etc.
```

jlandmann:
> So if there is a possibility to make some tests more flexible I'd be totally happy, but I think this is quite a piece of work.

Well, I fear this would be my part then if I want so ensure integrity of the model despite my changes/extensions. The question is when I manage to fix it (I need to get my own model running...😱) and what we do with this PR in the meanwhile.

> Indeed, but if we start to do this then we will have A LOT of keywords (which was the original argument for a cfg module).

I definitely see your point...making the functions less "stiff" would make the model more messy. You can think about it, and in the meanwhile I introduce tests as you suggested by resetting the cfg.PARAMS entries for each individual dxmethod test.

fmaussion:
Yes ok, the failing tests with fixed dx are quite ok to solve (i.e. changing the test, not the code), only the test_workflow is a bit more problematic and more surprising. This even looks like a bug in OGGM maybe (compute_downstream_line)

jlandmann:
Okay, I fixed some minor things and there is two points where I get stuck:

1. Twice same error during nosetests:
`ERROR: oggm.tests.test_models.is_performance_test
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Program Files\Anaconda3\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
  File "C:\Program Files\Anaconda3\lib\site-packages\nose\util.py", line 620, in newfunc
    return func(*arg, **kw)
TypeError: is_performance_test() missing 1 required positional argument: 'test'`

The function is not used and once imported...if you uncomment it, everything works fine (I don't know if that is your intention). Error is independent from `fixed`/`linear` method.

2. Message in `oggm.core.preprocessing.geometry._line_interpol`: 
`third argument of GEOSProject_r must be Point*`
Appears when using the `fixed` method,  `pbs` can become a `LineString` instead of a Point. I'm trying to imagine how this works....

fmaussion:
Thanks for all this work!

I hope that switching to ``pytest`` instead of ``nosetests`` should solve point 1 (we stopped using nosetests)

For point 2, I need the full traceback to get an idea of what could go wrong

jlandmann:
I reply: Thanks for your patience! ;-)

Indeed, using pytest solved 1), and also 2) disappeared using pytest. In case of 2), maybe it just suppresses the messages as it's basically just a message (no error/traceback in nosetests).

This leaves us with the failing workflow tests....can you maybe check if this is a "real" bug (independent from my changes)? Otherwise I can also try to make this work.

oggm/params.cfg
```diff
@@ -73,12 +73,14 @@ mpi_recv_buf_size = 131072
 ### CENTERLINE determination
 
 # Decision on grid spatial resolution for each glacier
-grid_dx_method = 'linear'
+grid_dx_method = 'fixed'
```
fmaussion: For now, please revert to the default here

oggm/utils.py
```diff
@@ -866,9 +866,9 @@ def srtm_zone(lon_ex, lat_ex):
 
     # quick n dirty solution to be sure that we will cover the whole range
     mi, ma = np.min(lon_ex), np.max(lon_ex)
-    lon_ex = np.linspace(mi, ma, np.ceil((ma - mi) + 3))
+    lon_ex = np.linspace(mi, ma, int(np.ceil((ma - mi) + 3)))  # int() to avoid Deprec warning
```
fmaussion: Is this line (and others below) shorter than 79 characters? We try to apply PEP8 as far as possible

fmaussion:
Yes, the workflow tests are likely a bug (https://github.com/OGGM/oggm/issues/201)

oggm/params.cfg
jlandmann: Done!

oggm/utils.py
jlandmann: Good comment, my vertical line in PyCharm had a wrong default...

fmaussion:
OK, all green. Thanks!

PR #206, opened 2017-04-01
Several non-backwards compatible changes

fmaussion:


fmaussion:
we still have an issue with Appveyor. Merging anyway

Issue #199, opened 2017-03-24
Get rid of intermediate DEM files


fmaussion:
Currently, get_dem_file will do following:
- given a lon/lat range, list ALL the raw DEM files needed to cover the desired area (this can be quite a lot for large glaciers at high latitudes)
- merge this list of files into one (large) DEM file (mosaicking)

The logic is here: https://github.com/OGGM/oggm/blob/71e3164a40962011a433db03db0001c60cd8a83d/oggm/utils.py#L1318-L1348

This is quite inefficient: it requires unnecessary space, lots of memory, and could be done much better: rasterio should have the option to do mosaicking / subsetting on the go.

Contributions VERY welcome!

fmaussion:
@jlandmann do you think you can find time to look into this? This could be interesting for you too, and eventually live in the geoget package

Also, we could get rid of the intermediate zip files for the same reason.

jlandmann:
I looked at it some days ago already...but just to be sure: Do you just want to add a subsetting to the given lat/lon bounds before saving (in the OGGM workflow) or do you want to implement at function which includes merging and subsetting directly in rasterio?

It might take some time until I make it, put I can put it on my agenda. Right now I am trying to make geoget more independent from OGGM/oggm and get it ready for testing.

fmaussion:
Right now, ``get_topo_file`` returns a path to a DEM file which is guaranteed to cover the lon and lat extent given by the arguments. If more that one file is needed, it makes  a new mosaic file merging all the original ones.

Instead, it should just return a list of files.

Then, it will be the task of ``define_glacier_region`` to be able to deal with this list of files, by using rasterio instead of GDAL. 

This would be enough for OGGM, but ``define_glacier_region`` could become a very nice general function in ``geoget``. It's easier to do after it's done in OGGM though, cause we can realy on the test suite here.

jlandmann:
Okay, I got what you mean, now. As I said, I will put it on my agenda and you can assign me, if you want ;-)

fmaussion:
Closed via #216

PR #211, opened 2017-04-07
DEM download: LRU handling of unzipped files

fmaussion:
The downloaded files are now extracted in a temporary directory located in the working directory. Altogether, this will spare a lot of disk space at a minimal computing cost (some zipped files might be unpacked more than once).

A LRU handler is deleting these temporary files when they new ones are being unpacked. This is motivated by the fact that RGI regions and neighbouring glaciers are likely to have similar DEMs: thus this will  avoid unnecessary unpacking.

Also, I simplified some download logic.

The LRU part of the code is a bit convoluted, but I see this as a temporary solution.

CC @TimoRoth @jlandmann


jlandmann:
Very good, I just started working on the changes mentioned in #199, but apart from the Docstring in `get_topo_file` there should be no conflict...

fmaussion:
This should not brake existing code, merging now

PR #214, opened 2017-04-10
Various Updates for File-Downloads

TimoRoth:
Offer a global Cache-Directory for all Downloads.
This enables easy Download-Caching on Travis and HLRN.

Switches to using boto3 for downloads from S3. Calling a command-line utility is not pretty, specially when there is an official library for the job.
Nothing changed about setting credentials for it.

Get rid of the FileLock. Specially on Windows, it's not working too great, getting stuck every once in a while.
This now uses a SyncManager from multiprocessing, which should work everywhere.

### DIFF
--- a/oggm/cfg.py
+++ b/oggm/cfg.py
@@ -225,10 +225,6 @@ def initialize(file=None):
     global RGI_REG_NAMES
     global RGI_SUBREG_NAMES
 
-    # Make sure we have a proper cache dir
-    from oggm.utils import _download_oggm_files
-    _download_oggm_files()
-
     if file is None:
         file = os.path.join(os.path.abspath(os.path.dirname(__file__)),
                             'params.cfg')
@@ -252,6 +248,17 @@ def initialize(file=None):
         cp['cru_dir'] = os.path.join(homedir, 'OGGM_data', 'cru')
     if cp['rgi_dir'] == '~':
         cp['rgi_dir'] = os.path.join(homedir, 'OGGM_data', 'rgi')
+
+    # Setup Download-Cache-Dir
+    if os.environ.get('OGGM_DOWNLOAD_CACHE') is not None:
+        cp['dl_cache_dir'] = os.environ.get('OGGM_DOWNLOAD_CACHE')
+    PATHS['dl_cache_dir'] = cp['dl_cache_dir']
+    if PATHS['dl_cache_dir'] and not os.path.exists(PATHS['dl_cache_dir']):
+        os.makedirs(PATHS['dl_cache_dir'])
+
+    # Make sure we have a proper cache dir
+    from oggm.utils import _download_oggm_files
+    _download_oggm_files()
 
     # Parse RGI metadata
     _d = os.path.join(CACHE_DIR, 'oggm-sample-data-master', 'rgi_meta')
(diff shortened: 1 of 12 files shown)


TimoRoth:
I'm not sure why the AppVeyor build is not firing, this should fix them.

fmaussion:
Thanks!

If filelock is not needed anymore we should remove all references (also in the docs)

oggm/cfg.py
```diff
@@ -253,6 +249,17 @@ def initialize(file=None):
     if cp['rgi_dir'] == '~':
         cp['rgi_dir'] = os.path.join(homedir, 'OGGM_data', 'rgi')
 
+    # Setup Download-Cache-Dir
+    if os.environ.get('OGGM_DOWNLOAD_CACHE') is not None:
+        cp['dl_cache_dir'] = os.environ.get('OGGM_DOWNLOAD_CACHE')
+    PATHS['dl_cache_dir'] = cp['dl_cache_dir']
+    if PATHS['dl_cache_dir'] and not os.path.exists(PATHS['dl_cache_dir']):
+        os.makedirs(PATHS['dl_cache_dir'])
```
fmaussion: What happens when the cache dir is left empty in the config file?

TimoRoth: PATHS['dl_cache_dir'] is empty/false, and it won't enter that branch.

oggm/utils.py
```diff
@@ -79,32 +79,40 @@
...
+    p = None
     try:
-        return urlretrieve(url, ofile, *args, **kwargs)
+        cache_dir = cfg.PATHS['dl_cache_dir']
+        if cache_dir and os.path.isdir(cache_dir):
+            p = urlparse(url)
+            p = os.path.join(cache_dir, p.netloc, p.path[1:])
+            if not os.path.exists(os.path.dirname(p)):
+                os.makedirs(os.path.dirname(p))
+            # TODO: Maybe figure out a way to verify the integrity of the cached file?
```
fmaussion: Yes, this would be awesome! In particular, there are three possible scenarios:
- the file is downloaded properly
- the file download is aborted and the the file is corrupted. How to check for this?
- the file doesn't exist online (this is a legitimate request from some coastal glaciers that implie trying to download a file without knowing that it doesn't exist. Currently, the failing download is accepted as legit)

oggm/cfg.py
fmaussion: Yes, but I mean later, with the download

fmaussion:
Can I merge this now? Would like to test it out.

fmaussion:
Squash or regular merge? ;-)

oggm/cfg.py
TimoRoth: The download functions have the same kind of check, and they keep working like they do now if no cache dir is set.

TimoRoth:
Should be fine to merge regularly.
And yes, I have a few more ideas about additions to this, but for now 
this is fine as it is.

TimoRoth:
Should be fine to merge regularly.
And yes, I have a few more ideas about additions to this, but for now 
this is fine as it is.

Am 10.04.2017 um 17:08 schrieb Fabien Maussion:

fmaussion:
Thanks!

PR #215, opened 2017-04-11
Small fix for sample-download from Github

TimoRoth:
We don't want to cache the sample data download from github.
It's always the same url to master, so even if the sha changes, the resulting file wouldn't, as it's served from the cache.
To avoid this, download based on the commit-hash, avoiding cache if possible.


fmaussion:
Thanks! Feel free to merge!

fmaussion:
Merged it ;-)

PR #218, opened 2017-04-12
Further Updates for Download-Cache

TimoRoth:
This changes the semantics of the two download-functions.
They don't take a destination path anymore, but instead return the path where they decided to download the file to, which is either the specified download cache dir, or a path inside of the configured working dir.

If a function then wants to make changes to the downloaded file, it should make a copy and not work on it in-place.
So far, nothing in the entire code base does that, so no additional copying is necessary.

Also makes use of the Cache on Travis now, so using Downloads in test should not be too horribly anymore now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,10 @@ branches:  # blacklist
     - appveyor
     - docs
 
+cache:
+  directories:
+    - $HOME/dl_cache
+
 matrix:
   fast_finish: true
   include:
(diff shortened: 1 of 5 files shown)


fmaussion:
Thanks! This is too interleaved for me to review online. I think I need to merge and have a look locally. 

We need to start documenting this. Could you add a few docstrings more, also to the private funcs? No need to write too much, I'll go over it later and make sure I understand. I'll also add a chapter to the docs, as more and more people are using OGGM for regions all over the globe.

TimoRoth:
There wasn't a way to split that update up, as it would leave stuff in a broken state in between.
Added short docs to all the small helper functions.

PR #216, opened 2017-04-11
Get rid of intermediate DEM files

jlandmann:
Adresses #199

PR #219, opened 2017-04-17
Inversion now takes ice divides into account

fmaussion:
This PR is a bit larger than I wish it would be, and it brakes backwards compatibility. Non-exhaustive list of changes:
- the inversion formula takes ice divides into account (this changes the thickness estimation locally)
- the filtering that occurred during the inversion is now happening in a separate task. It has been improved to conserve volume entirely. The implications of this change aren't fully explored
- had to make some change to remove rasterio warnings and other issues related to https://github.com/OGGM/oggm/pull/216 . The differences between rasterio 1.0 and previous is a pain. (Travis still uses 0.36 and so did I).
- docs improvements

To get something close to backwards compatibility one should set ``cfg.PARAMS['invert_with_rectangular'] = False`` in the script. However I would recommend OGGM users NOT to update OGGM until the rest of the workflow can cope with rectangular beds.

### DIFF
--- a/docs/_code/prepare_climate.py
+++ b/docs/_code/prepare_climate.py
@@ -66,8 +66,7 @@ res = t_star_from_refmb(gdir, mbdf.ANNUAL_BALANCE)
 # For the mass flux
 majid = gdir.read_pickle('major_divide', div_id=0)
 cl = gdir.read_pickle('inversion_input', div_id=majid)[-1]
-mbmod = ConstantMassBalanceModel(gdir, optim_zminmax=[np.min(cl['hgt'])-50,
-                                                      np.max(cl['hgt'])+50])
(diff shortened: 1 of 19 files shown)


fmaussion:
@anoukvlug @bearecinos fyi see message above

bearecinos:
sorry I come back to this until now. several questions
1. Is in here now that the last pixels of tidewater as kept as they were without filtering and then restore after filtering? 
2. After running the vol_inversion task I need to run this new filtering. 
3. Is there any update on using a different bed shape for the las 5 in TW?

fmaussion:
For your purposes I would try to avoid filtering altogether. What would e good for you though is to force the last pixels of the inversion to be rectangular. This should happen in the ``prepare_for_inversion`` task

PR #221, opened 2017-04-19
Multiple DEM files bug

fmaussion:
This PR corrects a bug when multiple files are needed.

It also adds a level in the ``per_glacier`` directory tree, thus breaking backwards compatibility again


fmaussion:
cc @jlandmann

jlandmann:
Sorry for that! Next time I will directly introduce a dataset in the test directory and write some test...

fmaussion:
I don't mind having no tests for some parts of the code, but having a short comment like ``# untested`` would be nice!

Thanks again for your pr, the code looks much better now

PR #232, opened 2017-05-12
Small slopes filtering now happens at the flowline determination

fmaussion:
This relatively small change in the code involves quite a bit of change in how the inversion and the numerical model are going to handle unrealistic flowlines in the future. The consequences of this change need to be tested on other glaciers than HEF, but here is the deal: slopes smaller than ``cfg.PARAMS['min_slope']`` are simply not allowed anymore **at the moment the flowlines are created** (``initialize_flowlines`` task). 

When the threshold is reached, the non-valid areas are expanded until the slope is valid again using a fill-void algorithm. Two cases are possible:
- the voids can be filled. In that case the algorithm will simply interpolate between the two valid points. Example with paint:
paint
- the problem happens at the very beginning (e.g. HEF if you omit the divides). In that case the flowline is cut so that you have a valid slope.

**Before**:

test_nodivide

**After**:
 
test_nodivide_corrected

I expect this change to increase the numerical stability of the model, and automatically avoid completely unrealistic flowlines. It will need some thorough testing on more glaciers to see if it works as expected.

### DIFF
--- a/docs/notebooks/getting_started.ipynb
+++ b/docs/notebooks/getting_started.ipynb
@@ -121,9 +121,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "collapsed": true
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "rgi_shp.plot();"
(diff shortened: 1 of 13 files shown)


fmaussion:
I added a control parameter to revert to the previous behavior if needed (``cfg.PARAMS['filter_min_slope']``, which is ``True`` per default. Set to ``False`` to get back to previous)

PR #234, opened 2017-05-15
Changes to the calving function

bearecinos:
I also added the RGI'ID of the glaciers in the McNabb data base

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,7 @@ install:
     cat <<EOF >test.sh
     set -x
     set -e
+    pip3 install --upgrade git+https://github.com/fmaussion/salem.git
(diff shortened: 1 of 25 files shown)


fmaussion:
Great, thanks! Would you like feedback on this or can I merge right away?

bearecinos:
Well if you have time, It will be good..to double-check that the calving function and the "theoretical calving" is what we agreed when you were here. I haven't change that just some things at the top that is now commented (new ID's to run)

fmaussion:
sorry it took me so long! I've looked at the function and it looks like what we agreed yes! But this doesn't mean it should be written in stone. Don't be afraid of improving it further!

PR #236, opened 2017-06-02
Improvements in preprocessing

fmaussion:
This PR implies several major improvements on several preprocessing steps. I'll make more detailed comments before merging, this is still a WIP.


fmaussion:
Here a non-comprehensive list of the changes involved:
- the downstream lines and their integration with glacier divides is now completely different. (this is the major change of this PR). This was necessary to be able to cope with very complex divide geometries like ice caps. Now the downstream lines can go in all directions, in which case they meet their tributary outside of the domain (see plot below)
- the parabolic bedshape computed from the real topography is now included to the model chain (finally). Again in the example below, one can see that the glacier width adapts to the surrounding topography
- the ``initialize_flowline`` task now filters for small and negative slopes. This was introduced by https://github.com/OGGM/oggm/pull/232 but is now used and tested by default
- The inversion takes rectangular beds into account. this happens when the flowline grid point normals touch a: (i) a neighbor glacier (ii) a neighbor divide (iii) a neighbor flowline catchment
- The ``init_present_time`` tasks now takes all this into account (this was the major bunch of work - quite boring but necessary). The default is now to produce a ``MixedFlowline`` in all cases, which is a mix of parabolic, rectangular and trapezoidal shapes.
- The model flowlines are better tested. Their constructors and documentation are still quite bad though
- I added some examples of random glacier evolution in the getting started notebook

I will do a little bit of cleaning a a few tests before merging.

test_chhota_shigri

PR #242, opened 2017-06-14
Changes in run_alaska.py and utils.py to study the glacier front width problem 

bearecinos:
- in utils.py I added the width of the calving front to the galcier_characteristics.csv file. 
- in run_alaska.py a lot has changed. **not the main way that the calving is calculated** but just adapted the loop so it can read the McNabb data base and fix the width for the calving glaciers inside the data base. Calving glaciers without a width takes oggm output.


oggm/utils.py
```diff
@@ -1709,13 +1709,18 @@ def glacier_characteristics(gdirs, to_csv=True):
...
             for i in gdir.divide_ids:
                 cl = gdir.read_pickle('calving_output', div_id=i)
                 for c in cl:
                     all_calving_data = c['calving_fluxes'][-1]
+                    all_width = c['t_width']
             d['calving_flux'] = all_calving_data
+            d['t_width'] = all_width
         else:
             d['calving_flux'] = 0
+            d['t_width'] = 0
```
fmaussion: Can your put a ``np.NaN`` here instead oz zero? Also, ``'t_width'`` is maybe not the best name. how about ``'calving_front_width'`` ?

bearecinos: yes I will do that I am also making all the lines fit in the margins :) ... I shall commit soon

bearecinos: Ok, I have made the changes... let me know if there is anything more. Cheers

fmaussion:
Thanks!

PR #245, opened 2017-06-23
Fix rasterio deprecation warnings

TimoRoth:
See https://github.com/mapbox/rasterio/issues/86 for reference.
Fixes a couple hundred deprecation warnings that come up during testing.

### DIFF
--- a/oggm/core/preprocessing/gis.py
+++ b/oggm/core/preprocessing/gis.py
@@ -409,7 +409,7 @@ def define_glacier_region(gdir, entity=None):
     if len(dem_list) == 1:
         dem_dss = [rasterio.open(dem_list[0])]  # if one tile, just open it
         dem_data = rasterio.band(dem_dss[0], 1)
-        src_transform = dem_dss[0].transform
+        src_transform = dem_dss[0].affine
     else:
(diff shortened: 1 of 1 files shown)


fmaussion:
Thanks!

fmaussion:
With rasterio > 1.0 this change is not desired. I get plenty of warnings:

```
/home/mowglie/disk/Dropbox/HomeDocs/git/oggm-fork/oggm/utils.py:1802: DeprecationWarning: 'src.affine' is deprecated.  Please switch to 'src.transform'. See https://github.com/mapbox/rasterio/issues/86 for details.
```

Will revert in an upcoming PR, or add a check for the rasterio version

TimoRoth:
Rasterio 1.0 is not released yet, it's in alpha. So this should not be showing up normally. At least pip does not install rasterio 1.0 unless explicitly requesting pre-release versions.

This deprecation back and forth and renaming is a bit odd of them. The only way to consistently not get warnings seems to be to use a version check...

fmaussion:
Yes: https://mapbox.github.io/rasterio/topics/migrating-to-v1.html

fmaussion:
Will add a check for rasterio, but I favor 1.0 on my environments - the alpha has been around quite a while now, I wonder when they will release it

PR #246, opened 2017-06-24
Render mpl test images without text

TimoRoth:
Can't really think of any automatic approach for this.
matplotlib can even come with its own statically linked version, so checking the system version is pointless.

### DIFF
--- a/oggm/utils.py
+++ b/oggm/utils.py
@@ -410,7 +410,7 @@ def _download_oggm_files_unlocked():
             if filename in out:
                 # This was a stupid thing, and should not happen
                 # TODO: duplicates in sample data...
-                k = os.path.join(os.path.dirname(root), filename)
+                k = os.path.join(os.path.basename(root), filename)
                 assert k not in out
                 out[k] = os.path.join(root, filename)
             else:


fmaussion:
I'm not very enthusiastic about this. When we add a new test we'll have to generate the baselines for the two versions, which seems complicated.

I see another way out: we could also generate the images without text (https://github.com/matplotlib/pytest-mpl#removing-text) and put them in a ``no-text`` baseline directory. When required, the tests should then compare without text.

TimoRoth:
The weird thing about this is that this change happened with freetype 2.7, which is around for quite a while. Ubuntu just never bothered to update from 2.6.3. Even 17.10 will still have 2.6.3.

Just not rendering test seems a good solution to me. Should the no-text version be the default then? It's not really possible to detect which version of freetype is being used.

TimoRoth:
Hm, this is bad. Even though the text is gone, the freetype version still changes the layout of things, breaking a lot of the tests. This is a horrible mess, and I'm not sure how to approach this. Freetype 2.6.3 is ancient by now and should really not be around anymore, but Ubuntu keeps it for some reason.

TimoRoth:
I need to think about this some more, just merging the trivial duplicate-file sample-data fix for now.

fmaussion:
Maybe we should see if the mpl folks are aware of the problem.

Could we avoid using freetype altogether, by selecting another font for the plots?

TimoRoth:
> Maybe we should see if the mpl folks are aware of the problem.

https://github.com/matplotlib/matplotlib/issues/8796

> Could we avoid using freetype altogether, by selecting another font for 
> the plots?

freetype is the font rendering library, not a specific font. I don't 
think there is a way to avoid using it.
The mpl people mentioned using vector graphics instead of pngs, that 
would indeed solve the issue, as there is no font rendering involved in 
creating those. But I can't see how to use them with pytest-mpl

PR #247, opened 2017-06-25
Use different baselines images depending on freetype version

TimoRoth:
Updating the images is not too complicated, on a system with a more recent freetype it's just doing

pip install --no-deps -I --no-binary :all: matplotlib
vs.
pip install --no-deps -I matplotlib

As the pre-built mpl comes with freetype 2.6.1, a self built one uses the system version.

Additionally this contains a small change for the workflow CPU count, now taking into account a potential Slurm CPU allocation.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,7 +13,7 @@ notifications:
 environment:
   matrix:
     - PYTHON: "C:\\Python3-conda64"
-      PYTHON_VERSION: "3.5"
+      PYTHON_VERSION: "3.6"
       PYTHON_ARCH: "64"
 
 install:
@@ -22,6 +22,7 @@ install:
 
   # Create environment with desired python version and activate it
   - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
+  - "SET GDAL_DATA=%PYTHON%\\Library\\share\\gdal"
   - "conda install --yes --quiet python=%PYTHON_VERSION%"
 
   # Check that we have the expected version and architecture for Python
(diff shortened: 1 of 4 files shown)


fmaussion:
Thanks for looking into this! I'm still not sure that having different baselines is a good idea. If you do this for the freetype version, why not doing it for the gdal version also? (we should check this, but I still think that the conda gdal output differs from a pip install which uses the system lib -- or are there manylinux wheels for rasterio/fiona now?)

So after reading the mpl [thread](https://github.com/matplotlib/matplotlib/issues/8796) there is no best way out, but we have a few possibilities:
- treat the travis builds as the "blessed" version and test with zero tolerance on Travis, with higher tolerance (or even skipping tests) elsewhere. This would imply that we have access to Travis' outputs (https://github.com/matplotlib/pytest-mpl#test-failure-example)
- have a higher tolerance everywhere (this is what we've done until now)
- have different baseline versions for all kind of setups, and lower tolerance everywhere

I don't really have a preference, although I think that option 3 is a bit overkill. The image tests are mostly qualitative, and are a way to check if the model more or less produces a reasonable output.

TimoRoth:
The appveyor checks are running on conda, and work just fine.
Conda/conda-forge are both up to date with gdal by now.
I don't think there is a need for old-gdal baseline images anymore, as even Ubuntu 16.04 comes with gdal 1.11, which is recent enough to produce compatible results.

The problem with this freetype situation is that there is no easy way to work around it. If your system has 2.7 or more recent, there's not much you can do about it. Disabling the graphics-based tests would be an option, but just having compatible baseline images really doesn't seem a lot of work to me.

The needed high tolerance for the freetype change would make the tests entirely useless. Pretty much every pixel of the graph changes, due to it being shifted upwards by a few pixels.

I'd prefer the multi-baseline approach, as implemented by this PR. The baselines don't change very frequently, and if they do, it's really not that much work to generate the two needed versions.

fmaussion:
OK, let's go for it then. Feel free to merge!

PR #249, opened 2017-07-01
Add some workflow tools and new defaults

fmaussion:
PR summary:
- the new default for the choice of the map's dx is now "squared". Dxmax is 200m (previously 100m). Large glaciers are easier (and faster) to process this way (Huss & Hock 2015 use 250m). However, the 200m limit is now reached for very large glaciers only.
- changes the glacier directory log directory to be a logfile instead
- it is now possible to ask the status of a task at run time, and therfore skip a task that has been run successfully. This is still experimental, as there is a risk that things go badly when using this option (I had to learn it the hard way)
- the numerical model's CFL criterias changed a bit
- the way downstream lines are computed change a bit (more robust for flat terrain, maybe a bit less for some cases)
- add a sandbox example for all glacier worldwide with MB


fmaussion:
@OGGM/oggm-devs this change contains (again) default parameter changes - this is for the best but will (again) change your results. 

See upcoming github issue for more details about upcoming changes

PR #251, opened 2017-07-04
New mu-check, more units in climate file, PEP8

jlandmann:
As agreed on the phone earlier, I introduced a new mu-check that check only if there is one mu existing.

Furthermore I put two small changes: I"allowed" more unit descriptions in the climate netcdf (even though this work never ends) and did some PEP8 corrections.


fmaussion:
Great, thanks!

PR #252, opened 2017-07-07
Small bugfix in BEGINSEC_IN_MONTHS

jlandmann:
The begin of the month was shifted by +/- one day per month in the old version.
It did not influence the monthly calculations so far, however, for possible future changes it could be relevant.

Regarding the way of counting: as we start counting with zero, it would be more logical for me that the begin second of February, e.g., was 2678400 and not 2678401. This would break the tests though.

### DIFF
--- a/oggm/cfg.py
+++ b/oggm/cfg.py
@@ -86,7 +86,7 @@ SEC_IN_HOUR = 3600
 DAYS_IN_MONTH = [31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31]
 SEC_IN_MONTHS = [d * SEC_IN_DAY for d in DAYS_IN_MONTH]
 CUMSEC_IN_MONTHS = np.cumsum(SEC_IN_MONTHS)
-BEGINSEC_IN_MONTHS = np.cumsum([0] + [(d + 1) * SEC_IN_DAY for d in DAYS_IN_MONTH[:-1]])
+BEGINSEC_IN_MONTHS = np.insert(CUMSEC_IN_MONTHS[:-1] + 1, [0], 0)
 
 RHO = 900.  # ice density
 G = 9.81  # gravity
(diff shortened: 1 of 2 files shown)


fmaussion:
> This would break the tests though.

Why?

jlandmann:
I knew you would ask ;-)

The (first or only) test that causes issues is this one: 
https://github.com/OGGM/oggm/blob/master/oggm/tests/test_utils.py#L101

and I guess it's because of the following (partly numerical!?) issue:
In `year_to_date` we have the following values at [line 972](https://github.com/OGGM/oggm/blob/master/oggm/utils.py#L972)  (1=we do add one second at the `BEGINSEC_IN_MONTH`, 2= we do not add one second in `BEGINSEC_IN_MONTH`)

`CUMSEC_IN_MONTHS = array([ 2678400,  5097600,  7776000, 10368000, 13046400, 15638400,
       18316800, 20995200, 23587200, 26265600, 28857600, 31536000], dtype=int32)`

1) sec =13046401.00000275  => out_m becomes 6, because it's less/equal than the 5th entry in CUMSEC_IN_MONTHS
2) sec = 13046399.999996582 => out_m becomes 5 (WRONG!), because it's already less/equal than the 4th entry in CUMSEC_IN_MONTHS

So if you introduce two further, small changes (see commit) it works: you have to round `sec` to the third digit and make the "less equal" [here](https://github.com/OGGM/oggm/blob/08d1cb835c28206d66999905d3b411df245c11fa/oggm/utils.py#L972) a "less", which would also be consistent with the idea that second 2678400 belongs already to February, for example.

fmaussion:
OK! Definitely not the best code I've ever written but whatever, as long as it works...

Thanks!

fmaussion:
Ah: what happens if you round to an integer btw? (instead of third digit). Since we don't need subsecond accuracy...

fmaussion:
OK i took the liberty to modify your commit a bit to see what happens

jlandmann:
This should also work!

PR #257, opened 2017-07-10
Some bugfixes due to the new map settings

fmaussion:
closes https://github.com/OGGM/oggm/issues/256

cc @TimoRoth


fmaussion:
I also added a real dynamical run to the benchmark

PR #270, opened 2017-07-21
Avoid very flat parabolas in downstream

fmaussion:
This PR has two purposes:
- simplify automated parsing of error messages by avoiding ":" in prints
- limit the values of parabolas obtained with the downstream algorithm.

Point 2 is very important: for extremely rare cases (e.g. one single glacier in Alaska), the values where so flat that the volume of the glacier could become huge.


fmaussion:
@anoukvlug I strongly recommend to update OGGM after this change is merged, this was quite a bad bug.

PR #273, opened 2017-08-04
run for gcm

anoukvlug:
This is a pull request for the function run_from_gcm. The function is for a large part a copy of random_glacier_evolution function. The function can be used to run the glaciers with a the climate from for instance a CESM simulation.

I could later include a 'spin-up' option in the function, but I would first like to discuss the basis.

### DIFF
--- a/oggm/core/models/flowline.py
+++ b/oggm/core/models/flowline.py
@@ -1653,3 +1653,63 @@ def iterative_initial_glacier_search(gdir, y0=None, init_bias=0., rtol=0.005,
         _ = past_model.run_until_and_store(y1, path=path)
     else:
         past_model.to_netcdf(path)
+
+@entity_task(log)
+def run_from_gcm(gdir, filename='cesm_data', filesuffix='', **kwargs):
+    """ Runs glacier with climate input from a general circulation model.
+
+     Parameters
+     ----------
+     y0 : is the start year of the model run
+     y1 : is the end year of the model run
+     filename: name of the climate file. (Can be made with the
+     process_cesm_data function)
+     filesuffix : for the input & output file
+     kwargs : kwargs to pass to the FluxBasedModel instance
+     """
+
+    if cfg.PARAMS['use_optimized_inversion_params']:
+        d = gdir.read_pickle('inversion_params')
+        fs = d['fs']
+        glen_a = d['glen_a']
+    else:
+        fs = cfg.PARAMS['flowline_fs']
+        glen_a = cfg.PARAMS['flowline_glen_a']
+
+    kwargs.setdefault('fs', fs)
+    kwargs.setdefault('glen_a', glen_a)
+
+    y0 = cfg.PARAMS['y0']
+    ye = cfg.PARAMS['y1']
+
+    mb_model = oggm.core.models.massbalance.\
+        PastMassBalanceModel(gdir, filename=filename, filesuffix=filesuffix)
+
+    # run
+    run_path = gdir.get_filepath('model_run', filesuffix=filesuffix,
+                                 delete=True)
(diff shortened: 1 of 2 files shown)


oggm/core/models/flowline.py
```diff
@@ -1653,3 +1653,63 @@ def iterative_initial_glacier_search(gdir, y0=None, init_bias=0., rtol=0.005,
...
+        glen_a = d['glen_a']
+    else:
+        fs = cfg.PARAMS['flowline_fs']
+        glen_a = cfg.PARAMS['flowline_glen_a']
+
+    kwargs.setdefault('fs', fs)
+    kwargs.setdefault('glen_a', glen_a)
+
+    y0 = cfg.PARAMS['y0']
+    ye = cfg.PARAMS['y1']
```
fmaussion: aha I see. Can you also change the name of the params to ``ys``, ``ye`` in the ``params.cfg`` file?

You can keep the idea of having them as keyword arguments in the function and use the params as default if the user didn't set them.

oggm/core/models/flowline.py
```diff
@@ -1653,3 +1653,63 @@ def iterative_initial_glacier_search(gdir, y0=None, init_bias=0., rtol=0.005,
...
+
+    mb_model = oggm.core.models.massbalance.\
+        PastMassBalanceModel(gdir, filename=filename, filesuffix=filesuffix)
```
fmaussion: use ``mbmods.PastMassBalanceModel`` instead

oggm/core/models/flowline.py
```diff
@@ -1653,3 +1653,63 @@ def iterative_initial_glacier_search(gdir, y0=None, init_bias=0., rtol=0.005,
...
+                               **kwargs)
+        try:
+            model.run_until_and_store(ye, run_path=run_path,
+                                      diag_path=diag_path)
+        except RuntimeError:
+            if step == 'ultra-conservative':
+                raise RuntimeError(
+                    '%s: we did our best, the model is still '
+                    'unstable.', gdir.rgi_id)
+            continue
```
fmaussion: This logic slitghly changed in the most OGGM version. See https://github.com/OGGM/oggm/blob/master/oggm/core/models/flowline.py#L1602-L1605 and copy from there

oggm/core/models/flowline.py
anoukvlug: I am not sure if I understand what you are suggesting. Do you mean that I add 
```
ys = y0
ye = y1
```
to the cfg.params?

anoukvlug:
Thank you for the feedback! 

I have also looked into the problem with the process_cesm_data function, but I didn't found a way to fix it yet. So that will probably need to wait a bit, as I will be out of office the next 2 weeks. 
While looking for a solution I found the following pull request at xarray: https://github.com/pydata/xarray/pull/1252. This might make handling data outside the 1677-2262 range much more easy in the future :)

fmaussion:
I have to leave on holidays now, so I'd suggest that either we merge this now or it waits until sept.

anoukvlug:
If you have to leave right this moment I wish you a very nice holiday! The PR can wait till September.

fmaussion:
@anoukvlug do you have time to work on this in the next days? Otherwise I might merge this soon and make the changes myself later.

anoukvlug:
No, I won't have time to work on this till the end of the ArcTrain Annual Meeting. So the earliest day I could start working on it will be the 25th of September.

fmaussion:
OK, no problem! I can start from here

fmaussion:
Merging this now, will add my modifs and tests in a subsequent PR.

Thanks!

PR #277, opened 2017-09-07
Add a test to see if RGI IDs can be extended with suffixes (they can)

fmaussion:
This will be useful when running divides as single glaciers.


fmaussion:
Mh. The tests won't trigger on this PR. I wonder why?

fmaussion:
Ah ok. it just needed some more time ;)

PR #279, opened 2017-09-08
Small refactoring of tests

fmaussion:
This PR is mostly to reduce the running time of the test suite. It splits the model tests into two modules for dummy parallel processing on Travis  and skips two tests which took much time without real benefits.


fmaussion:
OK, the code coverage doesn't check for tests code coverage anymore, which is actually a good thing.

The bad thing is that this significantly reduced our overall test coverage score, but we simply have to admit that the truth is hard to accept ;-)

We can work towards a higher coverage, but only at cost: a lot of the code can't easily be tested (graphics, workflow and downloads are the most poorly tested aspects of the code, while the model internals are very well tested!)

PR #281, opened 2017-09-10
Add a test for CESM data and xarray bug workaround

fmaussion:
This finally adds tests for the process_cesm_data task, and also addresses an issue with the latest xarray version (https://github.com/pydata/xarray/issues/1565)


fmaussion:
@anoukvlug : this should change your results (if it does it's a mistake)

PR #305, opened 2017-09-29
Reset multiprocessing in workflow tests

TimoRoth:
This fixes the workflow tests when running with multiprocessing.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+oggm/tests/funcs.py ident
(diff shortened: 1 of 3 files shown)


fmaussion:
Great, thanks! Do you think we could remove this:

```
https://github.com/OGGM/oggm/blob/master/oggm/tests/test_workflow.py#L62
```
also?

fmaussion:
@TimoRoth what do you think about using multiprocessing on travis as well?

fmaussion:
Nevermind, will try in a separate PR.

Thanks!

PR #310, opened 2017-10-01
Run workflow tests both with and without multiprocessing

TimoRoth:
This should make coverage happy and test both code paths, which might be a good idea.


fmaussion:
Thanks! Indeed this is probably the best solution.

PR #311, opened 2017-10-01
Add ELA to mb models and diagnostics

fmaussion:
This adds a get_ela() function to the mass-balance models. It uses a minimization function to find it, i.e. it is not suuuuper fast but it's OK. The ELA is now added to the model diagnostic files too.

cc @anoukvlug


anoukvlug:
Great that you added the ELA!

Issue #250, opened 2017-07-03
The problem with glacier divides


fmaussion:
This is a thing that bothers me since a long time.

OGGM currently has a bunch of logic to deal with "glacier divides" as explained in the doc: http://oggm.readthedocs.io/en/latest/glacierdir-gen.html#glacier-divides . The motivation for this method was (unfortunately) mostly motivated by Hintereisferner (HEF), which is the perfect example of where it "works". The arguments in favor for divides were (and still are):
- the RGI entity is the "holy grail" which ensure comparisons across runs and models
- this also holds true for the reference mass balance: we always assumed that the WGMS mass-balance timeseries are valid for the whole HEF as it is defined by RGI
- it makes HEF (and many other glaciers) look great: HEF can grow to become a unified glacier again, as it used to be... Example : 

![selection_046](https://user-images.githubusercontent.com/10050469/27800209-9dcd89fe-6018-11e7-83a6-73de8d0a5825.png)

The problem is that in practice, it increased code complexity a great deal without much gain. The reason is that most glaciers that need divides can be considered as *really* different. The best examples are undivided ice-caps (each entity can and should be be considered alone), or see for example the Chhota Shigri glacier:
 
![test_chhota_shigri](https://user-images.githubusercontent.com/10050469/27953526-70b6fe5e-630c-11e7-93f3-850853a87024.png)

I see absolutely no reason why these two are put together in RGI. To summarize the arguments against divides:
- the increase code complexity is very large: each divide has a downstream line, all the lines need to be merged to form one glacier in the end, and sometimes they have to merge "outside" of the domain like Chhota Shigri above.
- many files are duplicated, the glacier directories are harder to understand
- the spatial resolution chosen for the entire glacier is often not suitable for the smaller divides. If the divides was alone it would get a more meaningful dx
- in fact, each divide should have it's own temperature sensitivity mu*. Currently, we *have* to have different mu* for the inversion (for mass conservation reasons) but the we use one single mu for the glacier. With the example HEF, this means that the smaller outlet glaciers will always grow to join the larger one because of changing mus
- the same argument holds for any divide: if it is divided, then it has a different dynamic and a different mu

So right now, I am tending to **suppress the logic about divides as a whole**. In order not to rush into this we can keep the code as it is (since it works with a single divide!), and see if we still think so in a couple of months.

[...]

PR #324, opened 2017-10-17
WIP: remove the concept of divides from the codebase

fmaussion:
Closes https://github.com/OGGM/oggm/issues/250

The reason for pushing this forward earlier than planned is that this complexity overhead hindered a relatively simple fix to the important "negative flux" problem.

This is a *major* simplification of the codebase, but is still likely to cause some trouble for a while.


fmaussion:
> is this going to change the file distribution in each glacier directories?

yes it is. No more `divides_XX` directories, but the filename will remain same.

Altogether this is a not-backward compatible change again (i.e. you'll have to re-run the preprocessing again eventually), but this shouldn't change the result at all.

juliaeis:
How will we than dealing with the divides? I ask cause to the partitioning algorithm. Because I need to change it, as it create the outlines.shp at each subdirectory "divide_XX".

fmaussion:
> How will we than dealing with the divides?

The divides will be defined *upstream* of OGGM and (if needed) the data will be merged back at post-processing. The reason is that divides are nothing else as a standalone glacier, and should not be necessary if the RGI was perfect (see also https://github.com/OGGM/oggm/issues/250).

For you, this means that we will extend your tool a little so that the data is prepared for OGGM. Say that we have the following situation: we have two glaciers (RGI50-03.00001 and RGI50-03.00002), one of them needs to be divided. We will now feed OGGM with a "modified" RGI file with entities RGI50-03.00001, RGI50-03.00002_d01, RGI50-03.00002_d02. We can discuss the details of the creation of this file on slack.

TimoRoth:
There was a system update of the Docker-Images that seems to have caused 
these new failures. I'm not sure if it's broken, or uncovered a bug.

fmaussion:
> There was a system update of the Docker-Images

Thanks for the info. The tests pass here locally (except the image tests of course), I'll investigate :-(

fmaussion:
OK, I can't reproduce the remaining failures. I have everything up to date here except GDAL, so either it's a GDAL thing again, or it's some ghost.

fmaussion:
OK, I'm going to merge this now : we'll have to get back to these failing tests later, but first I'd like to see if the tests pass on the cluster.

Results  of the PR: 700 lines of code less and a decrease in coverage but a bunch of clumsy logic removed. The change is not backward compatible at all, i.e. all your old glacier directories will be broken if you update.

@OGGM/oggm-devs I'd recommend to wait a bit longer before updating, since I am on a refactoring mood I will probably tackle other changes in the next days.

fmaussion:
Ah, two other collateral changes happened:
- the graphics functions for regional plots are no more, they are replaced by the "normal" plot functions which now all accept a list of gdirs as argument
- the `min_shape` parameter is back to a smaller value (1e-4 instead of 1e-3). It felt better at HEF, no clue what the implications globally will be.

TimoRoth:
I just re-ran the tests in a Docker-Container that was already tested 
working a few days ago, and the tests are failing in there as well 
now(with OGGM master, prior to this PR).

So something that OGGM downloads must have changed, as the very same OS 
image that used to work just fine is now horribly broken.

fmaussion:
> So something that OGGM downloads must have changed

This is so bad. That means that I should be able to reproduce locally if I delete my download cache, right?

But I actually don't know what this could be, none of the tests relies on external downloads (everything is in the oggm-data repo...)

fmaussion:
> So something that OGGM downloads must have changed, as the very same OS image that used to work just fine is now horribly broken.

Ah wait, I changed the oggm-sample files! This could explain that

TimoRoth:
I'd prefer if you don't merge the workarounds, it would make testing for what broke very hard.

fmaussion:
> I'd prefer if you don't merge the workarounds, it would make testing for what broke very hard.

Yes that's a valid point let's see how it goes

fmaussion:
But also, I'm not sure how important these tests really are -- it could well be due to a difference in GDAL or something. I'll keep the tests failing and maybe we can make them less strict later

TimoRoth:
I think I managed to get a lot of the tests to pass by actually updating GDAL(And the entire OS from Ubuntu 17.04 to 17.10, which includes GDAL). It's just one single test that's still failing now for me.

fmaussion:
Which one?

TimoRoth:
TestOtherDivides.test_define_divides:

```
________________________________________________________________________________ TestOtherDivides.test_define_divides ________________________________________________________________________________

self = <oggm.tests.test_models.TestOtherDivides testMethod=test_define_divides>

    def test_define_divides(self):

        from oggm.core.preprocessing import (gis, centerlines, geometry,
                                             climate, inversion)
        from oggm import GlacierDirectory
        import geopandas as gpd

        hef_file = utils.get_demo_file('rgi_oetztal.shp')
        rgidf = gpd.GeoDataFrame.from_file(hef_file)

        # This is another glacier with divides
>       entity = rgidf.loc[rgidf.RGIId == 'RGI50-11.00719'].iloc[0]

usr/local/lib/python3.6/dist-packages/oggm/tests/test_models.py:180:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
usr/local/lib/python3.6/dist-packages/pandas/core/indexing.py:1328: in __getitem__
    return self._getitem_axis(key, axis=0)
usr/local/lib/python3.6/dist-packages/pandas/core/indexing.py:1749: in _getitem_axis
    self._is_valid_integer(key, axis)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <pandas.core.indexing._iLocIndexer object at 0x7f6353c6b940>, key = 0, axis = 0

    def _is_valid_integer(self, key, axis):
        # return a boolean if we have a valid integer indexer

        ax = self.obj._get_axis(axis)
        l = len(ax)
        if key >= l or key < -l:
>           raise IndexError("single positional indexer is out-of-bounds")
E           IndexError: single positional indexer is out-of-bounds

usr/local/lib/python3.6/dist-packages/pandas/core/indexing.py:1638: IndexError
```

fmaussion:
Yes this line should read `entity = rgidf.loc[rgidf.RGIId == 'RGI50-11.00719_d01'].iloc[0]` in my latest commit here

TimoRoth:
So lets see how this run of tests goes. Maybe doing the Ubuntu/GDAL update was all it took.

fmaussion:
> So lets see how this run of tests goes. 

Still red, but I'll do a couple more change and retrigger later

fmaussion:
OK so the graphics test are now also failing, all this indicating changes in GDAL. I will soon get a new computer where I'll install the latest GDAL too. The failing tests are not so bad, we're fine

fmaussion:
Merging now, OK @TimoRoth  ?

TimoRoth:
The failing graphics tests are most likely because the new Ubuntu 17.10 updated freetype 2.8, and you only updated the baselines for <2.8?

fmaussion:
Yes could be :-(

TimoRoth:
Do I just clone this PR, and run the baseline generation on a system with 2.8 to get the right images?

fmaussion:
Yes! I would be super interested in seing the differences (there are probably differences in the maps due to GDAL too)

TimoRoth:
Some random thought I just had, wouldn't it be better to pin OGGM to a specific commit of the oggm-sample-data repo? That way updating the baselines there won't break tests for older versions of OGGM anymore.

fmaussion:
> Some random thought I just had, wouldn't it be better to pin OGGM to a specific commit of the oggm-sample-data repo? 

Yes! The current way we do this is painful and wrong

TimoRoth:
Pushed the updated baselines:
https://github.com/OGGM/oggm-sample-data/commit/40820ba389dff20ad754701bcb0b95ff5c39a99f

Re-Started the Build for this PR on Travis.

fmaussion:
Thanks! So as you can see the difference are indeed in the plots, but also in the underlying topography data. Look for example the difference of the histogram in both plots (hard to see but there are some):

test_modelsection
test_modelsection_alt

So there are two possibilities:
- either GDAL changes something in the topography interpolation between versions (most likely, but totally crazy and bad)
- or we use GDAL the wrong way (also likely, but where?)

In both cases this is a pain in the *** ...

TimoRoth:
I generated all 2.8 baselines images on the cluster, which has so far never updated GDAL. It's on 2.2.0 from the beginning on.

fmaussion:
> which has so far never updated GDAL

Yes I suspect this problem was here for long time but was hidden because I had very loose tolerance for the images.

Will loosen the tests a little bit so that they pass and move on, it's OK for now

PR #332, opened 2017-10-22
Inversion params are not optimised per default any more

fmaussion:


fmaussion:
I also added some utility functions:
- start the model from any use-defined initial state
- read leclercq data

Issue #336, opened 2017-10-26
Transitioning to RGIv6 - WGMS links


fmaussion:
I am in the process of making OGGM RGIV6 compatible.  Everything is quite trivial to do, the hardest bit is without a doubt the links to other databases.

I think I will from now on use the link list provided by the RGI itself (https://www.glims.org/RGI/rgi60_files/00_rgi60_links.zip) instead of the WGMS one. The RGI one seems to be updated, and it's consistent with what other modelers will do.

The only drawback is that we had 263 glaciers in the list that we got from WGMS last year (including the job done by @jlandmann), and there are only 232 in the RGI one (we loose 31 glaciers). I know I've asked this already, but it would be good to have it written down somewhere: why are these lists different again?

jlandmann:
Hi Fabien,
first of all sorry for being late in correspondence in general...i'm busy with paperwork issues since ages...

Regarding the issue, I looked up the last mail correspondence with Graham Cogley in my mail folder: it said that he wanted to finalize RGI linkages and then look at the WGMS list to see what he can add from that list, but without delaying the RGI6.0 release. So either he did it and did not approve 31 links or the WGMS list is simply not included in the new RGI link list.

jlandmann:
I started with differences between the table on the OGGM data web site to the new RGI link table (ensuring that we only look at "safe" links). I found 94 FOG IDs that don't occur in the RGI link list. Okay, to be honest, 5 of them did not end up in the WGMS list either (for whatever reason, all automated links should be overwritten with the manual ones). However, the 89 remaining ones didn't make it in the RGI list. I can write a mail to Graham on the weekend an ask how far he got.

fmaussion:
Thanks Johannes!

>  the table on the OGGM data web site

You mean this one, right? This is what we use: 
https://github.com/OGGM/oggm-sample-data/blob/master/wgms/rgi_wgms_links_20170217_RGIV5.csv

> I can write a mail to Graham on the weekend an ask how far he got.

First. I'll try to make an RGI6 list based on ours by matching the glaciers between the two versions and get back to you.

fmaussion:
First thing I tried is to see how many glaciers are similar in the two lists (based on their WGMS-FoG id): we come to "only" 169 glaciers where I thought that we would have  232 (the entire RGI list)...

I give up, this is all way too tedious: I'll use the RGI list for now.

This is what I said to Graham and Michael long ago: having two lists is more confusing than helping.

[...]

PR #338, opened 2017-10-28
Add RGI6 tools

fmaussion:


fmaussion:
OK this is ready to go. This has some nice new features such as:
- the possibility to choose between RGI5 and 6 at run time (5 is still the default)
- the use of pre-calibrated tstars is now the default behavior and is automated  
- update CRU to use TS4.1
- the docs for the runs have been improved 

The default is still RGIv5 because of the problem of the WGMS links (see https://github.com/OGGM/oggm/issues/336). RGI6 currently uses 177 glaciers, which is much less than the 255 for RGI5...

PR #342, opened 2017-11-02
Add a filesuffix to inversion output

fmaussion:
It is now possible to write the inversion output in a separate file and then read it back like this:

```python
tasks.volume_inversion(gdir, filesuffix='my_experiment')
inv = gdir.read_pickle('inversion_output', filesuffix='my_experiment')
```
Would this suit your need @bearecinos ?


bearecinos:
Yes! Awesome fabi !

fmaussion:
OK, let me know if something is missing

PR #345, opened 2017-11-07
Automatically reset multiprocessing on cfg changes

### LINKED ISSUE #327: Reset multiprocessing pool
The fact that the runs params are stored for multiprocessing and not changeable any more hit me again while I was working on the last PR. I think this is quite unintuitive and dangerous. I see two solutions:
- we patch the `cfg.PARAMS` and `cfg.PATHS` dicts so that  the multiprocessing pool is reset each time a parameter is changed
- we reset the multiprocessing pool each time we call `run_entity_task`

What do you think @TimoRoth ? Do you have another idea?

TimoRoth:
Fixes #327

### DIFF
--- a/oggm/cfg.py
+++ b/oggm/cfg.py
@@ -30,6 +30,9 @@ if not os.path.exists(CACHE_DIR):
 # Path to the config file
 CONFIG_FILE = os.path.join(os.path.expanduser('~'), '.oggm_config')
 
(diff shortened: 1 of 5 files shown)


fmaussion:
Thanks a lot! I think the best way to test this would be to remove all occurrences of reset_multiproc in the test suite and see if the tests pass

PR #346, opened 2017-11-07
compile climate files

anoukvlug:
Here is a function with which you can compile the climate files that are made by OGGM.

It might be possible to code the saving of the time lot more efficiently. Time is now stored in a comparable format as is done by compile_run_output, but it takes many more lines of code. So far I didn't find more elegant way to do that.

### DIFF
--- a/oggm/utils.py
+++ b/oggm/utils.py
@@ -1859,11 +1859,7 @@ def compile_climate_input(gdirs, path=True, filename='climate_monthly',
     if str('numpy.datetime64') in str(type(ds_clim.temp.time.values[0])):
         y0 = ds_clim.temp.time.values[0].astype('datetime64[Y]')
         y1 = ds_clim.temp.time.values[-1].astype('datetime64[Y]')
-    elif str('netcdftime._netcdftime.DatetimeGregorian') \
-            in str(type(ds_clim.temp.time.values[0])):
-        y0 = ds_clim.temp.time.values[0].strftime('%Y')
-        y1 = ds_clim.temp.time.values[-1].strftime('%Y')
-    elif str('datetime.datetime') in str(type(ds_clim.temp.time.values[0])):
+    else :
         y0 = ds_clim.temp.time.values[0].strftime('%Y')
         y1 = ds_clim.temp.time.values[-1].strftime('%Y')
 
--- a/oggm/utils.py
+++ b/oggm/utils.py
@@ -1863,27 +1863,18 @@ def compile_climate_input(gdirs, path=True, filename='climate_monthly',
             in str(type(ds_clim.temp.time.values[0])):
         y0 = ds_clim.temp.time.values[0].strftime('%Y')
         y1 = ds_clim.temp.time.values[-1].strftime('%Y')
+    elif str('datetime.datetime') in str(type(ds_clim.temp.time.values[0])):
+        y0 = ds_clim.temp.time.values[0].strftime('%Y')
+        y1 = ds_clim.temp.time.values[-1].strftime('%Y')
+
     time = pd.period_range('{}-10'.format(y0), '{}-9'.format(y1), freq='M')
     pkeep = np.ones(len(time), dtype=np.bool)
     year = time.year.values[pkeep]
     month = time.month.values[pkeep]
-    months = month.astype(np.float32)
-    months[months == 1] = 31/365
-    months[months == 2] = 59/365
-    months[months == 3] = 90/365
-    months[months == 4] = 120/365
-    months[months == 5] = 151/365
-    months[months == 6] = 181/365
-    months[months == 7] = 212/365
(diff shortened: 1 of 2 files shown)


oggm/utils.py
```diff
@@ -1821,6 +1822,119 @@ def compile_run_output(gdirs, path=True, monthly=False, filesuffix=''):
...
+    months[months == 3] = 90/365
+    months[months == 4] = 120/365
+    months[months == 5] = 151/365
+    months[months == 6] = 181/365
+    months[months == 7] = 212/365
+    months[months == 8] = 243/365
+    months[months == 9] = 273/365
+    months[months == 10] = 304/365
+    months[months == 11] = 334/365
+    months[months == 12] = 0
```
TimoRoth: This months == syntax looks weird to me. It probably all evaluates to 0/False. Did you mean to just do months[0] to months[11]?

fmaussion: Yes this looks a little clumsy. I'll have a look, but I might merge this first and correct it locally in a second PR

juliaeis: You could write :

m = np.cumsum([31,28,31,30,31,30,31,31,30,31,30,31])
m[-1] = 0

for i in range(12):
    months[months == (i+1)] = m[i]/365

This do exactly the same, but looks a little bit nicer.

fmaussion: Thanks for your suggestions! Note that there is already some logic in utils.py about this date conversion things: https://github.com/OGGM/oggm/blob/master/oggm/utils.py#L1085-L1113

But in this case I think we can even keep the logic simpler by using pd.date_range as in process_cesm_data: https://github.com/OGGM/oggm/blob/master/oggm/core/climate.py#L205-L212 .

@anoukvlug what do you think: do you want to give it a try? Also it would be great to have a test for this [here](https://github.com/OGGM/oggm/blob/master/oggm/tests/test_prepro.py#L1848)

anoukvlug: Thanks for all the suggestions! I will try to improve the code and add a test for this.

fmaussion: Note that it might be a bit more convoluted than I previously thought because in the end we use netCDF4 to write the data, not xarray: https://github.com/OGGM/oggm/blob/master/oggm/core/climate.py#L247

Anyways, let me know if you want to spend more time on this @anoukvlug , otherwise I can have a look

anoukvlug: I can spend some time on this this afternoon. 
It works to use: 
`time = date_to_year(year, month) `
to give the years in a floating format. 

@fmaussion I am not sure if I fully understand your second suggestion/ how it would makes it simpler?

fmaussion: Sorry, i dont think this was a good idea. I think that the code in process_cesm_data is closer to what you need

fmaussion: Sorry, i dont think this was a good idea. I think that the code in process_cesm_data is closer to what you need

anoukvlug: Do you mean to use than the format of "days since" and not use the same format as the files that are created with compile_run_output?

fmaussion: Right, it would be better to use the same format (I m on my phone, sorry). Than the date_to_yrs function is surely useful, and also the .years and .month attributes of pd.date_range

But just do what works for you and add a test, no need to spend hours on this ;)

fmaussion: (What I mean is that the test and the final result is more important than the internals)

anoukvlug:
I just made some changes to the compile_climate_input function and added a test.

oggm/utils.py
```diff
@@ -1821,6 +1822,110 @@ def compile_run_output(gdirs, path=True, monthly=False, filesuffix=''):
...
+    if str('numpy.datetime64') in str(type(ds_clim.temp.time.values[0])):
+        y0 = ds_clim.temp.time.values[0].astype('datetime64[Y]')
+        y1 = ds_clim.temp.time.values[-1].astype('datetime64[Y]')
+    elif str('netcdftime._netcdftime.DatetimeGregorian') \
+            in str(type(ds_clim.temp.time.values[0])):
+        y0 = ds_clim.temp.time.values[0].strftime('%Y')
+        y1 = ds_clim.temp.time.values[-1].strftime('%Y')
+    elif str('datetime.datetime') in str(type(ds_clim.temp.time.values[0])):
+        y0 = ds_clim.temp.time.values[0].strftime('%Y')
+        y1 = ds_clim.temp.time.values[-1].strftime('%Y')
```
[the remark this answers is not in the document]
fmaussion: OK, no problem. Good job! It looks better now, what do you think?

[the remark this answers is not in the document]
anoukvlug: Thanks! 
Yes, I think so too :)

fmaussion:
Thanks!

PR #347, opened 2017-11-08
Update run_alaska

bearecinos:
run_alaska.py with the new oggm version.

### DIFF
--- a/oggm/sandbox/run_alaska.py
+++ b/oggm/sandbox/run_alaska.py
@@ -38,70 +38,67 @@ rgi_reg = '01'
 # Initialize OGGM
 cfg.initialize()
 
-# What run we will do: without calving or with calving module
-With_calving = True
+# Compute a calving flux or not
+No_calving = False
+With_calving = False
 
+# Set's where is going to run PC or Cluster
 Cluster = False
-PC = True
+PC = False
 
-# Local paths (where to write output and where to download input)
+# Local paths (where to write output and where are the input files)
 if PC:
     MAIN_PATH = '/home/beatriz/Documents/'
-    WORKING_DIR = os.path.join(MAIN_PATH,'OGGM_Alaska_run_new/work_dir/')
+    WORKING_DIR = os.path.join(MAIN_PATH,'OGGM_Alaska_run/work_dir/')
     RGI_PATH = os.path.join(MAIN_PATH,
         'global_data_base/RGI_inventory/All_alaska_glacier_goodformat/')
     RGI_FILE = os.path.join(RGI_PATH,'01_rgi50_Alaska.shp')
     DATA_INPUT = os.path.join(MAIN_PATH, 'OGGM_Alaska_run/input_data/')
 
-    print(WORKING_DIR)
-    print(RGI_FILE)
-
 if Cluster:
     SLURM_WORKDIR = os.environ["WORKDIR"]
     # Local paths (where to write output and where to download input)
     WORKING_DIR = SLURM_WORKDIR
-    RGI_PATH = '~/oggm_run_alaska/input_data/Alaska_tidewater_andlake/'
+    RGI_PATH = '/home/users/bea/oggm_run_alaska/input_data/Alaska_tidewater_andlake/'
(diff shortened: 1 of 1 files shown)


oggm/sandbox/run_alaska.py
```diff
@@ -38,70 +38,67 @@
...
 # ---------------------------------
 # Note that to be exact, this procedure can only be applied if the run
 # parameters don't change between the calibration and the run.
 # After testing, it appears that changing the 'border' parameter won't affect
 # the results much (expectedly), so that it's ok to change it. All the rest
 # (e.g. smoothing, dx, prcp factor...) should imply a re-calibration
-mbf = 'https://www.dropbox.com/s/23a61yxwgpprs9q/ref_tstars_no_tidewater.csv?dl=1'
+mbf = 'https://github.com/OGGM/oggm-sample-data/blob/master/wgms/oggm_ref_tstars_rgi5_cru4.csv'
 mbf = utils.file_downloader(mbf)
 shutil.copyfile(mbf, os.path.join(WORKING_DIR, 'ref_tstars.csv'))
```
fmaussion: This is not necessary anymore: the `distribute_tstars` task does it for you: https://github.com/OGGM/oggm/blob/master/oggm/core/climate.py#L1116
(that  is, if `cfg.PARAMS['run_mb_calibration']` is `False`, which is the default)

oggm/sandbox/run_alaska.py
```diff
@@ -154,34 +175,37 @@
...
-    #Calculate glacier masks
     execute_entity_task(tasks.glacier_masks, gdirs)
 
-# The rest Prepro tasks
+# Pre-processing tasks
 task_list = [
     tasks.compute_centerlines,
     tasks.initialize_flowlines,
+    tasks.compute_downstream_line,
+    tasks.compute_downstream_bedshape,
```
fmaussion: Why do you need these? They are only necessary for doing a run

oggm/sandbox/run_alaska.py
```diff
@@ -220,18 +249,19 @@
     for gdir in gdirs:
         gdir.inversion_calving_rate = 0
 
+    cfg.PARAMS['correct_for_neg_flux'] = False
+    cfg.PARAMS['filter_for_neg_flux'] = False
```
fmaussion: Yes, but be careful to use the very last oggm version here (after @TimoRoth last PR - without it these parameters might be ignored)

oggm/sandbox/run_alaska.py
bearecinos: So then I can just delete that whole thing?

oggm/sandbox/run_alaska.py
bearecinos: mm I wasn't sure really what to take out or leave ... :P

oggm/sandbox/run_alaska.py
bearecinos: I dont understand this comment? 

And also about this setting I keep think I will have to do separate runs. One for land terminating and one for tidewater glaciers. I think that there should be a distinction inside the function between non-tidewater and tidewater. So, we don't have to worry about making separate runs or making extra ifs in run_alaska.py

oggm/sandbox/run_alaska.py
fmaussion: Yes

oggm/sandbox/run_alaska.py
fmaussion: Disregard my comment - if you updated oggm after timo's last PR, you are good

bearecinos:
Ok I alredy deleted the unnecessary lines ... is there anything else @fmaussion ? I think I am on the latest version so I should not worry about the cfg.PARAMS[] stuff

bearecinos:
@fmaussion is there was something else wrong with this file? Is it possible to merge this?

fmaussion:
Sorry I thought there still was something going on with the plots.

Just merged

bearecinos:
well this script has no plotting ... and also I guess the plotting problem is not related to oggm

PR #349, opened 2017-11-09
Add rectangle info to model section plot

fmaussion:


TimoRoth:
I think since the latest Ubuntu update it does, yes.

Am 09.11.2017 um 17:13 schrieb Fabien Maussion:

fmaussion:
Arg - I have no way to update the baselines myself then. Or maybe on the cluster?

TimoRoth:
The cluster is also using freetype 2.8, so it will use and generate for 
the alt directory there.

I also feel like alt is a horrible name, and we should rename them to 
indicate that they are for different freetype versions, which is now 
seamlessly possible with the pinned commit being used.

Am 09.11.2017 um 22:39 schrieb Fabien Maussion:

fmaussion:
> I also feel like alt is a horrible name, and we should rename them 

Agreed ;)

PR #350, opened 2017-11-09
Minor tweaks to compile climate output

fmaussion:
This just suppresses the warnings and makes sure we close the files for windows.

### DIFF
--- a/oggm/tests/test_prepro.py
+++ b/oggm/tests/test_prepro.py
@@ -1950,18 +1950,16 @@ class TestGCMClimate(unittest.TestCase):
     def test_compile_climate_input(self):
 
         filename = 'cesm_data'
-        filesuffix = '1'
+        filesuffix = '_cesm'
 
         hef_file = get_demo_file('Hintereisferner.shp')
         entity = gpd.GeoDataFrame.from_file(hef_file).iloc[0]
 
-        gdirs = []
-
         gdir = oggm.GlacierDirectory(entity, base_dir=self.testdir)
         gis.define_glacier_region(gdir, entity=entity)
-        gdirs.append(gdir)
+
         climate.process_cru_data(gdir)
-        utils.compile_climate_input(gdirs)
+        utils.compile_climate_input([gdir])
 
         f = get_demo_file('cesm.TREFHT.160001-200512.selection.nc')
         cfg.PATHS['gcm_temp_file'] = f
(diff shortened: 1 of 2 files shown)


fmaussion:
No that I had time to think about it, I am not so sure anymore that the time format is the correct one: in the compiled output files from the models, the time is in float years (as it is here), but it is in hydrological years! (i.e. float year 1901.00 means 01/10/1901).

I'm not sure yet what is best. For now we keep it like this, maybe you'll be able to make a better decision when you use the data @anoukvlug

fmaussion:
Yes, I think we shoud keep the same format for consistency with the other outputs, but convert it to hydro years too. Let's talk about it tomorrow

anoukvlug:
Thanks for making the changes.

I agree that the times should have the same format. I think floating hydrological years can be a bit confusing. What would be the reason to not save regular floating years?

fmaussion:
> I think floating hydrological years can be a bit confusing. What would be the reason to not save regular floating years?

Because hydrological years are the convention in glaciology (the reference MB data is stored this way). All years you have looked at are already hydrological years -- I think this is also how @bmarzeion is doing it. 

I *really* should make this clear in the files attribute though -- this is very error prone (actually now that I think of it there are maybe even a few bugs in the codebase due to this)

fmaussion:
Will do this in a separate PR

bmarzeion:
Yes, hydrological years are necessary for the validation etc., and also for hydrological applications. But I agree that they are a pain when it comes to sea level questions etc - which are global (tow different hydrological years on two hemispheres!) and where no one cares about hydrology. I think ultimately we should provide monthly outputs - and then make sure we look at the correct differences for the applications we are using OGGM for.

fmaussion:
> two different hydrological years on two hemispheres

Is it so? Are you using different definitions for the two hemispheres? Currently we do October-Sept everywhere

bmarzeion:
Yes, I'm using April to March in the Southern Hemisphere.

fmaussion:
:( definitely not the best news for a Sunday evening. Why did I not know this? Are the WGMS data defined this way?

bmarzeion:
I think so - but I was using Graham's collection for validation, so I'm not entirely sure for WGMS. But it would make sense, wouldn't it?

fmaussion:
I had a look, and information about this is quite buried in the database itself (`E-MASS-BALANCE-OVERVIEW.csv`), where *some* glaciers provide information about "their" definition of Winter and Annual MB. Indeed, most glaciers in the SH are from April to March :(

I'll have a look about how to implement this in OGGM. It's a small thing, but still a bit annoying ;)

Issue #348, opened 2017-11-09
The OGGM "time format" should be clearer


fmaussion:
Because of the infamous pandas [time limitation](https://github.com/pandas-dev/pandas/issues/7307) there are several time "formats" in OGGM:
- when working with CRU, we use the default pandas/xarray functionalities 
- with CESM data we have to do more tricks in between but still store the data in a CF compliant format (in "units since ..."). This implies that we assume that CESM actually uses a real calendar (i.e. with leap years and stuff) although I'm not sure if this is the case
- for the runs, we then get rid of these calendar formats because they are annoying and use a "float year" format. (https://github.com/OGGM/oggm/blob/master/oggm/utils.py#L1085-L1113)

At the very least I will rename this "year" format to the more explicit "float year" name. 

Better would be to document all these things somewhere

anoukvlug:
The CESM-LME uses a noleap calendar, but both calendars with leap and no leap years can be used by CESM: http://www.cesm.ucar.edu/models/cesm1.2/cesm/doc/usersguide/x1923.html

PR #352, opened 2017-11-10
Clarify time handling (+bugfix) 

fmaussion:
closes https://github.com/OGGM/oggm/issues/348


fmaussion:
This PR:
- solves a very small bug where the floating months in output file where not of the exact correct length because I used calendar months instead of the hydrological ones
- clarifies the model output time format by naming the coordinates the right way and by storing calendar as well as hydrological dates
- the compile_input_climate time format is now in line with the model output

Working in hydrological years might be confusing at first but this is what we want for the model output. With this format it is now possible to compute annual averages based on both calendars. The recommended way is simply: `ds.groupby('hydro_year').mean(dim='time')` for example

fmaussion:
cc @jlandmann

oggm/utils.py
```diff
@@ -1083,39 +1084,69 @@ def polygon_intersections(gdf):
...
+                       include_last_year=False):
+    """Creates a monthly timeseries in units of float years.
 
-def monthly_timeseries(y0, y1=None, ny=None, include_last_year=False):
-    """Creates a monthly timeseries in units of floating years.
+    Parameters
+    ----------
+    hydro_year : bool
+        If the float year follows the  "hydrological year" convention or
+        not (default:True)
```
anoukvlug: It might be nice to include in the description that y1 or yn need to be defined. On the other hand it is very clear when you read the function how it works, so it might not be really needed.

oggm/core/massbalance.py
```diff
@@ -238,6 +238,7 @@ def __init__(self, gdir, mu_star=None, bias=None, prcp_fac=None,
             ny, r = divmod(len(time), 12)
             if r != 0:
                 raise ValueError('Climate data should be N full years')
```
anoukvlug: Maybe it is a good moment to also include checking if the climate data actually start in January.

oggm/core/flowline.py
```diff
@@ -603,17 +620,47 @@ def run_until_and_store(self, y1, run_path=None, diag_path=None):
...
+        diag_ds['volume_m3'].attrs['unit'] = 'm 3'
         diag_ds['area_m2'] = ('time', np.zeros(nm) * np.NaN)
+        diag_ds['area_m2'].attrs['description'] = 'Total glacier area'
+        diag_ds['area_m2'].attrs['unit'] = 'm 2'
         diag_ds['length_m'] = ('time', np.zeros(nm) * np.NaN)
+        diag_ds['length_m'].attrs['description'] = 'Glacier length'
+        diag_ds['length_m'].attrs['unit'] = 'm 3'
         diag_ds['ela_m'] = ('time', np.zeros(nm) * np.NaN)
+        diag_ds['ela_m'].attrs['description'] = ('Annual Equilibrium Line '
+                                                 'Altitude')
```
anoukvlug: I think that the Equilibrium Line Altitude is by definition annual, therefore I believe that the word 'annual' is redundant in this case.

oggm/utils.py
fmaussion: Yes you are right the docstring is not complete here

oggm/core/massbalance.py
fmaussion: Actually, we assume the that the data starts in October (because of the hydrological year)

oggm/core/flowline.py
fmaussion: Correct. I'll keep it that way though: because of implementation details, it is still stored at a monthly time step, so to make it clear I added the "Annual"

fmaussion:
Thanks for the review! Next time I'll try to wait a bit more before merging ;)

oggm/core/flowline.py
anoukvlug: I didn't think of that, but in that case it makes indeed a lot off sense to leave it like it is :)

oggm/core/massbalance.py
anoukvlug: Somehow I mixed things up, but you are right.

anoukvlug:
Thanks @fmaussion  for making the changes, it made things more clear to me :)

Issue #359, opened 2017-11-21
Time change in PastMassBalance model


sumnonpuella:
I'm getting odd errors trying to use the PastMassBalance model with a custom climate data. The input climate file has 12 months of data (October 1950 - September 1951). When I try to run until equilibrium, the model is searching for years that aren't there. With y0=1950, 

`mb = mbmods.PastMassBalance(gdir, filename='cesm_data')`
`model_fls = gdir.read_pickle('model_flowlines')`
`model = FluxBasedModel(model_fls, mb_model=mb, y0=1950, time_stepping='default')`
`model.run_until_equilibrium()`
`ValueError: Year 1950.0 not in record`

and with y0 = 1951,

`model = FluxBasedModel(model_fls, mb_model=mb, y0=1951, time_stepping='default')`
`model.run_until_equilibrium()`
`ValueError: Year 1952.013698630137 not in record`.

I looked at the times stored in `mb`, and the time stored in the model doesn't match up with the input file. 

`self.years = [1951, 1951, 1951, 1951, 1951, 1951, 1951, 1951, 1951, 1951, 1951,
       1951]`
`self.months = [ 1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11, 12]`

It looks like this time is switched in PastMassBalance (line 243).

`# This is where we switch to hydro float year format`
`# Last year gives the tone of the hydro year`
`self.years = np.repeat(np.arange(time[-1].year-ny+1, time[-1].year+1), 12)`
`self.months = np.tile(np.arange(1, 13), ny)`

Is this a bug in OGGM, or do I need to set up the mass balance model differently? I can try to adjust PastMassBalance to store the correct dates if need be.

fmaussion:
Thanks for the report!

I think I understand a bit better what's going on. 

[...]

PR #362, opened 2017-11-28
Update ConstantMassBalance

sumnonpuella:
Added kwargs to `Constant Mass Balance` so that it's usable with CESM input data. 

closes https://github.com/OGGM/oggm/issues/359


fmaussion:
Great! Thanks a lot.

oggm/core/massbalance.py
```diff
@@ -354,7 +354,7 @@ def __init__(self, gdir, mu_star=None, bias=None, prcp_fac=None,
 
         super(ConstantMassBalance, self).__init__()
         self.mbmod = PastMassBalance(gdir, mu_star=mu_star, bias=bias,
-                                     prcp_fac=prcp_fac)
+                                     prcp_fac=prcp_fac, filename='climate_monthly', input_filesuffix='')
```
fmaussion: I think the correct way here should be:

```filename=filename, input_filesuffix=input_filesuffix```

Otherwise the kwargs are not passed properly

fmaussion:
thanks @sumnonpuella !

PR #364, opened 2017-11-30
Simplify float year

fmaussion:


fmaussion:
This simplifies the float years by making it independent on the hydrological years. It is an even simpler calendar where each month has the same length.

I think that this makes the monthly MB a bit wrong but I guess we can cope with this.

fmaussion:
See also: https://github.com/OGGM/oggm/issues/365

PR #377, opened 2017-12-11
Add South Glacier benchmark tests

fmaussion:


fmaussion:
This PR adds a new benchmark test file where we can try the OGGM workflow on real cases. Here we use the South Glacier case from ITMIX, where good DEMs, MB data  and GPR measurements are available. 

We don't calibrate OGGM on the GPR data yet, but this is surely on the TODO list for later.

cc @alexjarosch FYI

PR #385, opened 2018-01-09
Add new terminus correction task for tidewater glaciers

fmaussion:
Together with @bearecinos we thought it would be useful to change the widths whilst conserving the rgi area


fmaussion:
Also, I found a bug in a utility function and corrected it

PR #387, opened 2018-01-10
Also pass file related kwargs in RandomMassBalance model

anoukvlug:
Follow up PR to https://github.com/OGGM/oggm/pull/362, where we added it the the ConstantMassBalance model


fmaussion:
Thanks @anoukvlug !

PR #388, opened 2018-01-10
Add file related kwargs to dynamical run tasks

anoukvlug:
Another simple PR allowing to use different climate files for the random and constant climate runs


fmaussion:
Thanks!

PR #393, opened 2018-01-12
Use fractional anomalies for precipitation timeseries

fmaussion:


fmaussion:
@anoukvlug you should consider using this code for your purposes from now on.

This change won't make it for version 1.0 of OGGM, but the CESM part could use this new default. Let's talk about the strategy to use offline in slack

anoukvlug:
Thanks for making these changes @fmaussion. Discussing it further on Slack sounds like a good plan.

fmaussion:
OK this should be ready. @anoukvlug can you have a look? Thanks!

docs/mass-balance.rst
```diff
@@ -70,6 +73,21 @@ folder for an example.
     example_plot_temp_ts()  # the code for these examples is posted below
 
 
+GCM data
+~~~~~~~~
+
+OGGM can also use climate model output to drive the mass-balance model. In
+these case we still rely on gridded observations (CRU) for the baseline
```
anoukvlug: "this case" instead of "these case"

fmaussion:
In it goes!

PR #395, opened 2018-01-13
Clean the sandbox

fmaussion:
In view of the upcoming release, this cleans out unnecessary code in the sandbox.

The sandbox was dangerous because it was not curated and quick'n dirty scripts, which should live elsewhere. The ITMIX and GMD paper scripts have now their dedicated repositories (see below), and so has @bearecinos 's run_alaska scripts.

Instead of the sandbox I should make more efforts to put example scripts in the documentation -- i.e. a kind of "cookbook"

Links: 

https://github.com/OGGM/itmix_phase1

https://github.com/OGGM/gmd_paper_2018


fmaussion:
3,609 lines of code "rmoved"!

bearecinos:
Noted. 

The run_alaska.py script can be found in my repository under calving_experiments.

Thanks for hosting my code all this time :-) (nostalgic)

Issue #406, opened 2018-01-23
Update the new reference t* table after recent CRU update


fmaussion:
This is critical for users using the latest version, probably @anoukvlug

anoukvlug:
Thanks for the warning @fmaussion.

PR #407, opened 2018-01-24
Update ref_tstars after PRCP update

fmaussion:
- [x] Closes https://github.com/OGGM/oggm/issues/406
 - [ ] Tests added/passed

Relevant commit in sample data: https://github.com/OGGM/oggm-sample-data/commit/17fe98844aed27199d2dfc7fd8af2bc3e356dbd0

It appears that the changes in t* are not so large, mostly the biases are affected. The new bias in standard deviation is 87% (instead of 90% before), indicating that we should probably go for a larger precip correction factor


fmaussion:
The remaining failing test is due to googlemap - will do in a separate PR

Issue #394, opened 2018-01-12
Add possibility to start a run with spinup


fmaussion:
It's not really hard to do, but the complicated part is the API: how to make it flexible enough?

PR #411, opened 2018-01-31
Add run from spinup option

fmaussion:
- [x] Closes https://github.com/OGGM/oggm/issues/394
 - [x] Tests added/passed


fmaussion:
Still a WIP: this first commit simply refactors some functions

fmaussion:
OK, this should be it. CC @anoukvlug for a review. 

Note that once again this is not backwards compatible (keyword names have changed). Altogether this is for a good purpose, the calls are now much simpler. Furthermore, no new task was needed, which is also nice.

fmaussion:
> I am just wondering, why do you keep the "filename" keyword and not also change it to "climate_filename"? 

That's a very good point! The actual reason is that it would mean even more changes -- and I didn't want to do it all at once without a clear strategy. Will open an issue for this

Issue #418, opened 2018-03-04
matplotlib._cntr not found


MartinLuethi:
Hi

matplotlib does *not* contain a module called _cntr. 
Please stick to standard modules that are easily installed.

This is matplotlib git version (pre 2.2)
version '2.2.0rc1+282.g3f3065d3a.dirty'

fmaussion:
OK, after doing some research: we should indeed find another way to compute the contours of our masks. I will dig into this, but the easiest for now would be to stick to mpl version 2.1

(mpl 2.2 is going to be officially released today or tomorrow, sigh...)

MartinLuethi:
Hi Fabien

Fabien Maussion <notifications@github.com> writes:

> Thanks for the report! The probably changed in the lastest Matpotlib
> version (2.2). Your installed package is still a release candidate and
> therefore has not been tested yet. I will try to implement a fix for
> this later today.

Yes, with stock matplotlib it works. It seems that ._cntr was not an
official part of the library (the underscore naming warns about that).

I'm considering using OGGM in our glacier modeling class, so I'll test
it somewhat more in the coming days.

Best, Martin

-- 
Dr. Martin Lüthi
Department of Geography, 3G
University of Zurich
Winterthurerstrasse 190
CH-8057 Zurich, Switzerland
Phone: +41-44-635-5146
Fax:   +41-44-635-6841
Email: martin.luethi@geo.uzh.ch

fmaussion:
Thanks a lot for the heads up! It's always unfortunate to start with an error. 

Thanks for considering OGGM for your class! If you are interested in idealized examples, a good way to start could be to have a look at the [flowline](https://github.com/OGGM/oggm/blob/master/docs/notebooks/flowline_model.ipynb) example. This [blog post](http://oggm.org/2017/10/01/specmb-ela/) is another example.

I think that one of the hardest bit is (as you've noticed) to get all the necessary python packages installed for all your students. I discussed a possible solution with Horst last time he came to visit, but if you are interested you can contact me offline or on the phone to discuss the deployment solution we have opted for our classes in Innsbruck.

MartinLuethi:
Fabien Maussion <notifications@github.com> writes:

> OK, after doing some research: we should indeed find another way to compute the contours of our masks. I will dig into this, but the easiest for now would be to stick to mpl version 2.1

Or you add the code from ._cntr to your repository under "src/contrib"
including the original license file. Many codes do that.

Best, Martin

[...]

PR #421, opened 2018-03-05
 Remove depency to mpl._cntr 

fmaussion:
- [x] Closes https://github.com/OGGM/oggm/issues/418
 - [x] Tests added/passed


fmaussion:
This now uses [skimage](http://scikit-image.org/docs/dev/auto_examples/edges/plot_contours.html) to do the job.

To my surprise, the new implementation does not seem to change *anything* in the output, which is great! It's also much nicer not to rely on private modules to do the job.

PR #429, opened 2018-03-08
Lateral drag / Shape factors

### LINKED ISSUE #423: Implement Cuffey and Patterson's shape factors to the flowline model
This should have high priority. The model is very sensitive to the shape of the glacier bed, and I hope that the shape factors will reduce this sensitivity. There are two places where this should happen:
- in the [inversion formulation](https://github.com/OGGM/oggm/blob/master/oggm/core/inversion.py#L129) (it's probably just about adding a multiplication factor somewhere, but we should do the maths)
- in the flowline model ([FluxBasedModel](https://github.com/OGGM/oggm/blob/master/oggm/core/flowline.py#L738) should be enough for a start). This might be a bit more intricate because of the API but should also not be too much work.

Altogether this is not a very big deal, but as always this is going to change the results and call for more testing.

phigre:
- [x] Closes #423 
 - [x] Passes Inversion Tests for runs without shape factors
 - [ ] Tests for inversion with shape factors added/passed
 - [x] Tests for shape-factor interpolation from Adhikari /Nye added and passed
 - [x] Fully documented, including `whats-new.rst` for all changes (remove if this change is small enough to be undocumented)
        not yet to a sufficient amount

For first review (@fmaussion)

### DIFF
--- a/oggm/cfg.py
+++ b/oggm/cfg.py
@@ -280,6 +280,8 @@ def initialize(file=None):
     # Flowline model
     _k = 'use_optimized_inversion_params'
     PARAMS[_k] = cp.as_bool(_k)
+    PARAMS['use_shape_factor_for_fluxbasedmodel'] = \
+                cp['use_shape_factor_for_fluxbasedmodel']
 
     # Make sure we have a proper cache dir
     from oggm.utils import download_oggm_files, SAMPLE_DATA_COMMIT
@@ -298,7 +300,8 @@ def initialize(file=None):
            'prcp_scaling_factor', 'use_intersects', 'filter_min_slope',
            'auto_skip_task', 'correct_for_neg_flux', 'filter_for_neg_flux',
            'rgi_version', 'allow_negative_mustar',
-           'use_shape_factor_for_inversion']
+           'use_shape_factor_for_inversion',
+           'use_shape_factor_for_fluxbasedmodel']
     for k in ltr:
         cp.pop(k, None)
 
--- a/oggm/cfg.py
+++ b/oggm/cfg.py
@@ -274,7 +274,8 @@ def initialize(file=None):
     PARAMS['invert_with_sliding'] = cp.as_bool('invert_with_sliding')
     _k = 'optimize_inversion_params'
     PARAMS[_k] = cp.as_bool(_k)
-    PARAMS['use_shape_factor'] = cp.as_bool('use_shape_factor')
+    PARAMS['use_shape_factor_for_inversion'] = \
+                cp['use_shape_factor_for_inversion']
 
     # Flowline model
     _k = 'use_optimized_inversion_params'
@@ -296,7 +297,8 @@ def initialize(file=None):
            'tstar_search_window', 'use_bias_for_run', 'hydro_month_sh',
            'prcp_scaling_factor', 'use_intersects', 'filter_min_slope',
            'auto_skip_task', 'correct_for_neg_flux', 'filter_for_neg_flux',
-           'rgi_version', 'allow_negative_mustar', 'use_shape_factor']
+           'rgi_version', 'allow_negative_mustar',
+           'use_shape_factor_for_inversion']
     for k in ltr:
         cp.pop(k, None)
 
--- a/oggm/cfg.py
+++ b/oggm/cfg.py
@@ -274,6 +274,7 @@ def initialize(file=None):
     PARAMS['invert_with_sliding'] = cp.as_bool('invert_with_sliding')
     _k = 'optimize_inversion_params'
     PARAMS[_k] = cp.as_bool(_k)
+    PARAMS['use_shape_factor'] = cp.as_bool('use_shape_factor')
 
     # Flowline model
     _k = 'use_optimized_inversion_params'
@@ -295,7 +296,7 @@ def initialize(file=None):
            'tstar_search_window', 'use_bias_for_run', 'hydro_month_sh',
            'prcp_scaling_factor', 'use_intersects', 'filter_min_slope',
            'auto_skip_task', 'correct_for_neg_flux', 'filter_for_neg_flux',
-           'rgi_version', 'allow_negative_mustar']
+           'rgi_version', 'allow_negative_mustar', 'use_shape_factor']
     for k in ltr:
         cp.pop(k, None)
 
--- a/oggm/core/flowline.py
+++ b/oggm/core/flowline.py
@@ -786,6 +786,17 @@ class FluxBasedModel(FlowlineModel):
         self.cfl_number = cfl_number
         self.calving_m3_since_y0 = 0.  # total calving since time y0
 
+        # Do we want to use shape factors?
+        self.sf_func = None
+        # Use .get to obtain default None for non-existing key
+        # necessary to pass some tests
+        # TODO: change to direct dictionary query after tests are adapted?
+        use_sf = cfg.PARAMS.get('use_shape_factor_for_fluxbasedmodel')
+        if use_sf == 'Adhikari' or use_sf == 'Nye':
+            self.sf_func = utils.shape_factor_adhikari
+        elif use_sf == 'Huss':
+            self.sf_func = utils.shape_factor_huss
+
         # Optim
         self._stags = []
         for fl, trib in zip(self.fls, self._trib):
@@ -797,9 +808,10 @@ class FluxBasedModel(FlowlineModel):
             a = np.zeros(nx+1)
             b = np.zeros(nx+1)
             c = np.zeros(nx+1)
-            d = np.zeros(nx-1)
-            e = np.zeros(nx)
-            self._stags.append((a, b, c, d, e))
+            d = np.ones(nx+1)  # shape factor default is 1
+            e = np.zeros(nx-1)
+            f = np.zeros(nx)
+            self._stags.append((a, b, c, d, e, f))
 
     def step(self, dt):
         """Advance one step."""
@@ -811,20 +823,8 @@ class FluxBasedModel(FlowlineModel):
         flxs = []
         aflxs = []
 
-        # Do we want to use shape factors?
-        sf_func = None
-        use_sf = None
-        # Use .get to obatin default None for non-existing key
-        # necessary to pass some tests
-        # TODO: remove after tests are adapted
-        use_sf = cfg.PARAMS.get('use_shape_factor_for_fluxbasedmodel')
-        if use_sf == 'Adhikari' or use_sf == 'Nye':
-            sf_func = utils.shape_factor_adhikari
-        elif use_sf == 'Huss':
-            sf_func = utils.shape_factor_huss
-
-        for fl, trib, (slope_stag, thick_stag, section_stag, znxm1, znx) \
-                in zip(self.fls, self._trib, self._stags):
+        for fl, trib, (slope_stag, thick_stag, section_stag, sf_stag,
+                       znxm1, znx) in zip(self.fls, self._trib, self._stags):
 
             surface_h = fl.surface_h
             thick = fl.thick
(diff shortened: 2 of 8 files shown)


oggm/cfg.py
```diff
@@ -274,6 +274,7 @@ def initialize(file=None):
     PARAMS['invert_with_sliding'] = cp.as_bool('invert_with_sliding')
     _k = 'optimize_inversion_params'
     PARAMS[_k] = cp.as_bool(_k)
+    PARAMS['use_shape_factor'] = cp.as_bool('use_shape_factor')
```
fmaussion: I think we should rename it to ``use_shape_factor_for_inversion``. It is likely that we will want to separate between inversion and dynamical model's factors.

For testing between the factors it might be useful to have this option as a str: ``'Huss'`` (or the original paper name), ``"Adhikari"`` or empty for no shape factor

oggm/core/inversion.py
```diff
@@ -126,6 +131,113 @@ def _inversion_simple(a3, a0):
     return (-a0)**cfg.ONE_FIFTH
 
 
+def shape_factor_huss(widths, heights, is_rectangular):
+    """Compute shape factor for inclusion of lateral drag
+    according to Huss and Farinotti (2012)
+
+    Discouraged to use, as this does not make any difference
+    between parabolic and rectangular beds
```
fmaussion: I'm not sure to understand: wouldn't it be better to return a factor of 1 when the bed is rectangular in this case?

oggm/core/inversion.py
```diff
@@ -126,6 +131,113 @@ def _inversion_simple(a3, a0):
...
+    """
+    return widths / (2 * heights + widths)
+
+
+def shape_factor_adhikari(widths, heights, is_rectangular):
+    """Compute shape factor for inclusion of lateral drag according to
+    Adhikari (2012)
+
+    TODO: should we expand this here to also include
+    the factors suggested for sliding?
```
fmaussion: Not yet I would say

oggm/core/inversion.py
```diff
@@ -126,6 +131,113 @@ def _inversion_simple(a3, a0):
...
+    # Table 1 from Adhikari (2012)
+    tabular_zetas = np.array([0.5, 1, 2, 3, 4, 5, 10])
+    factors_rectangular = np.array([0.313, 0.558, 0.790, 0.884,
+                                    0.929, 0.954, 0.990])
+    f_rect = interp1d(tabular_zetas, factors_rectangular,
+                      fill_value='extrapolate')
+    factors_parabolic = np.array([0.251, 0.448, 0.653, 0.748,
+                                  0.803, 0.839, 0.917])
+    f_parab = interp1d(tabular_zetas, factors_parabolic,
+                       fill_value='extrapolate')
```
fmaussion: This entire block could live at the module scope instead

oggm/core/inversion.py
```diff
@@ -126,6 +131,113 @@ def _inversion_simple(a3, a0):
...
+
+    # Ensure bool (for masking)
+    is_rectangular = is_rectangular.astype(bool)
+
+    # TODO: could check for division by 0, but at the moment
+    # this is covered by interpolation and clip, resulting in a factor of 1
```
fmaussion: It's just a problem if it throws numerical warnings. If you want you can put this statement in a catch as done [here](https://github.com/OGGM/oggm/blob/master/oggm/core/flowline.py#L1099)

oggm/core/inversion.py
```diff
@@ -126,6 +131,113 @@ def _inversion_simple(a3, a0):
...
+    a3
+    flux_a0
+    shape_factor
+    _inv_function
+
+    Returns
+    -------
+
+    """
+    a0s_shaped = a0s / (shape_factor ** 3)
```
fmaussion: you can safely overwrite ``a0s`` here

oggm/core/inversion.py
```diff
@@ -161,6 +273,11 @@ def mass_conservation_inversion(gdir, glen_a=cfg.A, fs=0., write=True,
     fd = 2. / (cfg.N+2) * glen_a
     a3 = fs / fd
 
+    # Shape factor params
+    sf_func = shape_factor_adhikari
+    sf_tol = 1e-2  # TODO: better as params in cfg?
```
fmaussion: This is too specific it's OK to leave it here

oggm/core/inversion.py
```diff
@@ -174,29 +291,43 @@ def mass_conservation_inversion(gdir, glen_a=cfg.A, fs=0., write=True,
...
-        for i, (a0, Q) in enumerate(zip(a0s, cl['flux_a0'])):
-            if Q > 0.:
-                out_thick[i] = _inv_function(a3, a0)
-            else:
-                out_thick[i] = 0.
-        assert np.all(np.isfinite(out_thick))
+        # Start iteration for shape factor with guess of 1
+        i = 0
+        sf = np.ones(slope.shape)
+        sf_diff = np.ones(slope.shape)
```
fmaussion: ``i = 0`` and ``sf_diff`` could be defined in the ``if`` block below

oggm/core/inversion.py
```diff
@@ -126,6 +131,113 @@ def _inversion_simple(a3, a0):
...
+    # Ensure bool (for masking)
+    is_rectangular = is_rectangular.astype(bool)
+
+    # TODO: could check for division by 0, but at the moment
+    # this is covered by interpolation and clip, resulting in a factor of 1
+    zetas = widths / heights
+
+    shape_factors = np.ones(widths.shape)
+
+    # TODO: higher order interpolation? (e.g. via InterpolatedUnivariateSpline)
```
fmaussion: Maybe? Depends on how the curve looks like

fmaussion:
Ping @alexjarosch for info

oggm/core/inversion.py
phigre: But then I cannot exploit the precomputed a0s in subsequent loops

oggm/core/inversion.py
phigre: I could not really find any information on the assumed bed form in the paper. Therefore I did not go into too much detail yet. But you're right, I will adapt it this way.

oggm/core/inversion.py
phigre: In case is_rectangular is of type integer, these integers (0/1) are directly interpreted as indices, which is not what is intended here. In case of floats I have to admit that I did not check.

oggm/core/inversion.py
phigre: Absolutely right

oggm/core/inversion.py
phigre: Already looks quite ok, i would say?

grafik

I can provide the notebook to this figure on demand

oggm/core/inversion.py
phigre: Agreed

oggm/cfg.py
phigre: Yes, I already also thought about an option like this, but just did not spend more time on it. Will do so.

oggm/core/inversion.py
phigre: Sorry, my mistake. Did not realize I was inside the function.

oggm/core/inversion.py
```diff
@@ -151,7 +151,29 @@ def shape_factor_huss(widths, heights, is_rectangular):
...
+    shape_factors = np.ones(widths.shape)
+    shape_factors[~is_rectangular] = \
+        (widths / (2 * heights + widths))[~is_rectangular]
+
+    return shape_factors
+
+
+# TODO: how to handle zeta > 10? at the moment extrapolation
+# Table 1 from Adhikari (2012) and corresponding interpolation functions
+_ADHIKARI_TABLE_ZETAS = np.array([0.5, 1, 2, 3, 4, 5, 10])
```
phigre: Move this to top of file?

fmaussion: I'm not sure there is a pep rule for this, but yes I think it's slightly better

oggm/params.cfg
```diff
@@ -189,6 +189,9 @@ optimize_inversion_params = False
 # If false, tell OGGM which should be used
 inversion_glen_a = 2.4e-24
 inversion_fs = 0.
+# Do you want to use shape factors to account for lateral drag?
+# Allowed is empty, "Adhikari", "Nye" equal to "Adhikari" or "Huss"
```
fmaussion: `` "Nye" (equivalent to "Adhikari")``

oggm/core/inversion.py
```diff
@@ -161,6 +287,16 @@ def mass_conservation_inversion(gdir, glen_a=cfg.A, fs=0., write=True,
     fd = 2. / (cfg.N+2) * glen_a
     a3 = fs / fd
 
+    # Shape factor params
+    sf_func = None
+    if cfg.PARAMS['use_shape_factor_for_inversion'] == 'Adhikari' or \
```
fmaussion: create a new variable, like ``use_shape_factor = cfg.PARAMS['use_shape_factor_for_inversion']`` and use it throughout the function to shorten the code a little bit

fmaussion:
@phigre should we merge this as is and add the tests at a later stage or are you still working on this?

fmaussion:
Ah, and don't forget to acknowledge your work in ``whats-new.rst`` !

oggm/core/flowline.py
```diff
@@ -851,9 +852,25 @@ def step(self, dt):
...
+                    'Adhikari' or \
+                    cfg.PARAMS['use_shape_factor_for_fluxbasedmodel'] == 'Nye':
+                sf_func = shape_factor_adhikari
+            elif cfg.PARAMS['use_shape_factor_for_fluxbasedmodel'] == 'Huss':
+                sf_func = shape_factor_huss
+
+            sf = sf_func(fl.widths_m, thick, fl.is_rectangular)
+            # TODO: avoid  unnecessary new array initialization by adding
+            # sf_stag to  self._stags
+            sf_stag = np.zeros(thick_stag.shape)
```
phigre: Staggered shape factors need to be computed to match with `thick_stag` and `u_stag`.
I decided to use shape factors from the unstaggered grid and combine upstream and downstream shape factors to obtain staggered values.
Directly computing shape factors on the staggered grid would make an additional `width_stag` necessary and the question arises what `is_rectangular_stag` would look like.

oggm/core/flowline.py
```diff
@@ -21,6 +21,7 @@
 from oggm import entity_task
 import oggm.core.massbalance as mbmods
 from oggm.core.centerlines import Centerline, line_order
+from oggm.core.inversion import shape_factor_adhikari, shape_factor_huss
```
phigre: I referenced the shape factor functions from the inversion module. IMHO, having them in a separate dedicated module and referencing this module from inversion and flowline would make more sense in regards of code structure.

fmaussion: Yes, probably. It could be added to ``utils`` for now (and the test moved to ``test_utils`` for now, even if the utils module is definitely overcrowded and could need a split-up one day. What do you think?

phigre: e.g. move to oggm.utils?

oggm/core/flowline.py
```diff
@@ -851,9 +852,25 @@ def step(self, dt):
...
+            sf_func = None
+            sf = np.ones(thick.shape)  # Default shape factor is 1
+            if cfg.PARAMS['use_shape_factor_for_fluxbasedmodel'] == \
+                    'Adhikari' or \
+                    cfg.PARAMS['use_shape_factor_for_fluxbasedmodel'] == 'Nye':
+                sf_func = shape_factor_adhikari
+            elif cfg.PARAMS['use_shape_factor_for_fluxbasedmodel'] == 'Huss':
+                sf_func = shape_factor_huss
+
+            sf = sf_func(fl.widths_m, thick, fl.is_rectangular)
```
phigre: Tests are broken, probably due to this line and probably therein by fl.is_rectangular, as this is not always defined. Any good ideas on how to fix? By default assume rectangular grid?

phigre:
There are still some issues like the storage of `sf_stag` in `self._stag` and how to handle not set `fl.is_rectangular` (yes, that is possible), which I didn't want to decide on my own. 

Otherwise ready to merge now?
A simple test for creating a synthetic glacier with shape factors in the FluxBasedModel and trying to recover the bed with mass-conservation inversion and the adequate shape factors for inversion worked out, so at least we ca assume that either both are right or broken ;)

oggm/tests/test_models.py
```diff
@@ -1377,6 +1377,96 @@ def test_inversion_parabolic(self):
             plt.plot(bed_shape_gl[:-3])
             plt.show()
 
+    def test_inversion_parabolic_sf_adhikari(self):
```
phigre: Tests for each case would probably be much better with one single callable function and write wrappers, which are actually called for testing and set the shape factor mode before calling the function.

phigre: Much better to define one single function and then write wrappers for different shape factor modes, which are actually called for testing and do nothing else than setting shape factor cfg and call the function doing the work. Would avoid a lot of ugly copied code

oggm/core/flowline.py
```diff
@@ -797,9 +808,10 @@ def __init__(self, flowlines, mb_model=None, y0=0., glen_a=None,
             a = np.zeros(nx+1)
             b = np.zeros(nx+1)
             c = np.zeros(nx+1)
-            d = np.zeros(nx-1)
-            e = np.zeros(nx)
-            self._stags.append((a, b, c, d, e))
+            d = np.ones(nx+1)  # shape factor default is 1
```
phigre: Set to `1.0` as a scalar if scalar*array is faster than array*array in numpy. Would be cool, if numpy could skip multiplication by 1 automatically

oggm/utils.py
```diff
@@ -3161,10 +3161,10 @@ def shape_factor_huss(widths, heights, is_rectangular):
 
     # Ensure bool (for masking)
     is_rectangular = is_rectangular.astype(bool)
```
phigre: Avoid typecast necessary? Did not look into the implementation and cannot say anything about perfomance impacts

oggm/utils.py
```diff
@@ -3161,10 +3161,10 @@ def shape_factor_huss(widths, heights, is_rectangular):
...
-
     shape_factors = np.ones(widths.shape)
```
phigre: always repeating initialization could be substituted by storing shape factor array in the flowmodel and passing the reference to the shape factor functions

oggm/utils.py
```diff
@@ -3161,10 +3161,10 @@ def shape_factor_huss(widths, heights, is_rectangular):
...
     shape_factors[~is_rectangular] = \
         (widths / (2 * heights + widths))[~is_rectangular]
+    shape_factors[heights <= 0.] = 1.
```
phigre: Need to avoid nans for areas without ice

fmaussion:
This is a brilliant piece of work, thanks a lot! I will merge this now, we can always go back to some details later. 

Experience shows that the global applications will require some tuning, but I'm eager to test this out. Note that this probably should become the default some day.

PR #430, opened 2018-03-09
Print more useful error messages in case of exceptions in tasks

TimoRoth:
Saves you from opening a debugger just to look at the Exception.


fmaussion:
> Saves you from opening a debugger just to look at the Exception.

Just a comment on this: the exception is printed in full (as far as I can remember) in the log files of the glacier working directory. 

Still, is a useful change I think

PR #453, opened 2018-04-20
Enhance/Fix tests to work distributed with pytest-xdist

TimoRoth:
With this, the cluster can run the tests in parallel on 56 Threads, completing all tests, including slow tests, in around one minute.

### DIFF
--- a/oggm/tests/__init__.py
+++ b/oggm/tests/__init__.py
@@ -47,6 +47,7 @@ RUN_WORKFLOW_TESTS = True
 RUN_GRAPHIC_TESTS = True
 RUN_BENCHMARK_TESTS = True
 RUN_PERFORMANCE_TESTS = False
+RUN_CREDENTIAL_TESTS = False
 if os.environ.get('TRAVIS') is not None:
     # specific to travis to reduce global test time
     ON_TRAVIS = True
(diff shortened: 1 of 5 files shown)


fmaussion:
Thanks a lot! I'm sorry about the download tests, there are badly defined. I actually almost never run them, and I am surprised they are still passing on the cluster. Good to know!

fmaussion:
I just noticed the new hook: THIS IS AWESOME!

fmaussion:
> This won't run for pull requests for security reasons though, only commits to all branches of OGGM/oggm

I like this. It's also good for new commiters, which might otherwise get red flags from cluster problems they have no clue about.

We should monitor what's happening on the cluster though. Maybe via email when run fails? Another possibility (but also prone to be overseen) are badges. Would it be possible to get a badge for this status? https://shields.io/ provides some dynamic hooks but I'm not sure it's worth the trouble.

TimoRoth:
I added a simple banner at https://cluster.klima.uni-bremen.de/~github/status.svg
Banner

fmaussion:
Cool! Is there a link to the latest test results? 

Like this PR had: https://cluster.klima.uni-bremen.de/~github/status/1926.html as url -> can we replace ``1926`` with latest or something?

TimoRoth:
https://cluster.klima.uni-bremen.de/~github/status/latest.html should 
work just fine.

PR #458, opened 2018-05-08
Relax constraint on RGI area conservation with new parameter

matthiasdusch:
- [x] Tests added/passed

with ``PARAMS['use_rgi_area'] = False``, OGGM will use the geometry's area instead of the one provided by the RGI shapefile.


fmaussion:
Thanks!

Issue #467, opened 2018-05-31
copy_to_basedir 


anoukvlug:
@fmaussion, like we discussed during the oggm workshop, I would like to make a pull request for adding ''cesm_data'' to the options 'run' and 'inversion' of the copy_to_basedir function.

While playing around with this function, I didn't find an entity task that can be used to use this function. I just wrote one. Shall I add it to the pull request?  Or is there already a more suitable way to use this function?

fmaussion:
Hi Anouk, until now the ``copy_to_basedir`` is a method available from the gdir object. I never considered making an entity task out of it, although maybe it would make sense for paralelizing the copy/paste. 

Yes, open a PR and we can discuss the details there

fmaussion:
Just to claridfy: until now I did:

```
for gdir in gdirs:
    gdir.copy_to_basedir()
```

But maybe this could be done in parallel yes

anoukvlug:
It make a sense to do it like that. Somehow the first thing that came to my mind was making an entity task for it. Anyway I will make a pull request to discuss this further.

PR #468, opened 2018-05-31
some additions to copy files

anoukvlug:
This pull request addresses issue #467.

1. It adds copying cesm_data files to the option 'inversion' and 'run' of the copy_to_basedir function.
2. It adds an entity task for copying files with the copy_to_basedir function.

### DIFF
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -10,6 +10,10 @@ v1.X (unreleased)
 Breaking changes
 ~~~~~~~~~~~~~~~~
 
+- The utils.copy_to_basedir() function is changed to being an entity task. In
+  addition cesm_data files, when present, will from now on always be copied
+  when using this task (:issue:`467` & :pull:`468`).
+  By `Anouk Vlug <https://github.com/anoukvlug>`_.
 - Accumulation Area Ratio (AAR) is now correctly computed (:issue:`361`).
   By `Fabien Maussion <https://github.com/fmaussion>`_.
 - The method used to apply CRU and GCM anomalies to the climatology has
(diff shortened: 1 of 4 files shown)


fmaussion:
Thanks a lot! Actually, I think that making an entity task out of it makes much sense. 

Could you move the entire code to the task? I.e. the ``gdir.copy_to_basedir`` method should be deleted, and instead the task should contain all the code (this would be very similar code, but with ``gdir`` instead of ``self``). This might brake some tests, so you'll have to adapt them too.

That would be great!

anoukvlug:
Yes, I just did. Thanks for the instructions! I will update my PR in a bit :)

fmaussion:
This is excellent job, thanks! One last thing: can you add a short description of your change in [whats-new.rst](https://github.com/OGGM/oggm/blob/master/docs/whats-new.rst)? It is a small but breaking change, so it should be documented in the "breaking change" section

anoukvlug:
Thanks :) Yes, I can add some documentation to [whats-new.rst](https://github.com/OGGM/oggm/blob/master/docs/whats-new.rst) today . 

I was wondering, when I click on the issues and pull requests on the [whats-new.rst](https://github.com/OGGM/oggm/blob/master/docs/whats-new.rst) page these don't work as a link, is it supposed to be like that?

fmaussion:
> I was wondering, when I click on the issues and pull requests on the whats-new.rst page these don't work as a link, is it supposed to be like that?

Yes, they are rendered by sphinx when building the documentation. See: http://oggm.readthedocs.io/en/latest/whats-new.html

anoukvlug:
That makes a lot of sense :)

fmaussion:
Thanks!

PR #475, opened 2018-06-13
Attempt to fix coveralls

TimoRoth:


TimoRoth:
Something is broken on our end as well, producing tons of warnings. But it seems to be a bug internal to one of the involved cov packages.

The issue with the coverage results not showing up here is on their end, as the report on their website is there and has all the infos.

fmaussion:
Thanks for looking into this. I don't like coveralls much anyways, but not enough to switch to another tool so we'll just be patient.

PR #481, opened 2018-06-21
Add initial asv support

TimoRoth:
I'm not sure why the dependencies in setup.py were not set.
But they are 100% required for asv to operate at all, and having them setup properly seems like a good idea anyway.


fmaussion:
Thanks so much, this is great! 

About the setup: I was reluctant to force install the packages because (at least on my system) GDAL needs to be installed manually (with links). I wanted to avoid that people do a simple ``pip install oggm`` and expect this to work. 

But we have documented the install procedure more than enough, so I guess this is ok now.

fmaussion:
I think I should release salem also. It's been far too long already: doing a release now

TimoRoth:
The current situation with setup.py is that it installs oggm without an error even if it doesn't have any chance to then even be successfully imported.
So having the right dependencies in setup.py seems quite a vital sanity check to me.

PR #487, opened 2018-06-22
Enforce use of netcdf4 v1.3

fmaussion:
See: https://github.com/Unidata/netcdf4-python/issues/809


fmaussion:
We could change the code to set_mask to false everywhere, but I don't have the time right now. Let's see what the netcdf4 guys are saying. I still think this returning masked arrays per default is a bad idea

fmaussion:
... or, we set_auto_mask to false per default everywhere. It's just that it's so error prone, and very likely to be forgotten next time.

PR #496, opened 2018-07-03
Remove old histalp task

fmaussion:
- [x] Tests added/passed

This task was never used for anything but testing. The task was the only one making use of joblib, so I removed the dependency also (in hope that we won't need it in the future)


fmaussion:
EDIT: salem needs joblib so we keep it

PR #502, opened 2018-07-09
Added sampling without replacement to RandomMassBalance

matthiasdusch:
Added an option to the RandomMassBalance model:
If without_replacement==False (default option and behavior of RandomMassBalance up till now), every model year an year from the given climate period will be picked, independent on the previous picks.
A 100-year model run with a given 30 year climate period could theoretically use the climate of 1 specific year for 100 consecutive model years.

If without_replacement==True (new option), all years from the given climate period will be used in random order before one year can be picked again.
A 300-year model run with a given 30 year climate will use these 30 years for 10 consecutive times but each time randomly shuffled.
A 100-year model run with a given 100 year climate will use every climate year exactly once, but randomly shuffled compared to the original climate series.

- no Tests added
- Documented in massbalance/RandomMassBalance and flowline/run_random_climate

### DIFF
--- a/oggm/core/flowline.py
+++ b/oggm/core/flowline.py
@@ -1576,7 +1576,7 @@ def run_random_climate(gdir, nyears=1000, y0=None, halfsize=15,
                        climate_input_filesuffix='',
                        output_filesuffix='', init_model_fls=None,
                        zero_initial_glacier=False,
-                       without_replacement=False,
+                       unique_samples=False,
                        **kwargs):
     """Runs the random mass-balance model for a given number of years.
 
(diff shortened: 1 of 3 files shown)


fmaussion:
Thanks a lot! This is quite clever, it took me some time to understand the algorithm here ;)

I'm not sure I like the name of the kwarg, but I don't have a brilliant idea either...

How about ``ensure_unique``? It is not strictly correct for periods > window size, but it is still more explicit than the current name. Thoughts?

fmaussion:
... and I would welcome some tests - we can discuss a strategy when we meet

matthiasdusch:
Updated the pull request:

- Changed parameter name from `without_replacement` to `unique_samples`
- Added some more comments on the sampling algorithm
- Added tests: `test_models/test_random_mb_unique`. Test structure is adapted from `test_models/test_random_mb` with some adjustments.

fmaussion:
Excellent job, thanks a lot!

PR #505, opened 2018-07-18
Changed SRTM download urls

### LINKED ISSUE #504: SRTM download-links are dead
The url to download SRTM topo-files within `oggm.utils._download_srtm_file_unlocked` seems to be deprecated.

See for example: http://droppr.org/srtm/v4.1/6_5x5_TIFs/srtm_06_01.zip

Possible fixes:
a) Find open access and reliable 3rd party download source
b) Host all files on our own
c) Get them directly from USGS (I think this needs an account)

For c) [usgs api](https://mapbox.github.io/usgs/index.html) might be useful.

matthiasdusch:
- [X] Closes #504 
 - [X] Tests already exist

Changes the dead 'droppr.org' url to 'http://srtm.csi.cgiar.org/SRT-ZIP/SRTM_V41/SRTM_Data_GeoTiff/' as proposed by @DavidParkes. 

A test to download a SRTM file already exists (`test_srtmdownload`).
But this test will not download the file if it still exits in the cache. This should probably be adapted.

### DIFF
--- a/oggm/tests/test_utils.py
+++ b/oggm/tests/test_utils.py
@@ -484,7 +484,8 @@ class TestFakeDownloads(unittest.TestCase):
                               fakefile='srtm_39_03.tif')
 
         def down_check(url, cache_name=None, reset=False):
-            expected = 'http://droppr.org/srtm/v4.1/6_5x5_TIFs/srtm_39_03.zip'
+            expected = 'http://srtm.csi.cgiar.org/SRT-ZIP/SRTM_V41/' + \
+              'SRTM_Data_GeoTiff/srtm_39_03.zip'
             self.assertEqual(url, expected)
             return tf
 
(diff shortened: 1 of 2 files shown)


fmaussion:
> A test to download a SRTM file already exists (test_srtmdownload).
> But this test will not download the file if it still exits in the cache. This should probably be adapted

It should, if you either comment out the ``@is_download`` decorator or set 

https://github.com/OGGM/oggm/blob/bb666cab9436f235f900f8106774f2377969752c/oggm/tests/__init__.py#L42

to ``True``. 

This is for testing locally though, we don't download files from Travis

fmaussion:
Thanks! Going to merge this but during the time Timo is on holidays the cluster cache won't be updated, which is bad -> @OGGM/oggm-devs don't update OGGM on the cluster, or revert this PR temporarily.

Issue #483, opened 2018-06-21
Suppress the time consuming compute_ref_tstars task


fmaussion:
This looked like a good idea at first but it's more confusing than helping (I think)

fmaussion:
closed via https://github.com/OGGM/oggm/pull/509

PR #509, opened 2018-07-31
Changes in climate and massbalance internals: part 1

fmaussion:
- [x] Closes https://github.com/OGGM/oggm/issues/483
 - [x] Tests added/passed

This PR is a first step towards some major changes in the mass-balance internals, not without consequences on the results of the model (unfortunately)

Some of the most important changes:
- temperature gradients are now added to the climate files only if computed from local regression (never used operationally). This was done in order to treat the temperature gradient parameter like any other parameter for the the model cross-validation and calibration
- precipitation factors are now *also* treated as a normal MB model parameter. They had a special treatment in the previous implementation because I thought it would be possible to decide of the factor automatically, but this never worked out. Hence, the code is now simplified and there is no more difference between, say, precip factor and temperature gradient or melt temperature
- when more than one tstar is available (the case for many glaciers), we now take the one with the smallest absolute bias (what Ben does) instead of the fancy global search I was doing before. This makes the results MUCH more predictable (and reproducible), and makes the cross-validation faster. **This is what will have strong consequences on the model output, and needs to be evaluated**
- the task ``distribute_tstar`` is now depcrecated in favor of the entity task ``local_mustar``, which does 
  exactly the same job but in parallel

Altogether this goes in the direction of the Zen: more explicit, simpler code.


appveyor.yml
```diff
@@ -24,6 +24,9 @@ install:
   - "SET PATH=%PYTHON%\\Scripts;%PYTHON%;%PATH%"
   - "SET GDAL_DATA=%PYTHON%\\Library\\share\\gdal"
 
+  # Tell appveyor to run the slow tests
+  - "SET OGGM_SLOW_TESTS=true"
```
TimoRoth: You can safely drop this now

fmaussion:
@TimoRoth are you sure that the --mpl option is used on Travis currently?

TimoRoth:
Seems like moving the script to an actual script broke that, since Docker didn't forward it so far.
Now it does, so it should work again.

fmaussion:
In it is!

PR #513, opened 2018-08-06
Use multiprocessing Pool.map for execute_parallel_tasks

### LINKED ISSUE #512: execute_parallel_tasks does not work properly on Windows
Discovered in https://github.com/OGGM/oggm/pull/509 : my homegrown parallelization for a single entity task on a single glacier:

https://github.com/OGGM/oggm/blob/704fd9fd1d8b4b7d8c83226b8da97919a9b55837/oggm/workflow.py#L147-L195

does not work properly on windows: the config is not passed properly to the workers

TimoRoth:
Fixes #512

### DIFF
--- a/oggm/workflow.py
+++ b/oggm/workflow.py
@@ -76,18 +76,23 @@ class _pickle_copier(object):
         self.call_func = func
         self.out_kwargs = kwargs
 
-    def __call__(self, gdir):
+    def __call__(self, arg):
+        if self.call_func:
+            gdir = arg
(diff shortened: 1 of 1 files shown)


fmaussion:
Thanks!

I don't remember why, but at the time of writing I didn't use `` mppool.map`` for a reason - but there might be a trick to get it work. Currently the test is failing...

fmaussion:
yay! Please merge when Appveyor is happy

PR #514, opened 2018-08-06
defined rgi_area_km2 as (lazy) property and changed how it is determined

matthiasdusch:
This change should only effect you if `cfg.PARAMS['use_rgi_area'] == False` (where `True` is the default setting).

Previously if a none-rgi-area was used, it was not correctly set in the GlacierDirectory.

- If `cfg.PARAMS['use_rgi_area'] == False`: The glacier area will be calculated from the polygon geometry within `gis.define_glacier_region`.
- If `cfg.PARAMS['use_rgi_area'] == True`: The area will not be calculated, instead the RGI database value will be used. This was also the default behavior up till now.

In both cases the glacier area will be written to the glaciers outline shapefile. And now `rgi_area_km2` is a (lazy) property of a GlacierDirectory and will open the shapefile and read the area.

 - [x] Tests did already exist

### DIFF
--- a/oggm/utils.py
+++ b/oggm/utils.py
@@ -2906,8 +2906,8 @@ class GlacierDirectory(object):
     @lazy_property
     def rgi_area_km2(self):
         """The glacier's RGI area (km2)."""
-        _area = gpd.read_file(self.get_filepath('outlines'))['Area'].values[0]
-        return float(_area.round(decimals=3))
+        _area = gpd.read_file(self.get_filepath('outlines'))['Area']
+        return np.round(float(_area), decimals=3)
 
     @property
     def rgi_area_m2(self):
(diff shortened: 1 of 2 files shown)


oggm/utils.py
```diff
@@ -2904,6 +2903,12 @@ def grid(self):
         """A ``salem.Grid`` handling the georeferencing of the local grid"""
         return salem.Grid.from_json(self.get_filepath('glacier_grid'))
 
+    @lazy_property
+    def rgi_area_km2(self):
+        """The glacier's RGI area (km2)."""
+        _area = gpd.read_file(self.get_filepath('outlines'))['Area'].values[0]
```
fmaussion: Change to ``float(gpd.read_file(self.get_filepath('outlines'))['Area'].values[0])``

fmaussion: or ``float(_area).round(decimals=3)`` at the line below, perhaps easier to read...

matthiasdusch:
Took care of the review.

But one error as shown by Travis: `rgi_area_km2` is used within `GlacierDirectory.__repr__`, and therefore raises an IOError if called before `define_glacier_region`. This will need further work.

A simple if-file-exists logic might not work because of the lazy_property behavior or am I mistaken?

fmaussion:
Thanks a lot!

PR #515, opened 2018-08-06
Switch to native pytest.mark annotations and options

TimoRoth:
This gets rid of most of the OGGM_* environment variables for testing and switches to pytest.mark native markings.


fmaussion:
Wow this is great!

This is a big change though, and I have to check a few things:
- it is important that all slow tests are run on Travis (the previous default), which doesn't seem to be the case now
- we should run the slow tests on appveyor too (it wasn't the default but could become it: the build time will be *much* longer but it's probably worth it) 
- what happens now when we run ``pytest .`` locally? The previous behavior was to run all the tests except the slow tests and download tests. I think it is a meaningful default, and ``pytest --run-test-env slow`` should run all the tests except the download tests (like appveyor)

fmaussion:
I like it when a change brings less lines of code :+1: 

selection_065

fmaussion:
Looks good now, thanks a lot! Much better than the previous setup

>  iirc last time it ran into a timeout on appveyor.

OK, good point. Let's see if this is sustainable, otherwise skip.

fmaussion:
The failing tests are expected and corrected in master and https://github.com/OGGM/oggm/pull/509 

Now the question is: are 36 mins test builds OK?

TimoRoth:
The image mismatch on test_modelsection_withtrib also happens on Anaconda on Linux right now btw.
36 Minutes seem fine to me. If it starts getting a problem, you(since the oggm appveyor repo is on your account name) can go to https://github.com/appveyor/ci/issues/517 and ask for more time, but right now things seem to run fine on the normal limit.

TimoRoth:
I don't think the tests are corrected in master, since I rebased this on master very recently. I'd go ahead and merge this now, and then start fixing the tests individually, as this PR is now what broke them.

fmaussion:
Yes please merge. I'll take care of the test by cherry picking in https://github.com/OGGM/oggm/pull/509

PR #517, opened 2018-08-08
Also test with Python 3.7

TimoRoth:


fmaussion:
Thanks! Could you post the failing images here?

TimoRoth:
test_modelsection
test_modelsection_withtrib

You can test yourself by using the oggm/untested_base:py37 Docker image.

TimoRoth:
Those seem to match with the old-freetype images. I think the difference in them is not caused by freetype at all, since the graphs are actually different. Must be some other version that somehow is different between the two docker containers, even though they are pypi-package-wise 100% identical.

fmaussion:
>  the graphs are actually different.

This is fascinating and a bit worrying I must say :-(

TimoRoth:
I have a suspicion it might by pyproj, since I'm using the latest git version on python3.7 one, as the latest release from 2016 doesn't work. That's the only real difference.

TimoRoth:
Yes, it definitely is the pyproj version. Now, which one is more correct?

fmaussion:
> Yes, it definitely is the pyproj version. Now, which one is more correct?

No idea :-( I guess I'll have to make some tests to track this down... OR we use the latest and that's all

TimoRoth:
The thing is, the "latest" released version is quite old, from 2016. And it's only that compared to git master that's different.

TimoRoth:
I'm at a loss as to why appveyor fails to run the tests. It just simply fails to find the matplotlib pytest plugin, no matter how hard I point it to it.

fmaussion:
I would worry about this: we can just switch off ``--mpl`` on windows, don't you think? Will make our life easier

TimoRoth:
With the changes I had in mind there wouldn't be any --mpl anymore, because of just using the matplotlib native comparator. But it breaks on windows for no obvious reason.
I decided to fork pytest-mpl instead and make it less broken instead...

PR #527, opened 2018-08-17
Update pep8 config

fmaussion:
This is just to see if the police comes or not


fmaussion:
This matches the defaults in pandas

PR #528, opened 2018-08-22
Cleanup tests a bit more

TimoRoth:
Avoids a lot of duplicated parameters in various places.
Also allows for future features like tests for the oggm version of pytest-mpl.


TimoRoth:
I have no idea why this keeps failing on Windows, something about those two tests must produce drastically different results there.

fmaussion:
Can we rerun the tests with the newest pyproj? ( https://github.com/conda-forge/pyproj-feedstock/issues/10#issuecomment-416680815 ) Then I will have a look

PR #536, opened 2018-09-24
updated notebook to work with recent changes

### LINKED ISSUE #535: Notebook "Glacier dynamics and length changes"
Why in the run_from_climate_data function I need to set up the parameter 'ys' to 1902 whereas HISTALP climate data starts earlier ?

Here is the error: 2018-09-24 16:52:49: oggm.core.flowline: ValueError occurred during task run_from_climate_data hist_from_current on RGI50-11.00897: year 1855.0 out of the valid time bounds: [1902, 2016]

matthiasdusch:
- [X ] Closes #535 

Also fixed 1 issue in the same Notebook, by switching from `cfg.A` to `cfg.PARAMS['glen_a']`.

I might test the other Notebooks at "docs" within the next days.


fmaussion:
Thanks!

Issue #478, opened 2018-06-14
Allow different mu's for different flowlines


fmaussion:
This is something I've been considering for a while now. Recent works by @matthiasdusch made it more urgent.

### Problem

The major issue with the way we compute mu currently (one mu per glacier) is that sometimes, a single mu cannot explain the existence of all the branches in a glacier. Some lower lying tributaries have a negative mass-balance and therefore a negative mass-flux for the inversion. 

### Current workarounds 

This problem lead to two waves of "improvements" in the model:
- the first solution (now deprecated but still available as option in the config file) is to correct for these negative fluxes *for the inversion only* and ignore them the rest of the time. This was later discussed as being not good because not physically consistent and leading to issues of mass conservation in the t* experiments (see GMD paper)
- the second solution (current default) is to *suppress* these unphysical tributaries. This seemed like a good idea at the time and surely is more mass-conservative, but it is clunky and leads to several other problems I don't want to list here (mostly related to the OGGM workflow itself)

### A better solution?

A solution which would take the best out of the two workarounds above it to allow for different mus for each flowline. With this mechanism, a tributary which cannot be explained by the glacier-wide mu would get a different climate sensitivity. This mu would then be used for the inversion **and** for the ice-dynamics runs, therefore ensuring consistency. 

This is I think the better solution of all three. The only problem with it is: how to choose mu? The simplest way would by to treat this lonesome tributary as a single glacier and define mu so that its tributary-wide mass-balance is equal to zero. This would lead to a zero mass-flux into the underlying main branch (which is clearly wrong) but would be extendable to better solutions in the future.

### Consequences

The consequences for the model internals will be consequent, but manageable. The default behavior of one mu per glacier should still remain available, internally it will be handled as a multiple mu case though.

A positive consequence will be that it is going to be relatively straightforward to allow the dynamical model to merge glacier complexes before runtime, i.e. by merging all glaciers in a single valley together.

### When?

I am quite keen on starting this ASAP, but this might take more time than expected. Let's see...

fmaussion:
Closed via https://github.com/OGGM/oggm/pull/539

bearecinos:
Hi, 
I spent sometime yesterday updating OGGM and my frontal ablation code. There is no error per say but when I assign the new calving rate to the MB model there seems to be not change in Volume at all after the inversion happens. That is really odd, I will like to ask where is my mistake on using this functions?

```
gdir.inversion_calving_rate = F_calving

                # Recompute mu with calving
[...]

PR #539, opened 2018-09-29
WIP: Allow different mu's for different flowlines

fmaussion:
- [x] Fixes https://github.com/OGGM/oggm/issues/478
 - [x] Tests added/passed

This is a longstandig issue and not an easy one. 

I *think* I've managed to find an elegant way to deal with the calibration: it uses recursive calls to the calibration to calibrate the flowline's mu* until the entire glacier is done.

This still needs more test, and the mass-balance models need an update, hence still a WIP


matthiasdusch:
The recursive approach is indeed elegant.

I have one comment/question regarding readability and also performance:
With this commit, only negative-flux-flowlines will get a different `mu_star`. So if all tributary flowlines are good, it should be possible to immediately use the already calculated `mu_star_glacierwide`. The `_mu_star_per_minimization` would then only be called on recursive calls if one or more flowlines need corrections.

But I might miss a point here or you have already some future adaptions in mind.

fmaussion:
Thanks for looking into it!

Yes, there are some issues regarding performance here. Even the call to "minimize" is superfluous altogether since the problem is linear in mu anyway. 

However, I would like to avoid doing premature optimizations, following the good old adage*. They add complexity (here, at least one "if" and probably more), and we don't know yet if it's so slow. 



*: "The First Rule of Program Optimization: Don’t do it. 
The Second Rule of Program Optimization (for experts only!): Don’t do it yet"

fmaussion:
This is slowly taking shape - I have added a new class to the mass-balance models which handles multiple flowlines. I spent the entire day on it - not because it's a lot of code (it's not that much code at all :D) but because it was quite hard to find the right API. I hope this is the right one now...

Anyways, still need to make this new MB the default in the production runs and also need to get the cross-validation working.

conftest.py
```diff
@@ -1,6 +1,3 @@
-import matplotlib
-matplotlib.use('Agg')  # noqa: E402
```
matthiasdusch: shouldn't  this be part of PR #542 ?

fmaussion: ups. Yes.

oggm/core/massbalance.py
```diff
@@ -785,7 +832,12 @@ def get_annual_mb(self, heights, year=None, fl_id=None):
 
         return self.mb_models[fl_id].get_annual_mb(heights, year=year)
 
-    def get_specific_mb(self, fls=None, year=None):
+    def get_specific_mb(self, heights=None, widths=None, fls=None,
+                        year=None):
+
+        if heights is not None or widths is not None:
+            raise ValueError('`heights` and `heights` kwargs do not work with '
```
matthiasdusch: `heights` and `widths` kwargs...

fmaussion: thanks!

fmaussion:
OK, the last commit is bigger than I wish it would, BUT:
- I now added an option to compute the reference t* using the multiple flowlines approach. It is much slower than using the glacier wide one for only minimal differences in the bias (not even in the t*)! So this was done for nothing, other than helping my conscience ;-). The default remains to use glacier-wide t* calibration. 
- The climate workflow has changed it's name, I'd be happy to hear about your thoughts on this
- compute_ref_tstar is now done in parallel

So, why changing the API again? Let me try to explain.

**Before**, `mu_candidates` and `t_star_from_refmb` were separate tasks. It didn't make much sense, because they always needed to be called together:  `t_star_from_refmb` needs `mu_candidates`  and `mu_candidates`  _without_ `t_star_from_refmb` is useless.

**Now**, the mu candidates are computed and used in the same task (which kept its name): `t_star_from_refmb` . The `mu_candidates` task is still available for test and documentation purposes, but it is not used in a normal workflow. To make sure there is no confusion, I renamed it to `glacier_mu_candidates` (could be another name though).

Since I was changing things already, I reconsidered the names of the other climate calibration tasks:
- `local_mustar` is actually interpolating `t*` from the reference table. For convenience, it is also calculating the glacier wide mu* 
- `apparent_mb` was using the local mu* to add the mass flux to the branches for the inversion. Now, it is actually calculating the multiple flowline's mu* 

In my last commit, I renamed `local_mustar` to `local_tstar` and `apparent_mb` to `mu_star_calibration` to better reflect what they do.

Now that I think about it, I believe they could even be merged in one big single task, which does both the interpolation of t* and the mu* calibration in one step. I'll consider this option, more on this later...

fmaussion:
@TimoRoth for some reason the graphics tests pass with 3.7 but not with 3.6 -> this is suspicious .

Any idea what's going on?

fmaussion:
OK, Travis is green, let's merge this - sorry @OGGM/oggm-devs for all the changes. I suggest not to update until V1.1 is out (next couple of weeks), after which we will stabilize the code base (I promise ;)

bearecinos:
Sorry again.... one question while using this` local_t_star() `

In my calving loop I do this:
```
    for gdir in gdirs:
        if gdir.terminus_type == 'Marine-terminating': ...... # a bunch of stuff then

                     # Lets compute the theoretical calving out of it
                     pre_calving = k * thick * w_depth * width / 1e9
                     gdir.inversion_calving_rate = pre_calving
                     print('pre_calving', pre_calving)

                     # Recompute all with calving
                     tasks.distribute_t_stars([gdir], minimum_mustar=0.)
                     tasks.apparent_mb(gdir)
                     tasks.prepare_for_inversion(gdir, add_debug_var=True)
                     tasks.volume_inversion(gdir, glen_a=cfg.A, fs=cfg.FS)
                     df = pd.read_csv(gdir.get_filepath('local_mustar')).iloc[0]
                     mu_star = df['mu_star']

```

Several questions: 
1. The minimum_mustar now I can set it up at the beginning with:  `cfg.PARAMS['min_mu_star'`]
2. after running this tasks.mass_conservation_inversion, instead of volume_inversion I want to get the local_mustar file in the glacier directory. Before it was a csv ... how can I read this new .json file?

PR #542, opened 2018-10-07
Do not force Agg backend in conftest

fmaussion:
@TimoRoth any reason why this was added? This just drove me crazy as I was looking for a reason as why my plots didn't show up while developing tests. If needed it has to be hidden behind a "if on machine" block. Also, we can't do this silently, this is too intrusive.

I've seen there are a couple of other ``matplotlib.use('Agg')`` in the codebase: in doc and benchmark builds. Leaving them here for now

### DIFF
--- a/benchmarks/__init__.py
+++ b/benchmarks/__init__.py
@@ -1,2 +1,0 @@
-import matplotlib
-matplotlib.use('Agg')
--- a/conftest.py
+++ b/conftest.py
@@ -1,6 +1,3 @@
-import matplotlib
-matplotlib.use('Agg')  # noqa: E402
-
 import pytest
 import logging
 import getpass


TimoRoth:
Otherwise it defaults to the QtAgg backend, which makes the tests and benchmarks fail on headless machines without Qt installed.

fmaussion:
Yes, but then there should be a kind of try ... catch close that we could use (or maybe even a matplotlib function?) to check if we have to set the Agg backend or not, and set it only if necessary

TimoRoth:
The problem is that the backend cannot be changed at runtime. It has to be set at the very start before any other mpl functions have been called. Which is why it's so high up and in conftest in the first place.

fmaussion:
> Otherwise it defaults to the QtAgg backend, which makes the tests and benchmarks fail on headless machines without Qt installed.

Are you sure about this? I think that if you don't *call* plt.show() but just plt.savefig() it shouldn't make an error.

TimoRoth:
Pretty sure, but this should still be mergeable, as the testing and benchmarking environment variable sets the backend via environment variable. Just don't set the QtAgg backend anywhere manually.

PR #562, opened 2018-10-19
Don't use "sudo pip", rely on system packages instead

TimoRoth:
Later versions of Ubuntu and Debian disabled pip as root for a good reason. It can seriously damage the system python installation.
Use system packages instead.


fmaussion:
Nice! Much easier anyway

PR #579, opened 2018-10-24
WIP: doc upgrades

fmaussion:


TimoRoth:
Why are the benchmarks trying to run for this? The Webhook is not configured to send Pull Requests, and the service is clearly not prepared for it.

fmaussion:
because its a local branch - you should only benchmark master I guess

TimoRoth:
Ah! Should be fixed now.

PR #581, opened 2018-10-25
Move conftest down a level to fix testing installed oggm

TimoRoth:
Let's see if this still breaks the tests.


TimoRoth:
I assume it's the two that require rasterio >= 1.0. The test env is forced to 0.36. I guess I can remove that now?

fmaussion:
> I assume it's the two that require rasterio >= 1.0. The test env is forced to 0.36. I guess I can remove that now?

Yes, since rio 1+ is now stable, we can finally move on ;-)

Issue #469, opened 2018-05-31
Running from GCM 


anoukvlug:
After some off-line discussions, with @fmaussion and @nchampollion, about how OGGM can be forced with different GCM's, I thought it would be good to open an issue about this.

Currently temperature and precipitation of CESM simulations can be used to force OGGM. Different GCM's structure their output slightly differently. Therefore the process_cesm_data function can currently not process files from other GCM's. There are a couple of ways to make it possible to use other GCM's in OGGM.

**Here follows a list with some options (there are probably more options), to continue the conversation:**

1. Make the process_cesm_data function more flexible. This would require quite some keyword arguments in the function. 

2. Add different functions for the different GCM's.

3. Write documentation on how people can make their own function to treat their input data.

4. Make a general process_gcm_data function, with documented how the input should look like. This could than include some documentation on how to make these basic changes to the input (netCDF) files by using NCO tools (e.g. summing different type of precipitation to get a file with the total precipitation or changing variable names).

**Disadvantages of the different options:**
1. There might be many things that differ between the models. To make one function that can treat all GCM's might be hard and I am afraid that it could be sensitive to bugs.

2. This adds a lot of code repetition to the model.

3. This makes people repeat what other people have already done.

4. It puts problem of having difference in how the GCM's structure their output with the user.

fmaussion:
Very good points! I will spend the week-end at the ECMWF [hackathon](https://events.ecmwf.int/event/79/overview), and I might come back with some ideas. Let's keep in touch!

nchampollion:
Yes, good points ! I'll also come later to this issue but my first thought is that a mix between options 1 & 3 seems reasonable. Solution 2 not feasible since GCM landscape change over time ... and solution 4 is probably the best solution but require lot of time and much more complex.

fmaussion:
One of the first thing I am thinking about changing is the computation of the anomalies. Currently the anomalies are computed by the ``process_cesm_data`` but I would like to have a task especially for this. 

the reasoning is that this part of the code is very specific to our way to compute the mass-balance and also the same for any kind of data we are using. The step which is difficult to standardize is the input (gridded data -> time series at the glacier location): after this is done (i.e. by an external tool like Hugues did), OGGM should provide the rest of the tools

fmaussion:
[...]

PR #580, opened 2018-10-24
WIP: Added GCM edits and process_equil_ccsm_data prepro function

sumnonpuella:
I made the edits suggested in #567, and added the process_equil_ccsm_data function i've been working on. So far, the function only properly works with one year of data, and i'm working on more than transient data, but I wanted to get the PR out to keep it from growing to much more. This is my first big PR - any constructive advice is welcome!

- [X] Closes #566 
- [X] Closes #567  
- [ ] Tests added/passed
- [X] Fully documented
- [ ] Entry in `whats-new.rst`

Issue #567, opened 2018-10-22
Baseline GCM data


fmaussion:
`BASENAMES['cesm_data']` should be renamed to `BASENAMES['gcm_data']` (see https://github.com/OGGM/oggm/issues/566 for context).

https://github.com/OGGM/oggm/blob/0a25564d9e2c306812566498cf99dd3595960802/oggm/cfg.py#L171-L172

PR #582, opened 2018-10-25
splits process_cesm_data() into 2 functions

anoukvlug:
This PR is a result of the OGGM hack day. It is meant to contribute to issue https://github.com/OGGM/oggm/issues/469. This PR is not meant as a final solution, but more as first step and as food to continue the discussion. So don't hesitate to comment on it @fmaussion @nchampollion @sumnonpuella  

- [x] Tests added/passed
- [ ] Fully documented

With this PR `process_cesm_data()` is split into 2 functions. `prepro_cesm_data` and `process_gcm_data()`. `process_gcm_data` is meant to make it easier to run OGGM with the climate of other GCM's. `prepro_cesm_data` replaces `process_cesm_data()`. `prepro_cesm_data` uses the `process_gcm_data()` to generate the climate files for OGGM. It also can be used as an example on how to use other GCM's as input.

- `prepro_cesm_data` is placed is a new file. I am not sure what the best name and location for this file would be.

### DIFF
--- a/oggm/core/climate.py
+++ b/oggm/core/climate.py
@@ -112,51 +112,65 @@ def process_custom_climate_data(gdir):
 
 @entity_task(log, writes=['gcm_data'])
 def process_gcm_data(gdir, filesuffix='', prcp=None, temp=None,
-                     time_unit=None, time2=None):
+                     time_unit='days since 1801-01-01 00:00:00',
+                     calendar=None):
     """ Applies the anomaly method to the climate data and stores the data in a
     format that can be used by the OGGM mass balance model.
+
+    For an example implementation see
+    :ref:`core.climate_prepro.process_cesm_data()`
 
     Parameters
     ----------
     filesuffix : str
         append a suffix to the filename (useful for ensemble experiments).
     prcp    : xarray.DataArray - format:
-        monthly total precipitation [mm]
+        monthly total precipitation [mm month-1]
             Coordinates:
             lat      float64
             lon      float64
-            time     (time) object
-            month    (time) int64
-            year     (time) int64
-            dates    (time) object format YYYY-MM-DD hh:mm:ss
+            time     (time) cftime object
     temp : xarray.DataArray
         monthly temperature [K]
             Coordinates:
             lat      float64
             lon      float64
-            time     (time) object
-            month    (time) int64
-            year     (time) int64
-            dates    (time) object format YYYY-MM-DD hh:mm:ss
+            time     (time) cftime object
     time_unit : str
-            format: days since e.g.: days since 0850-01-01 00:00:00
-
-    for an example see: climate_prepro.process_cesm_data()
-
+        The unit conversion for NetCDF files. It must be adapted to the
+        length of the time series.
+        For example: 'days since 0850-01-01 00:00:00'
+    calendar : str
+        If you use an exotic calendar (e.g. 'noleap')
     """
+
+    # Standard sanity checks
+    months = temp['time.month']
+    if (months[0] != 1) or (months[-1] != 12):
+        raise ValueError('We expect the files to start in January and end in '
+                         'December!')
+
+    if (np.abs(temp['lon']) > 180) or (np.abs(prcp['lon']) > 180):
+        raise ValueError('We expect the longitude coordinates to be within '
+                         '[-180, 180].')
+
+    # from normal years to hydrological years
+    sm = cfg.PARAMS['hydro_month_' + gdir.hemisphere]
+    prcp = prcp[sm-1:sm-13].load()
+    temp = temp[sm-1:sm-13].load()
 
     # compute monthly anomalies
     # of temp
-    ts_tmp_avg = temp.sel(time=(temp.year >= 1961) & (temp.year <= 1990))
-    ts_tmp_avg = ts_tmp_avg.groupby(ts_tmp_avg.month).mean(dim='time')
-    ts_tmp = temp.groupby(temp.month) - ts_tmp_avg
+    ts_tmp_avg = temp.sel(time=slice('1961', '1990'))
+    ts_tmp_avg = ts_tmp_avg.groupby('time.month').mean(dim='time')
+    ts_tmp = temp.groupby('time.month') - ts_tmp_avg
     # of precip -- scaled anomalies
-    ts_pre_avg = prcp.isel(time=(prcp.year >= 1961) & (prcp.year <= 1990))
-    ts_pre_avg = ts_pre_avg.groupby(ts_pre_avg.month).mean(dim='time')
-    ts_pre_ano = prcp.groupby(prcp.month) - ts_pre_avg
+    ts_pre_avg = prcp.sel(time=slice('1961', '1990'))
+    ts_pre_avg = ts_pre_avg.groupby('time.month').mean(dim='time')
+    ts_pre_ano = prcp.groupby('time.month') - ts_pre_avg
     # scaled anomalies is the default. Standard anomalies above
     # are used later for where ts_pre_avg == 0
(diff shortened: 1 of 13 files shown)


sumnonpuella:
@anoukvlug I like the idea a lot! Do you see then see other GCM prepro functions being added to `climate_prepro.py`? That might help keep `climate.py` clean. Once this is merged, I might copy the structure for the `process_equil_ccsm` function.

oggm/core/climate.py
```diff
@@ -110,87 +110,39 @@ def process_custom_climate_data(gdir):
...
-    from Otto-Bliesner et al. (2016), to the high-resolution CL2 climatologies
-    (provided with OGGM) and writes everything to a NetCDF file.
+@entity_task(log, writes=['gcm_data'])
+def process_gcm_data(gdir, filesuffix='', precp=None, temp=None,
+                     time_unit=None, time2=None):
+    ''' Applies the anomaly method to the climate data and stores the data in a
+    format that can be used by the OGGM mass balance model.
 
-    Parameters
+        Parameters
```
fmaussion: Remove indent here

oggm/core/climate.py
```diff
@@ -110,87 +110,39 @@ def process_custom_climate_data(gdir):
...
-    """Processes and writes the climate data for this glacier.
-
-    This function is made for interpolating the Community
-    Earth System Model Last Millenial Ensemble (CESM-LME) climate simulations,
-    from Otto-Bliesner et al. (2016), to the high-resolution CL2 climatologies
-    (provided with OGGM) and writes everything to a NetCDF file.
+@entity_task(log, writes=['gcm_data'])
+def process_gcm_data(gdir, filesuffix='', precp=None, temp=None,
+                     time_unit=None, time2=None):
+    ''' Applies the anomaly method to the climate data and stores the data in a
```
fmaussion: Use `"""` instead of `'''` (for consistency)

oggm/core/climate.py
```diff
@@ -110,87 +110,39 @@ def process_custom_climate_data(gdir):
...
-def process_cesm_data(gdir, filesuffix='', fpath_temp=None, fpath_precc=None,
-                      fpath_precl=None):
-    """Processes and writes the climate data for this glacier.
-
-    This function is made for interpolating the Community
-    Earth System Model Last Millenial Ensemble (CESM-LME) climate simulations,
-    from Otto-Bliesner et al. (2016), to the high-resolution CL2 climatologies
-    (provided with OGGM) and writes everything to a NetCDF file.
+@entity_task(log, writes=['gcm_data'])
+def process_gcm_data(gdir, filesuffix='', precp=None, temp=None,
```
fmaussion: For consistency with the rest of the code, can you call this `prcp` instead of  `precp`?

oggm/core/climate.py
```diff
@@ -110,87 +110,39 @@ def process_custom_climate_data(gdir):
...
+        monthly temperature [K]
+            Coordinates:
+            lat      float64
+            lon      float64
+            time     (time) object
+            month    (time) int64
+            year     (time) int64
+    time_unit : str
+            format: days since e.g.: days since 0850-01-01 00:00:00
+    time2 : numpy.ndarray
```
fmaussion: what's `time2` here? I think another name and doc would help. Maybe we can even don't use it and extract it from the dataarrays?

oggm/core/climate.py
```diff
@@ -110,87 +110,39 @@ def process_custom_climate_data(gdir):
...
+
+    for an example see: climate_prepro.process_cesm_data()
+
+    '''
```
fmaussion: `"""`

oggm/core/climate_prepro.py
```diff
@@ -0,0 +1,122 @@
...
+    Earth System Model Last Millennium Ensemble (CESM-LME) climate simulations,
+    from Otto-Bliesner et al. (2016), to the high-resolution CL2 climatologies
+    (provided with OGGM) and writes everything to a NetCDF file.
+
+    Parameters
+    ----------
+    filesuffix : str
+        append a suffix to the filename (useful for ensemble experiments).
+    fpath_temp : str
+        path to the temp file (default: cfg.PATHS['gcm_temp_file'])
```
fmaussion: Here (and below) I guess we should change if for `cfg.PATHS['cesm_temp_file']` . What do you think?

oggm/core/climate_prepro.py
```diff
@@ -0,0 +1,122 @@
...
+
+    # load dates in right format to save
+    dsindex = salem.GeoNetcdf(fpath_temp, monthbegin=True)
+    time1 = dsindex.variables['time']
+    time2 = time1[sm-1:sm-13] - ndays  # to hydrological years
+    time2 = netCDF4.num2date(time2, time1.units, calendar='noleap')
+    time_unit = time1.units
+
+    return process_gcm_data(gdir, filesuffix=filesuffix, precp=precp,
+                            temp=temp, time_unit=time_unit, time2=time2)
```
fmaussion: I don't think we should return anything here (`process_gcm_data` doesn't have a return values).

In all cases, you should not return from the function *before* closing the files. I would recommend to shift the lines below to above, and the last call of the function should be `process_gcm_data(...)`

oggm/tasks.py
```diff
@@ -21,7 +21,7 @@
 from oggm.core.climate import process_cru_data
 from oggm.core.climate import process_histalp_data
 from oggm.core.climate import process_custom_climate_data
-from oggm.core.climate import process_cesm_data
+from oggm.core.climate import process_gcm_data
```
fmaussion: here you can add your new task as well: `from oggm.core.climate_prepro import process_cesm_data`

anoukvlug:
@fmaussion thanks for doing the review so fast :) I agree the changes you suggest and just implemented some of them. The remaining part I will try to include later this week.

oggm/core/climate.py
anoukvlug: I agree that it is very unclear in this way, I include the doc and another name. I was thinking maybe `time_dates` would be better. 

`time2` is the variable `time` that is needed in the `write_monthly_climate_file()`. I agree that it would be better if the time could be extracted from the dataarrays. I only don't know yet how to include it.

fmaussion: OK. You can give it a try and if not I'll have a look

fmaussion:
@anoukvlug this looks great! Are you still working on it? If not I'd like to go over it one last time tomorrow morning and merge it ;-)

anoukvlug:
Thank you @fmaussion! 

I am not working on the PR anymore. Though I am not sure if it should be merged already. I did some renaming in this PR in anticipation of PR #580, that like this PR also addresses issue #567. I thought about adjusting my PR to PR #580 once it is merged. What do you think?

fmaussion:
@anoukvlug regarding https://github.com/OGGM/oggm/pull/580 : as I said in this PR I think we should close it and start it again from a cleaner state, i.e. once *your* PR is merged.

I just made some edits to your code, mostly simplifications regarding which are now possible since xarray v0.11. The code is much cleaner and shorter now, and I believe we are on a very good track.

Let's see if the test still pass ;-)

anoukvlug:
@fmaussion thank you for making the changes. 

I am wondering if the failing test have something to do with the timestamp issue #157. I think the reason for not using:
`ts_tmp_avg = temp.sel(time=slice('1961', '1990'))`
before in PR #188 was related to that issue.

fmaussion:
> I am wondering if the failing test have something to do with the timestamp issue

Yes that's correct! But recent xarray updates have now finally made it possible to handle non-conventional calendars the same way as conventional ones. You'll need xarray v0.11 for this to work

fmaussion:
Thanks Anouk this is really a great step towards easy use of GCM data!

PR #583, opened 2018-10-26
Convert conftest.py to an actual plugin

TimoRoth:
This makes it required to install oggm to test it, which it pretty much is already anyway.


TimoRoth:
Failing tests are actual image mismatches, will look into them later, this PR seems to work as intended.

PR #586, opened 2018-10-29
Add a simple check for inversion velocities

fmaussion:
This is a test for a reference way to compute bulk velocities: cc @bearecinos 

Note that these are the "flowline velocities", i.e. some corrections should be made to interpret them as surface velocities


bearecinos:
Thanks ! @fmaussion I will do the same then. I don't want real surface velocities so, I think this is fine for now.

PR #587, opened 2018-10-31
Replace colormaps with python-colorspace palettes

### LINKED ISSUE #548: Bring HCL to OGGM
With the release of [python-colorspace](https://python-colorspace.readthedocs.io), we now have a way to use expertly created colormaps in OGGM. 

We should replace the existing colors (which are meh) with nice ones!

sadielbartholomew:
To close #548.

Currently a 'proof of concept' **waiting on agreement that the palette choices are the best available in context**. The ``python-colorspace`` package [provides a selection of default palettes](https://python-colorspace.readthedocs.io/en/latest/selectpalette.html) & I made a choice for the three colour sets used by OGGM (for *altitude*, *section thickness* & *glacier thickness*) based on the considerations below, however before I continue, I would like to ensure there is agreement on the choices made (which are trivial to change by editing the palette name strings in the code).

* Thickness & altitude are continuous, where (I believe, though I am not a glaciologist so correct me if I am wrong) extremes rather than central values would be of most interest, so a choice should come from the 'sequential' palettes and not the 'diverging' or 'qualitative'. 
* Colour sets used should not clash with the greyscale of the background terrain image, nor with the 'rainbow' colours of the plotted flowline. In particular, this means palettes with yellow should be chosen over those with white for one side of the scale.
* Altitude colouring should subtle, as it is subsidary information, but the thickness colour sets should stand out as much as possible.
* The two thickness colours should easy to distinguish from each other. In fact, ideally the three palettes would be distinct to allow for immediate comprehension of the depicted data.
* Glaciers physically are 'cool blue' in colour, so it would be nice in one case to use a palette close to the reality!

Therefore, I went for **"Greens 2"** for altitude (like grass, with white 'snow' at higher altitudes), **"Blue-Yellow"** (glacier-like colours) for section thickness  & **"Heat"** for glacier thickness (red-based to make it clearly different to the blues of section thickness). Example images (compare to those in ``OGGM/oggm-sample-data/baseline_images/freetype_28`` i.e. [topmost](https://github.com/OGGM/oggm-sample-data/blob/master/baseline_images/freetype_28/test_chhota_shigri.png), [central](https://github.com/OGGM/oggm-sample-data/blob/master/baseline_images/freetype_28/test_downstream.png) & [last](https://github.com/OGGM/oggm-sample-data/blob/master/baseline_images/freetype_28/test_thick_alt.png)):

new_test_chhota_shigri
new_test_downstream
new_test_thick_alt

Let me know your thoughts! There is no point adapting the documented & test baseline images until you are happy with the choices.

#### To do:
- [x] Apply code to get HCL color palettes working (images correct by eyeball except for colour changes, & ``pytest oggm/tests/test_graphics.py`` passes without ``--mpl`` option).
- [x] Add new dependencies to the OGGM installation environment.
- [x] Document the change & new dependency (& contributor).
- [x] Agree palette choices **[input required]**.
- ~~Change test baseline images to account for new colour sets so ``pytest --mpl oggm/tests/test_graphics.py`` passes (i.e. update test images to the new standards).~~[with ``python-colorspace`` as an optional dependency this is no longer required]
- ~~Update documentation images to account for new colour sets.~~ [as above]

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -63,7 +63,8 @@ req_packages = ['numpy',
                 'progressbar2',
                 'boto3',
                 'requests',
-                'salem']
+                'salem',
+                'python-colorspace']
 
 
 setup(
--- a/setup.py
+++ b/setup.py
@@ -63,8 +63,7 @@ req_packages = ['numpy',
                 'progressbar2',
                 'boto3',
                 'requests',
-                'salem',
-                'python-colorspace']
+                'salem']
 
 
 setup(
--- a/.coveragerc
+++ b/.coveragerc
@@ -14,6 +14,7 @@ omit =
     */tests/*
     */mpi.py
     */__init__.py
+    */pytest_plugin.py
 
 # Regexes for lines to exclude from consideration
 exclude_lines =
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,3 @@
 oggm/tests/funcs.py ident
+deployment/docker/Dockerfile ident
 oggm/_version.py export-subst
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ services:
 
 notifications:
   email: false
+  webhooks: https://coveralls.io/webhook
 
 branches:  # blacklist
   except:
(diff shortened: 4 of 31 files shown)


fmaussion:
Thanks a lot! This is looking good ;-)

This will need some discussion indeed but I'm a bit on a rush now so I'll try to get back to you next week!

fmaussion:
Sorry it took me so long to get back to you. I hope you still have some time to spend on this! I like the new colors! Thanks a lot

My few comments:
- I like the use of global module variables to define the colormaps - especially because:
- I would prefer not to make of python-colorspace a hard dependency of OGGM (yet) - i.e. the global variables should default to the previous cmaps when python-colorspace is not installed. This would also allow for not installing it on Travis and let the graphics tests pass for now.

Now to the colors:
- I like the blue-yellow colormap for ice thickness most, and I would use it for both the glacier and section thickness. However, I feel it should be inverted (blue for thicker areas)
- I can't help it: I understand the need for HCL cmaps for everything **but** topography. See for example:

img

As opposed to the one with "Greens" above. Maybe @retostauffer wants to chime in and let us know his suggestion for colors and topography ;)

sadielbartholomew:
As well as the changes discussed above, I have upped the ``CMAP_DIVS`` integer in ``graphics.py`` to create a smoother mapping in line with the current (master branch) colormaps.

**Sample images** (same examples as in my initial comment) with the new colour set choice specifications (the altitude colormap is now back to as in the main codebase, so the central image from my initial comment is identical to the original reference linked there):

test_chhota_new1
test_thick_alt_new1

The pale yellow hue for small section thickness isn't the most distinct against the white/grey of the background terrain image, but as long as this does not concern you I think these palette choices work well. If you prefer, however, I could crop the lower end of the color palette so that the scale zero is mapped to a slightly greener tinge?

setup.py
```diff
@@ -63,7 +63,8 @@
                 'progressbar2',
                 'boto3',
                 'requests',
-                'salem']
+                'salem',
+                'python-colorspace']
```
fmaussion: Can you also remove it from setup.py?

sadielbartholomew: I'm just doing that now & about to squash it in (sorry, I didn't think anyone would be online at the present time).

sadielbartholomew:
Glad to help @fmaussion. Thanks for the rapid feedback.

sadielbartholomew:
(One Travis CI job fails but it is unrelated I believe. Code coverage decreases a little, naturally, but I think it would be overkill at this stage, i.e. for this PR, to create image comparison tests.)

TimoRoth:
Docker Hub seems to be having issues the last couple days. Builds are randomly failing left and right on various services because of it.

fmaussion:
Thanks!

PR #591, opened 2018-11-09
Change order in which MultipleFlowlineMassBalance reads flowlines

matthiasdusch:
If a `MultipleFlowlineMassBalance` is initialized without explicit Flowline objects, it tries to read the `model_flowlines` and `inversion_flowlines` from the GlacierDirectory. 
But till now, if both files were available, it used `inversion_flowlines` instead of the desired `model_flowlines`.
This is now changed.

This PR also applies the same logic used in the Massbalance-Model to `graphics.plot_centerlines`: If called with `use_flowlines=True`, it uses `model_flowlines` if they exists and falls back to `inversion_flowlines` if not.

### DIFF
--- a/oggm/core/massbalance.py
+++ b/oggm/core/massbalance.py
@@ -903,17 +903,6 @@ class MultipleFlowlineMassBalance(MassBalanceModel):
                                        '`inversion_flowlines` to continue!')\
                         from None
 
-        """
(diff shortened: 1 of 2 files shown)


matthiasdusch:
I don't like the 3 unnecessary commits.

fmaussion:
@matthiasdusch this is not important at all, we squasch them at merge!

PR #592, opened 2018-11-09
Change order in which MultipleFlowlineMassBalance reads flowlines

matthiasdusch:
If a `MultipleFlowlineMassBalance` is initialized without explicit Flowline objects, it tries to read the `model_flowlines` and `inversion_flowlines` from the GlacierDirectory. 
But till now, if both files were available, it used `inversion_flowlines` instead of the desired `model_flowlines`.
This is now changed.

This PR also applies the same logic used in the Massbalance-Model to `graphics.plot_centerlines`: If called with `use_flowlines=True`, it uses `model_flowlines` if they exists and falls back to `inversion_flowlines` if not.

### DIFF
--- a/oggm/graphics.py
+++ b/oggm/graphics.py
@@ -245,11 +245,6 @@ def plot_centerlines(gdirs, ax=None, smap=None, use_flowlines=False,
     if add_downstream and not use_flowlines:
         raise ValueError('Downstream lines can be plotted with flowlines only')
 
-    # Files
-    filename = 'centerlines'
-    if use_flowlines:
-        filename = 'inversion_flowlines'
-
     gdir = gdirs[0]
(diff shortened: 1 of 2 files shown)


oggm/graphics.py
```diff
@@ -272,7 +267,18 @@ def plot_centerlines(gdirs, ax=None, smap=None, use_flowlines=False,
                               color='black', linewidth=0.5)
 
         # plot Centerlines
-        cls = gdir.read_pickle(filename)
+        if use_flowlines:
+            try:
+                cls = gdir.read_pickle('model_flowlines')
```
fmaussion: Can you solve this with a keyword argument instead? I.e. `use_model_flowlines`

fmaussion:
@matthiasdusch I have one small comment - can you change this in a subsequent PR?

Issue #590, opened 2018-11-06
Idea: start to freeze package versions on oggm-deps


fmaussion:
@TimoRoth I would like to discuss a possibility here. It's just an idea, but the more I think of it the more sense it makes: I think we should start to freeze (e.g: pin) the oggm-deps (and, per analogy, the docker container and cluster) to a set of package versions that works. 

This freeze wouldn't be a "hard one", meaning that OGGM will still work with other package versions. (However, it might issue a warning or similar when not, although I'm not sure this is necessary).

We will then update the package versions when deemed necessary, i.e. for major OGGM tags or when required by a new feature.

The advantages are:
- avoid the install chaos we have right now
- less work for you
- more reproducibility - we ensure that everybody works with the same package versions

The disadvantage is that we might oversee important updates, but this can be avoided if we test on two streams: "frozen" and "latest".

What do you think?

fmaussion:
@OGGM/oggm-devs if you want to chime in, please do ;-)

TimoRoth:
I can try setting it up, but it seems like a massive hassle. From my experience, hard-pinnen dependencies cause more headaches down the road when trying to combine versions. Specially with gdal, which depends on the library version present on the system.

fmaussion:
> Specially with gdal, which depends on the library version present on the system.

True. The alternative is to use conda binaries (or wheels) everywhere, in order not to rely on system versions at all.

fmaussion:
Other alternative: we use two frozen versions:
- cluster version
- conda version

and we make sure that the results are in-line together. Correct me if I'm wrong, but pinning conda's oggm-deps would be of a great help right?

TimoRoth:
I would kind of expect a full freeze of the deps on the oggm-deps package to make it break even more often, as not all deps have fully frozen deps on their own, and over time this grows into a mess.
The only way to avoid this would be to put absolutely everything in there, which would be over 100 packages in total.

fmaussion:
> I would kind of expect a full freeze of the deps on the oggm-deps package to make it break even more often, as not all deps have fully frozen deps on their own, and over time this grows into a mess

I don't understand this. If we have a list of dependencies defined as such:

```
channels:
  - conda-forge
dependencies:
  - python=3.6
  - rasterio=1.1
  - numpy=1.3
  - ...
```

With all packages pinned to a version, then this setup will work for as long as the packages are available in this version in conda-forge channel. Which is, I hope, going to be a long time.

[...]

PR #594, opened 2018-11-11
Add a roadmap for our policy regarding dependencies

fmaussion:
Closes https://github.com/OGGM/oggm/issues/590 

@TimoRoth @alexjarosch this a draft roadmap on how we could handle the dependency problem.

@TimoRoth of course this is a draft and your input is welcome. My objective here is to reduce both your and my maintenance work, and I believe the guidelines here will help a litte.

We can discuss here or on slack what needs to be done to finish this, but as I understand it we are pretty close already.

### DIFF
--- a/DEPENDENCIES_ROADMAP.md
+++ b/DEPENDENCIES_ROADMAP.md
@@ -1,0 +1,182 @@
+# Package dependencies and reproducibility in OGGM: a roadmap
+
+
+OGGM relies on a large number of external python packages (dependencies).
+Many of them have complex dependencies themselves, often compiled binaries
+(for example `rasterio`, which relies on a C package: `GDAL`).
+
+The complexity of this dependency tree as well as the permanent updates of 
+both OGGM and its dependencies has lead to several unfortunate situations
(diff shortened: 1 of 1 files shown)


fmaussion:
You can see it rendered [here](https://github.com/OGGM/oggm/blob/919bb2b3ac1f6fb193aaa4bac7189872124167af/DEPENDENCIES_ROADMAP.md)

alexjarosch:
Hi guys,
for your situation 5 as well as for general reproducibility as a scientific standard, it is vital to publish a container along with some results from OGGM so that everybody can reproduce it. This is possible as you can nicely store the respective container on github (LFS extension) and thus it will end up in your zenodo doi numbered repo snapshot which will be in your paper.

I think we should establish this as "good practice" when working with OGGM and require that from our users when they publish with OGGM.

Alex

fmaussion:
Thanks to the great work by Timo, I think we are following this roadmap quite closely now. I propose to merge this, and keep it available for future changes.

fmaussion:
Regarding "good practices" I'm working on the documentation right now and will submit a PR shortly

PR #596, opened 2018-11-12
Remove implicit fallback to inversion_flowlines in MultipleFlowlineMassBalance

matthiasdusch:
If a MultipleFlowlineMassBalance-model was initialised without explicitly passing a list of flowlines, `model_flowlines` were loaded. If they were not available the model implicitly used `inversion_flowlines` as fallback.

As discussed with @fmaussion this is unpythonic and a possible source of unintentional model behavior which would be hard to trace back.

Reading the `model_flowlines` at initialisation if no flowlines are passed is wanted for convenience and fits into the general model handling.
This commit introduces a new keyword `use_inversion_flowlines` (default==False). When set to True the `inversion_flowlines` will be read from the GlacierDirectory.

### DIFF
--- a/oggm/core/massbalance.py
+++ b/oggm/core/massbalance.py
@@ -871,7 +871,8 @@ class MultipleFlowlineMassBalance(MassBalanceModel):
     """
 
     def __init__(self, gdir, fls=None, mu_star=None,
-                 mb_model_class=PastMassBalance, **kwargs):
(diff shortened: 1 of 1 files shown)


fmaussion:
Looks good, but this brake some tests -  let's see ;-)

fmaussion:
Ha! It didn't - na dann...

PR #597, opened 2018-11-12
Test in multiple images in parallel, only depending on the stable one to pass

TimoRoth:
Won't work right away, as that docker tag is still building. Expect it to be around in 2-4 hours.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,105 +19,47 @@ cache:
 matrix:
   fast_finish: true
   allow_failures:
-    - os: windows
-    - os: osx
+    - env: TEST_CONTAINER=py37 OGGM_TEST_ENV=prepro MPL=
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -19,105 +19,57 @@ cache:
 matrix:
   fast_finish: true
   allow_failures:
-    - os: windows
-    - os: osx
+    - env: TEST_CONTAINER=latest
+    - env: TEST_CONTAINER=py37
```
fmaussion: I think you need to list all envs here, e.g.:

```
    - env: TEST_CONTAINER=latest OGGM_TEST_ENV=utils MPL=
      os: linux
```

TimoRoth: What do you think about removing the latest tests instead?

fmaussion: > What do you think about removing the latest tests instead?

Yes! Good idea.

PR #599, opened 2018-11-13
Try fixing coveralls

TimoRoth:


fmaussion:
Good news! Coveralls is communicating with us again (I think this is due to my log out - log in action)

Bad news! We are still at 52% :D

TimoRoth:
I think this is fixed now, just testing around a bit now, will merge when done with that.

TimoRoth:
I _think_ the 76% it reports now are accurate, but I'm not sure how to exactly read the reports it gives. There are lines in red, green, and some marked with uncov.

fmaussion:
Thanks Timo, 76% does seem much more reasonable.

The reports also look better, although some of the red lines are strange (e.g. function definition lines or module level imports are also red), but at least this is something we can work with now.

PR #600, opened 2018-11-13
WIP: Allow individual glaciers to merge together

matthiasdusch:
- [ ] Tests added/passed
- [ ] Fully documented
- [ ] Write a new GlacierDirectory for merged glaciers
- [ ] Come up with a smarter solution how to efficiently handle the climate files

OGGM glaciers are initialised with current (~2003) glacier geometry and modeled as entities. For all scenarios with glacier retreat or only small advances this is perfectly fine. But for simulations where glaciers advance significantly nearby glaciers might merge together and behave different then when modeled individually.
The first requirement to tackle this issue was solved with #539, as each Flowline got an individual mu*.

This Pull Request will most likely only deal with individual glaciers and their tributary glaciers. Future work could adapt this approach and apply it to regional scales. But I doubt merging glaciers will ever become a fully automated process within all OGGM simulations. More likely merging glaciers will be a deliberate choice for specific simulations.

In this 1. commit I provide two new functions:
`intersect_downstream_lines` finds tributary glaciers to a main glacier by intersecting downstream lines. This will separate true tributaries from ones flowing into other valleys.
`merge_tributary_flowlines` will merge the model_flowlines of the tributary glaciers with the ones of the main glacier. The result is one glacier containing all tributary flowlines.

One necessary addition in the process was to add a `Centerline.climatefile` attribute. This allows to use a different climate file for each individual flowline, as tributary glaciers might likely use a different climate reference file than the main glacier one.

This PR is still work in progress. I at least want to add a more automatic selection of possible tributary glaciers. And tests are missing as well of course. But I am happy to take comments or suggestions!

-----
For illustration a simulation with artificial climate around Hintereisferner: For the first run all 5 pictured glaciers (Hintereisferner, two small previously connected pieces, Kesselwandferner in the north and Hochjochferner in the east) are modeled in basic OGGM fashion as entities.

The second plot shows a merged run, where all flowlines of the other 4 glaciers are merged to Hintereisferner.
Note how the upper part of the glaciers are identical in depth and width, but the lower part of the merged glacier is thicker, wider and further advanced as there is now more mass contributing to it.

individual_hef

merged_hef


matthiasdusch:
Thanks for the comments. I updated the docstrings and added some goals to this PR.

fmaussion:
@matthiasdusch FYI I'd really like to have this in before the v1.1 release - even as a prototype. Do you think it's manageable? 

I'm very keen on this because of Hintereisferner mostly, which looks MUCH better with proper divides ;-)

PR #605, opened 2018-11-24
Avoid using pytest-cov plugin to escape not measuring early imports.

TimoRoth:


fmaussion:
Looks good! Thanks!!!!

PR #606, opened 2018-11-24
Move process_gcm_data to gcm_climate

fmaussion:
After some thought, it feels more right to have all things "GCM related" in the same module. With this logic:
- `climate.py` -> all things relative to the baseline climate (e.g. CRU, HISTALP, and maybe soon Reanalysis)
- `climate_prepro` -> all things relative non baseline climate

I'm actually looking for a better name than "climate_prepro", will think about it ;-)

cc @anoukvlug


anoukvlug:
@fabien maybe we can change the file name to `gcm_climate`.

fmaussion:
> maybe we can change the file name to gcm_climate.

I like it! Renamed

PR #608, opened 2018-11-29
Add rgi_id to flowlines and pass them to the massbalance models

matthiasdusch:
This commit ads a attribute `rgi_id` to Centerlines and Flowlines in order to associate them with their glacier.

The intention comes from merged glaciers where flowlines from several glaciers are merged to one new glaicer. See #600 for more details on that.

Here I add the `rgi_id` to the `input_filesuffix` of the massbalance model. This allows to load a climat file from a different glacier if flowlines of several glaciers are merged to one big glacier.

The `rgi_id` is added to the front of the filesuffix. A climatefile could now for example have the structure: `climate_monthly_RGI60-01.23456_suffix.nc`

`Centerline.rgi_id` is now set by default when initialized with `centerlines.initialize_flowlines`.

### DIFF
--- a/oggm/core/centerlines.py
+++ b/oggm/core/centerlines.py
@@ -87,6 +87,7 @@ class Centerline(object):
         self.mu_star_is_valid = False  # if mu* leeds to good flux, keep it
         self.flux = None  # Flux (kg m-2)
         self.flux_needs_correction = False  # whether this branch was baaad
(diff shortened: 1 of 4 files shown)


matthiasdusch:
I updated some changes after discussion with @fmaussion and updated the initial PR text. Parts of the first commits have become obsolet and were undone with the latest commit.

PR #611, opened 2018-11-30
Allow traceback when using executing_entity_task

fmaussion:
For debugging, raising this way was hidding the traceback and it was not possible to trace  the origin of the error.

With this change, the traceback is now correctly showing when multiprocessing is set to False 

(multiproc = True is not good for debugging, maybe I should add this to the docs somewhere)

cc @TimoRoth - btw, do we still need this pickle_copier now that we are a python3 shop only?

### DIFF
--- a/oggm/workflow.py
+++ b/oggm/workflow.py
@@ -81,22 +81,12 @@ class _pickle_copier(object):
             call_func = self.call_func
         else:
             call_func, gdir = arg
-        try:
-            if isinstance(gdir, Sequence):
-                gdir, gdir_kwargs = gdir
-                gdir_kwargs = _merge_dicts(self.out_kwargs, gdir_kwargs)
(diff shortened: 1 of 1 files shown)


TimoRoth:
Yeah, we can't really get rid of it, as it's doing a lot more things that are not really replaceable by standard python functions without essentially writing the same thing again.

TimoRoth:
Also, didn't we originally introduce this handler to prevent a single failing glacier from crashing the entire run?

fmaussion:
Thanks for the feedback!

> Also, didn't we originally introduce this handler to prevent a single failing glacier from crashing the entire run?

Maybe in the beginning, I don't remember, Now this logic is in the `entity_task` decorator:

https://github.com/OGGM/oggm/blob/586ae0f1ae44e436714444aeb9cd81ec5fa407ad/oggm/utils.py#L2712-L2726

PR #624, opened 2018-12-10
Allow individual glaciers to merge together

matthiasdusch:
**This is the same as #600.** I made some awful merge/commit errors there. Opening a new PR seemed cleaner and easier.

- [x] Tests added/passed
- [x] Fully documented
- [x] Write a new GlacierDirectory for merged glaciers
- [x] Come up with a smarter solution how to (efficiently) handle the climate files
- [x] Write a function `gis.merged_glacier_masks` to provide necessary geometry files for graphics
- [x] Write a workflow-function to easily process a set of glaciers and tributaries

OGGM glaciers are initialised with current (~2003) glacier geometry and modeled as entities. For all scenarios with glacier retreat or only small advances this is perfectly fine. But for simulations where glaciers advance significantly nearby glaciers might merge together and behave different then when modeled individually.
The first requirement to tackle this issue was solved with #539, as each Flowline got an individual mu*.

This Pull Request will most likely only deal with individual glaciers and their tributary glaciers. Future work could adapt this approach and apply it to regional scales. But I doubt merging glaciers will ever become a fully automated process within all OGGM simulations. More likely merging glaciers will be a deliberate choice for specific simulations.

In this 1. commit I provide two new functions:
`intersect_downstream_lines` finds tributary glaciers to a main glacier by intersecting downstream lines. This will separate true tributaries from ones flowing into other valleys.
`merge_tributary_flowlines` will merge the model_flowlines of the tributary glaciers with the ones of the main glacier. The result is one glacier containing all tributary flowlines.

One necessary addition in the process was to add a `Centerline.climatefile` attribute. This allows to use a different climate file for each individual flowline, as tributary glaciers might likely use a different climate reference file than the main glacier one.

This PR is still work in progress. I at least want to add a more automatic selection of possible tributary glaciers. But I am happy to take comments or suggestions!

-----
For illustration a simulation with artificial climate around Hintereisferner: For the first run all 5 pictured glaciers (Hintereisferner, two small previously connected pieces, Kesselwandferner in the north and Hochjochferner in the east) are modeled in basic OGGM fashion as entities.

The second plot shows a merged run, where all flowlines of the other 4 glaciers are merged to Hintereisferner.
Note how the upper part of the glaciers are identical in depth and width, but the lower part of the merged glacier is thicker, wider and further advanced as there is now more mass contributing to it.

individual_hef

merged_hef

----
I think the last two things to do are
- write a small function `gis.merged_glacier_masks`, adapted from the original glacier_masks which provides glacier outlines to the graphics functions. For this I need to merge the outlines of the tributary glaciers and combine them in one shp-file.
- write a workflow-function which takes a list of glaciers and a list of possible tributary glaciers and returns the merged glaciers


oggm/core/flowline.py
```diff
@@ -1804,3 +1807,133 @@ def run_from_climate_data(gdir, ys=None, ye=None,
...
+            # 5. set grid size attributes
+            dx = [shpg.Point(tfl.line.coords[i]).distance(
+                shpg.Point(tfl.line.coords[i+1]))
+                for i, pt in enumerate(tfl.line.coords[:-1])]  # get distance
+            # and check if equally spaced
+            if not np.allclose(dx, np.mean(dx), atol=1e-2):
+                raise RuntimeError('Flowline is not evenly spaced.')
+            tfl.dx = np.mean(dx).round(2)
+            tfl.map_dx = mfl.map_dx
+            tfl.dx_meter = tfl.map_dx * tfl.dx
```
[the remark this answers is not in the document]
matthiasdusch: Added issue #638 and will do that later.

oggm/utils/_workflow.py
```diff
@@ -1819,3 +1819,127 @@ def copy_to_basedir(gdir, base_dir, setup='run'):
...
+    # write gridded data and geometries for visualization
+    merged_glacier_masks(merged, merged_geometry)
+
+    # reset dx method
+    cfg.PARAMS['grid_dx_method'] = 'square'
+
+    # copy main climate file and climate info to new gdir
+    climfilename = filename + '_' + main.rgi_id + input_filesuffix + '.nc'
+    climfile = os.path.join(merged.dir, climfilename)
+    shutil.copyfile(main.get_filepath(filename, filesuffix=input_filesuffix),
```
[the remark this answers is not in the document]
matthiasdusch: No, here (`initialize_merged_gdir`) I create the new GlacierDirectory and copy the files of the 'old' glacier (=the glacier of interest). In `merge_tributary_flowlines` I copy the files of the tributary glaciers to the new GDir.

oggm/utils/_workflow.py
```diff
@@ -1819,3 +1819,127 @@ def copy_to_basedir(gdir, base_dir, setup='run'):
...
+        # remove inflow points, will be set by other flowlines if need be
+        fl.inflow_points = []
+
+        # 5. set grid size attributes
+        dx = [shpg.Point(fl.line.coords[i]).distance(
+            shpg.Point(fl.line.coords[i+1]))
+            for i, pt in enumerate(fl.line.coords[:-1])]  # get distance
+        # and check if equally spaced
+        if not np.allclose(dx, np.mean(dx), atol=1e-2):
+            raise RuntimeError('Flowline is not evenly spaced.')
```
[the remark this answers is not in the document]
matthiasdusch: see above

oggm/core/flowline.py
```diff
@@ -1804,3 +1807,133 @@ def run_from_climate_data(gdir, ys=None, ye=None,
...
+
+            if nr == 0:
+                # change the array size of tributary main flowline attributs
```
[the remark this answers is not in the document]
matthiasdusch: Well I agree mutilating the existing Centerline is not very pretty.
But to initialize a new one, I think we would either have to use some manipulated glacier geometries or copy, adapt and redo the whole centerline processes. Both seems more error prone to me than cutting the line where we know it was initialized for a working glacier.

But I am happy to discuss and change this later on.

oggm/tests/test_models.py
```diff
@@ -2658,3 +2658,97 @@ def test_elevation_feedback(self):
...
+        workflow.gis_prepro_tasks(gdirs)
+        workflow.climate_tasks(gdirs)
+        workflow.inversion_tasks(gdirs)
+        workflow.execute_entity_task(tasks.init_present_time_glacier, gdirs)
+
+        # run parameters
+        years = 200  # arbitrary
+        y0 = 1950  # arbitrary
+        tbias = -1.0  # arbitrary
+        mbbias = 0.0  # necessary?
```
fmaussion: necessary only if you want the MB at t* , i.e. a balanced MB

oggm/workflow.py
```diff
@@ -369,3 +371,78 @@ def inversion_tasks(gdirs):
...
+    main_rgi_ids = utils.tolist(main_rgi_ids)
+
+    # split main glaciers from candidates
+    maindf = glcdf.loc[glcdf.RGIId.isin(main_rgi_ids)].copy()
+    gdirs_main = [gd for gd in gdirs_all if gd.rgi_id in main_rgi_ids]
+
+    # preprocess the main glaciers to the end
+    climate_tasks(gdirs_main)
+    inversion_tasks(gdirs_main)
+    execute_entity_task(tasks.init_present_time_glacier, gdirs_main)
```
fmaussion: Intuitively I wouldn't (re)do all the preprocessing tasks here. I thought the workflow would be: 1 prepro some glaciers, 2 merge them, 3 run the merge simulation. This task should start at step 2 and assume the glaciers are preprocessed already.

(note that it's also because of my recent developments which allow to start from a pre-processed state, which will become the default workflow for most users)

oggm/workflow.py
```diff
@@ -369,3 +371,78 @@ def inversion_tasks(gdirs):
...
+
+    # check if all tributaries are only used once
+    rgiids = [gd.rgi_id for gd in gdirs_tribs]
+    if not len(np.unique(rgiids)) == len(rgiids):
+        raise RuntimeError('Every tributary glacier should only be used once!')
+
+    # preprocess the tributaries to the end
+    climate_tasks(gdirs_tribs)
+    inversion_tasks(gdirs_tribs)
+    execute_entity_task(tasks.init_present_time_glacier, gdirs_tribs)
```
fmaussion: I guess you are doing this to preprocess only the glaciers you really need. I can see why this is useful, but it feels quite redundant.

oggm/tests/test_models.py
matthiasdusch: When I wrote this, I wasn't sure how to handle the MB calibration bias and have since then forgotten about it....
I now implemented the `local_mustar` files to be copied to the merged GDir. MultipleFlowlineMB can now use a different bias for each flowline if they originate from different glaciers.

oggm/workflow.py
matthiasdusch: Yes, my initial idea was to only preprocess the none relevant glacieres as far as neccessary. Purely with performance in mind.

But I agree, this bloats the workflow function. I changed this. For the unlikely case one runs into performance problems because of this, it's easier to write a dedicated workflow.

PR #626, opened 2018-12-13
Add support for ArcticDEM and TanDEM-X

fmaussion:
TanDEM-X is downloaded via system calls to scp, which is far from optimal but a compromise for now.

(it won't work on the cluster, will need some more tweaks if we start to use it in production)

- [ ] Tests added/passed
- [ ] Fully documented

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -30,7 +30,7 @@ Get in touch
 - Follow us on `Twitter`_.
 - Or you can always send us an `e-mail`_ the good old way.
 
-.. _e-mail: info@oggm.org
+.. _e-mail: https://mailman.zfn.uni-bremen.de/cgi-bin/mailman/listinfo/oggm-users
 .. _on GitHub: https://github.com/OGGM/oggm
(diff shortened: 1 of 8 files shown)


TimoRoth:
The more correct way would be to reference the original http or ftp URL of the files, which do require user login, but will be found in the cache on the cluster. Could possibly read the login data from an env-var or something.

fmaussion:
The relevant part is here:

https://github.com/OGGM/oggm/blob/093a1d6819e40037391d12ed2cba3b2f3cf4a5b1/oggm/utils/_downloads.py#L448-L451

Which should be replaced with:

https://github.com/OGGM/oggm/blob/093a1d6819e40037391d12ed2cba3b2f3cf4a5b1/oggm/utils/_downloads.py#L407

Probably some new keywords regarding passwords will be needed

fmaussion:
In addition to an environment variable I would be quite keen to have the logins in a config file as well, for example `.oggm_config`?

PR #642, opened 2018-12-28
Switch off multiproc in merge glacier tests

fmaussion:
I have the strong intuition that multiprocessing is the reason why code coverage is missing.

cc @matthiasdusch


fmaussion:
One line of code, +2.4% in code coverage haha. See https://github.com/OGGM/oggm/issues/643 for a follow-up

PR #645, opened 2018-12-29
Do not download oggm files at import

fmaussion:
It lead to unwanted downloads in the working directory on the cluster and clashes between nodes


fmaussion:
OK this seems to work but this still feels like its not the best solution. Leaving it as is now.

PR #659, opened 2019-01-06
Try to make the extract operation more robust

fmaussion:
This is to avoid random test fails like this one: https://travis-ci.org/OGGM/oggm/jobs/476007883

### DIFF
--- a/oggm/utils/_workflow.py
+++ b/oggm/utils/_workflow.py
@@ -6,6 +6,7 @@ import os
 import tempfile
 import gzip
 import json
+import time
 import shutil
 import tarfile
 import sys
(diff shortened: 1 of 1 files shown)


TimoRoth:
I'm not sure if deleting the target dir is such a good idea. One 
configuration or code mistake, and it goes on wiping your home or workdir.

fmaussion:
> I'm not sure if deleting the target dir is such a good idea. 

We delete dirs all the time in OGGM, since always. This part of the code is pretty robust and hidden anyway. 

-> but I understand your concern and it might be worthwhile reviewing all the occurrences of dir delete in OGGM in a separate PR

fmaussion:
In this case I'm not even sure why this is needed in the first place, the extractall operation really randomly failed with a fileexists error from time to time (also on cluster)

PR #662, opened 2019-01-07
Volume Area Scaling (VAS) model from Marzeion et.al., 2012

oberrauch:
Adding the first version of the original mass balance model and volume area scaling model used by Marzeion et.al., 2012, including an altered calibration method. The `test_vascaling.py` file contains the corresponding unit tests.

- [x] Tests added/passed.
- [x] Fully documented.
- [x] Entry in `whats-new.rst`: todo


oggm/utils.py
```diff
@@ -1080,13 +1080,18 @@ def rmsd(ref, data, axis=None):
...
+    return rmsd(ref - np.mean(ref), data - np.mean(data))
+
+
 def rel_err(ref, data):
     """Relative error. Ref should be non-zero"""
     return (np.asarray(data) - ref) / ref
 
 
 def corrcoef(ref, data):
-    """Peason correlation coefficient."""
```
oberrauch: Typo in *Pearson* correlation coefficient.

fmaussion:
Hi @oberrauch , I had an overview look: this is awesome! Congrats on the great coding and the tests, this all looks good. Let's aim to get this merged this week. 

Can you do a couple of things before I have a look again?

- Check all the docstrings to follow the [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html)
- I agree to the suggestion to change the pre-computed reference lists. Add the `oggm_` prefix to the existing keys so that you can name your files in `oggm-sample-data` with a `vas_` prefix. Then, you should be able to use the existing code in `cfg.py` with a simple loop: ```with prefix in ['oggm', 'vas'] do:```
- add the `@entity_task` decorator to your VAS tasks (but just the tasks, so I guess: `local_t_star`, `t_star_from_refmb`, `find_start_area`, `run_random_vas_climate`. This is necessary for multiprocessing and error handling (among other things).
- we have to think about the name of the tasks. Currently, they have the same name as the OGGM defaults pendants: it makes sense from a python module perspective, but ideally we would have the VAS tasks listed in the `oggm.tasks` module alongside the others. The reason of the `oggm.tasks` module is simply to allow users to run OGGM without having to import or touch any of the `core` modules. So, for full integration of VAS we need to add a `vas_` prefix to all your tasks... This is going to be redundant with your module name but I don't see any other way right now.
- once this is done, add your tasks to `oggm.tasks` as well.

When this is done I will have another look, but I'd like to merge this ASAP and re-iterate later when we start using it "in production" and when we will write a doc page for it ;-)

Good job!!!

oberrauch:
Hey @fmaussion, thanks for the comments and suggestions.
I'm working on it and will let you know once I'm done.
Cheers

oberrauch:
> * we have to think about the name of the tasks. Currently, they have the same name as the OGGM defaults pendants: it makes sense from a python module perspective, but ideally we would have the VAS tasks listed in the `oggm.tasks` module alongside the others. The reason of the `oggm.tasks` module is simply to allow users to run OGGM without having to import or touch any of the `core` modules. So, for full integration of VAS we need to add a `vas_` prefix to all your tasks... This is going to be redundant with your module name but I don't see any other way right now.
> * once this is done, add your tasks to `oggm.tasks` as well.

Hi @fmaussion, I have an idea... Do you think it would work if we import the tasks in the `tasks.py` file with alias that have the `vas_` prefix?!

`from oggm.core.vascaling import local_t_star as vas_local_t_star`
`from oggm.core.vascaling import compute_ref_t_stars as vas_compute_ref_t_stars`
`...`

fmaussion:
> Do you think it would work if we import the tasks in the tasks.py file with alias that have the vas_ prefix?!

Yes I believe it would work - I like the idea, but there are small cons as well (name conflicts possible at he package level with two funcs having the same name), but I don't mind too much

oberrauch:
> but I don't mind too much

Ok, nice. And if problems arise, we can change it later anyways...

oberrauch:
Hi @fmaussion. I think now we are ready to merge... unless you have some more suggestions.

fmaussion:
Let's get this beast in - congrats @oberrauch , great addition! :tada:

oberrauch:
Nice. Feels good, even though there is still a lot to do...

PR #663, opened 2019-01-10
Allow fallbacks for failing entity tasks + fallback functions for 'local_t_star' and 'mu_star_calibration'

matthiasdusch:
- [ ] Tests added/passed
- [ ] Fully documented

This PR expands the `execute_entity_task` decorator with a fallback argument:
If a task fails and `cfg.PARAMS['continue_on_error']==True` a fallback-function will be executed.

I suggest storing these fallback functions in `utils/_fallbacks.py`.

Hera also two first fallback functions for climate.local_t_star and climate.mu_star_calibration are included.
`local_t_star` writes a `local_mustar.json`  to the glacier directory with information on tstar, mustar, and bias. If this task fails but `cfg.PARAMS['continue_on_error']==True` later tasks like 'mu_star_calibration' would read a possible preexisting json file (e.g. if the same glacier is run several times with different parameter settings).

### DIFF
--- a/oggm/core/climate.py
+++ b/oggm/core/climate.py
@@ -977,7 +977,7 @@ def calving_mb(gdir):
     return gdir.inversion_calving_rate * 1e9 * rho / gdir.rgi_area_m2
 
 
-@entity_task(log, writes=['local_mustar'])
+@entity_task(log, writes=['local_mustar'], fallback=utils.fb_local_t_star)
 def local_t_star(gdir, *, ref_df=None, tstar=None, bias=None):
     """Compute the local t* and associated glacier-wide mu*.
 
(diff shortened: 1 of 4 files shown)


fmaussion:
Looking good, thanks! I think it should be relatively easy to add a test for this, by setting unrealistic mu_start boundaries and let the run fail, then check that the files are nans. Could you add that?

matthiasdusch:
After discussion with @fmaussion I removed `utils/_fallbacks.py` and added `fallback_local_t_star` directly to the climate module. It's a specific fallback function and not a general one.

I also got rid of `fallback_mu_star_calibration`: It is not really necessary as subsequent tasks will fail anyway, not matter if e.g. `local_mustar['mu_star_allsame']` is NaN or does not exist et all. The important part is, that `fallback_local_t_star` writes and NaN-filled file to the GDir to avoid reuse of a previous run.

PR #669, opened 2019-01-15
add k as a parameter in the cfg file

bearecinos:
I need to be able to change the value of k when calling the function: `calving_flux_from_depth(gdir, k=None, water_depth=None)`

For me this works let me know if there is any issues with adding this to OGGM


fmaussion:
Perfect, thanks a lot!

PR #670, opened 2019-01-15
Verify size and crc32 of downloaded files

TimoRoth:
This bundles a list of all files OGGM might currently download, listing the size and sha256 checksum of each file.

The list originates from the Cluster in Bremen, which should currently hold absolutely all files OGGM might ever download.

Not fully tested yet, PR for tests and review.


oggm/utils/_downloads.py
```diff
@@ -199,6 +230,42 @@ def _cached_download_helper(cache_obj_name, dl_func, reset=False):
...
+                sha256.update(b)
+        size = os.path.getsize(path)
+
+        if cache_obj_name not in data:
+            logger.warning('No known hash for %s: %s %s' %
+                           (path, size, sha256.hexdigest()))
+        else:
+            data = data[cache_obj_name]
+            if data[0] != size or data[1] != sha256.digest():
+                err = '%s failed to verify!\nis: %s %s\nexpected: %s %s' %
```
matthiasdusch: `err = '%s failed to verify!\nis: %s %s\nexpected: %s %s' % (`
 ` ` ` ` ` ` ` ` `path, size, sha256.hexdigest(), data[0], data[1].hex())`

should work

TimoRoth:
Judging from the coveralls report, it seems to never successfully verify anything, but always does not find the file in question in the list.
Can't see any logs of those incidents anywhere on Travis though.

matthiasdusch:
I do like the possibility to verify the downloaded files.

What seems a bit bulky to me is this one massive file with the hashes. It's bigger than the actual module and has to be recreated and redistributed every time one file changes or a new file is added.

Wouldn't it be easier to have individual hash-files at some location, e.g. the cluster in Bremen? If a user only needs a couple of DEM or CRU tiles oggm would only need to download these individual and small files. This might be a bottleneck for global simulations. But here a function to download all necessary hashes in advance might help.

Also what might be a good idea is to have one hash collection for all the oggm-sample-data.

TimoRoth:
The idea is to verify that the file is what this specific version of OGGM expects.
If the files are silently changed upstream, which has happened in the past, it would be caught immediately.

There is a major problem though. I forgot a flag when generating that file, so _a lot_ of files are missing.
There are actually 4.4 million files stored on the cluster. 4.3 Million of them are in the files that @fmaussion generated recently.
Those 4.3 million extra lines blow up the hash file size to almost 1GB, making this entire idea impossible to sensibly implement.

fmaussion:
Yeah, we need to think more deeply about that. We need to set-up a phone call for some time next week

TimoRoth:
An idea I would have would be to only store a list of (crc32(filename), crc32(filedata), uint32(filesize)), which would bring the size for all the 4.4 million files down to roughly 50MB when uncompressed(though I doubt that format can be compressed in a meaningful fashion) and should still be a decent safeguard against modification or broken downloads.

fmaussion:
yeah. And, we reduce the number of files I generated. It's overkill anyway.

 crc32(filedata) or uint32(filesize) might even be enough anyway. We just want to make sure that all the data is downloaded properly

fmaussion:
OK, so how about an inelegant (but pragmatic) solution...

First, we could just store uint32(filesize) in some binary format (maybe netCDF): the idea is simply to make sure that we downloaded the file correctly, not to protect against malicious files or whatever. This would already make the file quite a bit smaller.

Then, for all files except the numerous gdirs, we could have one hash file per server: e.g. dem3, SRTM, etc. This file could be stored either in oggm-sample data, or be downloaded from the cluster by OGGM only when needed. If we are very exact, we could have a hash file of the hash files ;-)

For the gdirs (millions of files), we should separate the hash files a bit more cleverly. The current structure is following:

`oggm_version/rgi_version/border_N/LX/RGI_Region...`

At the `LX` level, we have approx 200k files: I guess that this is just the limit for a hash file. So for these directories we need one hash file per `border_N/LX` combination.

All this is more involved but should do the job.

Thoughts?

TimoRoth:
This is using crc32 now, significantly reducing the size of the checksums.
Now I just need to figure out if it's actually working.

TimoRoth:
Judging by https://coveralls.io/builds/21357094/source?filename=oggm%2Futils%2F_downloads.py#L261 it is not actually working, but always unable to find the downloaded file in its list.
Though I can't seem to find the warning coverity clearly shows it's printing anywhere in the logs.
Are we somewhere disabling all logged warnings in our tests?

fmaussion:
> Judging by ...

Our tests are never downloading from cluster. When I test for downloads, I mock the download functions. 

To see if its really working you can run the download tests in test_utils.py, with `pytest --download oggm/test/test_utils.py`. But we could add one test with a small file though, I already added some for the preprocessed glacier directories


> Are we somewhere disabling all logged warnings in our tests?

No - but pytest is suppressing stdout per default, add `-s` to show it.

TimoRoth:
On 30.01.2019 17:50, Fabien Maussion wrote:
>     Judging by ...
> 
> Our tests are never downloading from cluster. When I test for downloads, 
> I mock the download functions.

It shouldn't matter from where they download. All files that are cached 
on the cluster have a checksum. Something is hitting the verification 
function, but isn't in the list.

> To see if its really working you can run the download tests in 
> test_utils.py, with |pytest --download oggm/test/test_utils.py|. But we 
> could add one test with a small file though, I already added some for 
> the preprocessed glacier directories
> 
>     Are we somewhere disabling all logged warnings in our tests?
> 
> No - but pytest is suppressing stdout per default, add |-s| to show it.

I'll try with that tomorrow.

fmaussion:
The only files which are downloaded by our tests are:
- the OGGM demo files on github
- some files in: https://cluster.klima.uni-bremen.de/~fmaussion/test_gdirs/
- some files in: https://cluster.klima.uni-bremen.de/~fmaussion/demo_gdirs/

Everything else is prohibited: see https://github.com/OGGM/oggm/blob/eedcfe0ec6defedb7eab0184e4792e2b28e1b5d1/oggm/tests/funcs.py#L213

TimoRoth:
At least the code itself is tested now and seems to work as expected.
Not sure if we need a test that actually downloads a file and verifies it.

TimoRoth:
Also, coveralls seems unmotivated to post its reports here again.
It was successfully generated though: https://coveralls.io/builds/21381193

TimoRoth:
This seems to be working for all I can tell, and is definitely a setup up in terms of reliability of downloads.
I'll merge it later today or tomorrow.

fmaussion:
> This seems to be working for all I can tell, and is definitely a setup up in terms of reliability of downloads.

I agree - thanks a lot!

> I'll merge it later today or tomorrow.

if you can wait a little more I would be thankful: I'm creating new files for the list.

Have you disabled notifications on slack? It's very hard to reach you there

TimoRoth:
Checksums file has just been freshly generated.

Also added a check for the Glacier-Dir to exist after extraction.

PR #671, opened 2019-01-15
Merged glaciers need to pass individual tstars to certain MassbalanceModels

matthiasdusch:
RandomMassBalance and ConstantMassBalance models will use `t_star` as a center of the climate period if no `y0` is passed to them.
For a merged glacier this value has to be read from the tributaries `local_mustar` files. This is now done while initializing a MultipleFlowlineMassBalance model.

### DIFF
--- a/oggm/core/massbalance.py
+++ b/oggm/core/massbalance.py
@@ -915,6 +915,7 @@ class MultipleFlowlineMassBalance(MassBalanceModel):
                                    'use_inversion_flowlines=True.') from None
 
         self.fls = fls
+        _y0 = kwargs.get('y0')
 
         # User mu*?
         if mu_star is not None:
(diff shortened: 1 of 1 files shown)


oggm/core/massbalance.py
```diff
@@ -915,6 +915,7 @@ def __init__(self, gdir, fls=None, mu_star=None,
                                    'use_inversion_flowlines=True.') from None
 
         self.fls = fls
+        _y0 = kwargs.get('y0')
```
fmaussion: `kwargs.get('y0', None)` to set a default value in case y0 isn't provided at all (otherwise this might raise a ValueError

matthiasdusch: `dict.get()` defaults to `None` if the key is not found. But yes, explicit is better than implicit ;-).

matthiasdusch:
- Implemented explicit None for kwargs.get().
- Adapted the existing merge-test: If the constant climate is run without an `y0`, and merged and singular glaciers have the same results, we are good to go.

PR #677, opened 2019-01-22
Gcm climate

nchampollion:
First try to have a function to extract GCM climate data (here from CIMP5 in order to use to GlacierMIP) in a format that OGGM can use, similar the function for CESM data.

I changed also the name of _process_gcm_data_ to _process_climate_data_ in order to be more general, and I called the two functions to extract CESM and GCM/CMIP5 data process_cesm_data and _process_gcm_data_. Leaving the possibility in the future to add functions to extract other climate data ...

- [ ] Tests added/passed
- [ ] Fully documented

### DIFF
--- a/oggm/core/gcm_climate.py
+++ b/oggm/core/gcm_climate.py
@@ -1,11 +1,15 @@
 """Climate data pre-processing"""
+
 # Built ins
 import logging
 from distutils.version import LooseVersion
+
 # External libs
 import numpy as np
 import netCDF4
+import datetime
 import xarray as xr
+
 # Locals
 from oggm import cfg
 from oggm import utils
@@ -16,14 +20,11 @@ log = logging.getLogger(__name__)
 
 
 @entity_task(log, writes=['gcm_data', 'climate_info'])
-def process_gcm_data(gdir, filesuffix='', prcp=None, temp=None,
-                     time_unit='days since 1801-01-01 00:00:00',
+def process_climate_data(gdir, filesuffix='', prcp=None, temp=None,
+                     time_unit='days since 1800-01-01 00:00:00',
                      calendar=None):
     """ Applies the anomaly method to the climate data and stores the data in a
     format that can be used by the OGGM mass balance model.
-
-    For an example implementation see
-    :ref:`process_cesm_data()`
 
     Parameters
     ----------
@@ -43,8 +44,8 @@ def process_gcm_data(gdir, filesuffix='', prcp=None, temp=None,
             time     (time) cftime object
     time_unit : str
         The unit conversion for NetCDF files. It must be adapted to the
-        length of the time series.
-        For example: 'days since 0850-01-01 00:00:00'
+        length of the time series. Default is 'days since 1800-01-01 00:00:00' (from CRU data)
+        For example: 'days since 0850-01-01 00:00:00' (CESM) or 'days since 1870-01-15 12:00:00' (GCM)
     calendar : str
         If you use an exotic calendar (e.g. 'noleap')
     """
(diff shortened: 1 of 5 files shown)


oggm/core/gcm_climate.py
```diff
@@ -16,15 +20,12 @@
 
 
 @entity_task(log, writes=['gcm_data', 'climate_info'])
-def process_gcm_data(gdir, filesuffix='', prcp=None, temp=None,
-                     time_unit='days since 1801-01-01 00:00:00',
+def process_climate_data(gdir, filesuffix='', prcp=None, temp=None,
+                     time_unit='days since 1800-01-01 00:00:00',
```
fmaussion: I think we should remove this time_unit keyword altogether - I'll have a look at what I can do

oggm/core/gcm_climate.py
```diff
@@ -58,30 +59,36 @@ def process_gcm_data(gdir, filesuffix='', prcp=None, temp=None,
...
-    temp = temp[sm-1:sm-13].load()
-
-    # compute monthly anomalies
-    # of temp
+    ##################################################################################
+    # ISSSUE: in glacier model intercomparison is it useful to do that ?    
+    # From normal years to hydrological years
+    #sm = cfg.PARAMS['hydro_month_' + gdir.hemisphere]
+    #prcp = prcp[sm-1:sm-13].load()
+    #temp = temp[sm-1:sm-13].load()
```
fmaussion: Yeah, we need to keep that. This is how OGGM works. When we output the data we can do something else, maybe we'll miss the last year or so

oggm/core/gcm_climate.py
```diff
@@ -198,13 +207,117 @@ def process_cesm_data(gdir, filesuffix='', fpath_temp=None, fpath_precc=None,
...
     # Here:
     # - time_unit='days since 0850-01-01 00:00:00'
     # - calendar='noleap'
-    process_gcm_data(gdir, filesuffix=filesuffix, prcp=prcp, temp=temp,
+    process_climate_data(gdir, filesuffix=filesuffix, prcp=prcp, temp=temp,
                      time_unit=time_units, calendar=calendar)
+
+
+@entity_task(log, writes=['gcm_data', 'climate_info'])
+def process_gcm_data(gdir, filesuffix='', fpath_temp=None, fpath_precip=None):
```
fmaussion: Why not `process_cmip_data`?

oggm/utils.py
```diff
@@ -401,7 +401,7 @@ def gettempdir(dirname='', reset=False, home=False):
 
     basedir = (os.path.join(os.path.expanduser('~'), 'tmp') if home
                else tempfile.gettempdir())
-    return mkdir(os.path.join(basedir, 'OGGM', dirname), reset=reset)
+    return mkdir(os.path.join(basedir, dirname), reset=reset)
```
fmaussion: Why changing this?

oggm/utils.py
```diff
@@ -3339,13 +3339,32 @@ def write_monthly_climate_file(self, time, prcp, temp,
...
-                numdate = netCDF4.date2num([t for t in time], time_unit,
-                                           calendar=calendar)
-            else:
-                numdate = netCDF4.date2num([t for t in time], time_unit)
+            timev.long_name = 'time'
+            timev.standard_name = 'time'
+
+
+            ##################################################################################
+            ### ISSSUE: Why converting date to numbers ?
```
fmaussion: I don't understand why you needed to change something here, will have to have a look in more detail. But in principle the baseline climate will remain CRU, GCM will never be a baseline climate

oggm/core/gcm_climate.py
nchampollion: Or it should be determined from the climate dataset we are using

oggm/core/gcm_climate.py
nchampollion: Ok, and it also makes more sense. However sometimes experiment requirements are not the same

oggm/core/gcm_climate.py
nchampollion: Yes, better. I just had the same idea right now because I was looking to CMIP6 experiments ! So let's call it `process_cmip5_data`

oggm/utils.py
nchampollion: Don't remember exactly why ... maybe due to an error I had. Or, probably it was to give more flexibility to the user to decide.

oggm/utils.py
nchampollion: Yeah, the baseline climate will remain CRU, so maybe this is not the best place to change that. This change was because the time format in CRU netcdf and CMIP5 GCM netcdf were not the same .

oggm/core/gcm_climate.py
anoukvlug: @fmaussion, @nchampollion could you let me know if you remove the `time_unit` keyword before merging the PR? So I can have a look :)

oggm/core/gcm_climate.py
```diff
@@ -198,13 +207,117 @@ def process_cesm_data(gdir, filesuffix='', fpath_temp=None, fpath_precc=None,
     assert r == 0
     ndays = np.tile(cfg.DAYS_IN_MONTH, ny)
     prcp = prcp * ndays * (60 * 60 * 24 * 1000)
-
+    print(tempds.time.values)
+    print('')
```
anoukvlug:
```
    print(tempds.time.values)
    print('')
```
Could this print statements be removed?

fmaussion:
> could you let me know if you remove the time_unit keyword before merging the PR? So I can have a look :)

@anoukvlug we are closing this PR in favor of https://github.com/OGGM/oggm/pull/679 (it's the same code but a little simpler).

I made sure @nchampollion is acknowledged in `whats-new`

PR #679, opened 2019-01-24
Add cmip5 support

fmaussion:

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -93,6 +93,7 @@ the majority of OGGM's tasks). They are parallelizable.
     tasks.process_custom_climate_data
     tasks.process_gcm_data
     tasks.process_cesm_data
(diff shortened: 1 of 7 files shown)


nchampollion:
It looks good, great ! I can test it after PR is accepted.

A shortcut to the core functions in tasks.py is missing but maybe you don't want to add right now as `from oggm.core.gcm_climate import process_cmip5_data`.

fmaussion:
> A shortcut to the core functions in tasks.py is missing but maybe you don't want to add right now as from oggm.core.gcm_climate import process_cmip5_data.

Thanks! Yes I forgot about it. 

We can test on the cluster, but for the actual runs it would be good to use the preprocessed directories once @TimoRoth has computed them

nchampollion:
Yes of course ! What is the location of the preprocessed directories ?

Issue #675, opened 2019-01-21
about: ValueError: cannot reshape array of size 1 into shape (13872,)


HoloceneCESI:
Dear Sir,

When I input the command 'python gcm_run.py cesm -e',there are some errors as shown in the attachment.
![1](https://user-images.githubusercontent.com/45832101/51483170-b7f48280-1d98-11e9-80e5-e7b181c8aaac.PNG)

anoukvlug:
I am not sure what causes the error, but I can try to help you.

To start there is a typo in:
`execute_entity_task(tasks.process_gcm_data, ......)`

> scalendar 

should be

> calendar

Could you correct the typo and check if the error stays the same?

HoloceneCESI:
> I am not sure what causes the error, but I can try to help you.
> 
> To start there is a typo in:
> `execute_entity_task(tasks.process_gcm_data, ......)`
> 
> > scalendar
> 
> should be
> 
> > calendar
> 
> Could you correct the typo and check if the error stays the same?



> I am not sure what causes the error, but I can try to help you.
> 
> To start there is a typo in:
> `execute_entity_task(tasks.process_gcm_data, ......)`
> 
> > scalendar
> 
> should be
> 
> > calendar
> 
> Could you correct the typo and check if the error stays the same?

![capture](https://user-images.githubusercontent.com/45832101/51498508-0e78b580-1dc7-11e9-9d03-0c9abf1aaa3c.PNG)

HoloceneCESI:
> > I am not sure what causes the error, but I can try to help you.
> > To start there is a typo in:
> > `execute_entity_task(tasks.process_gcm_data, ......)`
> > > scalendar
> > 
> > 
> > should be
> > > calendar
> > 
> > 
> > Could you correct the typo and check if the error stays the same?
> 
> > I am not sure what causes the error, but I can try to help you.
> > To start there is a typo in:
> > `execute_entity_task(tasks.process_gcm_data, ......)`
> > > scalendar
> > 
> > 
> > should be
> > > calendar
> > 
> > 
> > Could you correct the typo and check if the error stays the same?
> 
> ![capture](https://user-images.githubusercontent.com/45832101/51498508-0e78b580-1dc7-11e9-9d03-0c9abf1aaa3c.PNG)

Thank you,but this is not the point.....

fmaussion:
Hi, a short checklist:
1. Do you have the latest version of OGGM installed?
2. Why are you using RGI Version 5? The default should be 61
[...]

PR #681, opened 2019-01-29
Add a list of demo glaciers to OGGM

fmaussion:
This adds a curated list of demo glaciers that are faster to download and can be addressed by certain keys, like their name of acronym - will write the documentation about it in a separate PR!

### DIFF
--- a/oggm/cfg.py
+++ b/oggm/cfg.py
@@ -85,6 +85,7 @@ PARAMS = ResettingOrderedDict()
 PATHS = PathOrderedDict()
 BASENAMES = DocumentedDict()
 LRUHANDLERS = ResettingOrderedDict()
+DEMO_GLACIERS = None
 
 # Constants
 SEC_IN_YEAR = 365*24*3600
@@ -284,6 +285,7 @@ def initialize(file=None, logging_level='INFO'):
     global IS_INITIALIZED
     global PARAMS
     global PATHS
+    global DEMO_GLACIERS
 
     set_logging_config(logging_level=logging_level)
 
@@ -394,6 +396,11 @@ def initialize(file=None, logging_level='INFO'):
     # Pre extract cru cl to avoid problems by multiproc
     from oggm.utils import get_cru_cl_file
     get_cru_cl_file()
+
+    # Read in the demo glaciers
+    file = os.path.join(os.path.abspath(os.path.dirname(__file__)),
+                        'data', 'demo_glaciers.csv')
+    DEMO_GLACIERS = pd.read_csv(file, index_col=0)
 
 
 def oggm_static_paths():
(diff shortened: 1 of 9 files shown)


nchampollion:
Great idea ! 

Few comments/questions:
   1. Why Coxe glacier instead of Columbia glacier ? (which terminates into the ocean whereas Coxe not anymore if I remember well)
   2. Should we be more representative by including one glacier of HMA ? 
   3. in the ` def test_demo_glacier_id(self)` function, you test with Mer de Glace. How does the thickness inversion works in the high slope areas between Mer de Glace glacier and Glacier du Géant ?
   4. Would it be nice to include a notebook doing equilibrium or climate runs for those glaciers as well as producing nice plots ?

fmaussion:
>   Why Coxe glacier instead of Columbia glacier ? (which terminates into the ocean whereas Coxe not anymore if I remember well)

Just because Coxe is smaller and we test on Coxe. But this list is just a quick and dirty shot, we should/could add plenty of them - I'm happy to welcome a PR to make the list better. One can check for the correct IDs directly in the RGI file or on https://www.glims.org/maps/glims

>    Should we be more representative by including one glacier of HMA ?

Yeah, plus many others

 >  in the def test_demo_glacier_id(self) function, you test with Mer de Glace. How does the thickness inversion works in the high slope areas between Mer de Glace glacier and Glacier du Géant ?

I don't know!

>    Would it be nice to include a notebook doing equilibrium or climate runs for those glaciers as well as producing nice plots ?

Sure - time is our best enemy ;-) Right now I'm focussed on getting the documentation ready for these things in order to reduce the number of problem our "less tech-savvy" users have (like https://github.com/OGGM/oggm/issues/675)

nchampollion:
> > Why Coxe glacier instead of Columbia glacier ? (which terminates into the ocean whereas Coxe not anymore if I remember well)
> 
> Just because Coxe is smaller and we test on Coxe. But this list is just a quick and dirty shot, we should/could add plenty of them - I'm happy to welcome a PR to make the list better. One can check for the correct IDs directly in the RGI file or on https://www.glims.org/maps/glims

I was thinking to Columbia because Bea did lot of work on it. I will check the RGI IDs of few of them, test them and update the list.

> 
> > Should we be more representative by including one glacier of HMA ?
> 
> Yeah, plus many others
> 
> > in the def test_demo_glacier_id(self) function, you test with Mer de Glace. How does the thickness inversion works in the high slope areas between Mer de Glace glacier and Glacier du Géant ?
> 
> I don't know!

Ok, I will have a look.

> 
> > Would it be nice to include a notebook doing equilibrium or climate runs for those glaciers as well as producing nice plots ?
> 
> Sure - time is our best enemy ;-) Right now I'm focussed on getting the documentation ready for these things in order to reduce the number of problem our "less tech-savvy" users have (like #675)

Yes, of course !

PR #683, opened 2019-01-30
Fully switch to MANIFEST.in

TimoRoth:
This is actually broken on master right now:
https://cloud.docker.com/u/oggm/repository/registry-1.docker.io/oggm/oggm/hub-builddetail/bef9mpbovh93q3pptti4rs4

It only worked on Travis because it was tested from inside the oggm directory, so the paths happened to match up.
So this moves everything to MANIFEST.in, and changes the tests to test from outside of the OGGM root dir, to catch such cases in the future.


TimoRoth:
I'm not actually sure why this doubles the amount of lines coveralls sees, it goes up to 15723 from 7643.
But that generally does seem like a good thing to me. It does not appear to be testing anything twice, it just sees a lot more files for some reason.

fmaussion:
> I'm not actually sure why this doubles the amount of lines coveralls sees, it goes up to 15723 from 7643.
But that generally does seem like a good thing to me. It does not appear to be testing anything twice, it just sees a lot more files for some reason.

Still, this sounds strange. Let's see the report first

TimoRoth:
Seems like we can't actually test from outside the source dir. Too many things expect to be in there. Guess we'll just have to be careful about not relying on it too much.

fmaussion:
Yeah I'd prefer not change this - I mean, this kind of MANIFEST stuff happen very rarely

PR #689, opened 2019-02-04
Try making things more robust

fmaussion:
Again, random errors on cluster lead to these random sleeps in the code....

### DIFF
--- a/oggm/utils/_workflow.py
+++ b/oggm/utils/_workflow.py
@@ -7,6 +7,7 @@ import tempfile
 import gzip
 import json
 import time
+import random
 import shutil
 import tarfile
 import sys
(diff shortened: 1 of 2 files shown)


oggm/utils/_workflow.py
```diff
@@ -1253,9 +1254,11 @@ def _robust_extract(to_dir, *args, **kwargs):
     while count < 10:
         try:
             if count > 1:
-                time.sleep(0.1)
+                time.sleep(random.uniform(0.05, 0.1))
             with tarfile.open(*args, **kwargs) as tf:
                 tf.extractall(os.path.dirname(to_dir))
+                # Add a sleep just because
+                time.sleep(random.uniform(0.05, 0.1))
```
fmaussion: I'm not sure about this. My feeling is that there is a timing issue between the extract and the following commands, i.e. things start to compute in OGGM although the extract is not finished.

But is this even possible, @TimoRoth ? 

This sleep is very short in nature, but times 200k glaciers it's still some time...

TimoRoth: I would expect extractall to only return once it has finished unpacking, and all files should be visible on the filesystem at that point.
Can't tell how it's internally implemented, but anything else would seem like a bug to me.

fmaussion: yeah, that's what I thought. But I've really had non-predictable errors since this extract part is happening.

Non-predictable means: sometimes it works, sometimes not...

PR #693, opened 2019-02-08
Small refactoring of dl_verify parameter

fmaussion:
This moves a couple of things around for convenience and coherence.

cc @TimoRoth


TimoRoth:
No complaints from my side!

fmaussion:
> No complaints from my side! 

except it doens't work haha. I think that if `cfg.PARAMS` is not initialized, we'll just set it to true per default (I think its just for the OGGM sample data download anyway)

PR #695, opened 2019-02-11
changes in utils.calving_from_depth plus test

bearecinos:
The terminus width and depth should be given as output of the calving function. I also changed the test.


benchmarks/track_model_results.py
```diff
@@ -304,14 +304,15 @@ def setup_cache(self):
                 f_calving = 0
             elif i == 1:
                 # Second call we set a very small positive calving
-                f_calving = utils.calving_flux_from_depth(gdir, water_depth=1)
+                f_calving = utils.calving_flux_from_depth(gdir,
```
TimoRoth: Shouldn't this also be f_calving, _, _?

fmaussion: yeah this is all stupid pep8 issues ;-)

fmaussion:
Thanks @bearecinos !

PR #698, opened 2019-02-13
Add a calving test without overshoot

fmaussion:
Nothing big happening here


fmaussion:
closed in favor of https://github.com/OGGM/oggm/pull/699

PR #700, opened 2019-02-21
Bugfix for unlucky calving flux on certain glaciers

fmaussion:
cc @bearecinos


fmaussion:
I also made the convergence threshold to 1% instead, because if the one before the last is within 1%, there is a good chance that the last one is even closer

bearecinos:
Ok I will take a look ... re run things and let you know. Thanks!!!

PR #704, opened 2019-02-25
Downloading preprocessed GDirs with multiprocessing fails

matthiasdusch:
As preprocessed GDirs are regionally grouped, downloading two or more GDirs from one region will cause a `DownloadVerificationFailedException` if multiprocessing is enabled: If the file is not already downloaded, the first entity task will start downloading it, while the subsequent tasks see the existing but only partially downloaded file and will raise the above error.

This PR implements a multiprocessing Lock structure, following other download functions.

There might be smarter/faster ways to improve this in near feature.

### DIFF
--- a/oggm/utils/_downloads.py
+++ b/oggm/utils/_downloads.py
@@ -295,13 +295,6 @@ def _requests_urlretrieve(url, path, reporthook):
 
 
 def oggm_urlretrieve(url, cache_obj_name=None, reset=False, reporthook=None):
-    with _get_download_lock():
-        return _oggm_urlretrieve_unlocked(url, cache_obj_name, reset,
-                                          reporthook)
-
-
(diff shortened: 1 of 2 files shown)


fmaussion:
Thanks! 

> There might be smarter/faster ways to improve this in near feature.

yeah there is room for refactoring - this has all grown organically and is now a mess of duplicated code... :-(

fmaussion:
It seems this breaks travis quite badly - maybe at the download of the sample data already?

matthiasdusch:
Travis times out after 10min during initialization. Seems like locking `oggm_urlretrieve` and therefor downloading all files sequentially takes to much time.
To be honest I did not realize the tests so heavily depend on this function.

I'll think about a solution.

fmaussion:
Thanks a lot @matthiasdusch !!!

PR #705, opened 2019-02-25
Except DownloadVerificationFailedException in file_downloader

matthiasdusch:
This PR implements a `except` in the main `file_downloader` function in order to retry the download several times if the checksums do not match.
The file will be deleted and redownloaded.

Needs PR #704 to be merged, as unlocked downloads would fail otherwise.

### DIFF
--- a/oggm/exceptions.py
+++ b/oggm/exceptions.py
@@ -24,7 +24,9 @@ class NoInternetException(Exception):
 
 
 class DownloadVerificationFailedException(Exception):
-    pass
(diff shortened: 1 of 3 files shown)


fmaussion:
I just made a small modif to the error message and removed the `sleep` (in case the file has changed online you just want to re-download right away).

fmaussion:
I also just force-pushed to your branch, sorry ;-)

fmaussion:
Thanks Matthias, this is really looking good ;-)

PR #706, opened 2019-02-26
doc run examples

matthiasdusch:
- [x] the MB-models in the crossvalidation run examples explicitly need `use_inversion_flowlines=True`
- [x] showing the barplot is more straight forward than storing it somewhere on the users disk in my opinion

might receive more commits


fmaussion:
> might receive more commits

Ups! Sorry just read this -> I hope I wasn't too fast in merging

PR #719, opened 2019-03-19
Add HTTP auth for downloads from TanDEM-X

TimoRoth:


TimoRoth:
Still WIP, but Travis won't run on Drafts.

TimoRoth:
Passes the test and generally works fine for me now. @fmaussion 
Might still do one or two minor changes, so please don't merge just yet.

oggm/utils/_downloads.py
```diff
@@ -526,18 +532,25 @@ def _download_tandem_file_unlocked(zone):
...
+    with open(tdmauthfile, 'r') as f:
+        tdmuser = f.readline().strip()
+        tdmpass = f.readline().strip()
+    if not tdmuser or not tdmpass:
+        raise DownloadCredentialsMissingException(
+            'Could not read credentials from ' + tdmauthfile)
+
     # Did we download it yet?
-    scpfile = ('/home/data/download/tandemx-90m.dlr.de/90mdem/DEM/'
+    wwwfile = ('https://download.geoservice.dlr.de/TDM90/files/'
```
fmaussion: This change means we will have to update the paths on cluster, right?

oggm/utils/_downloads.py
```diff
@@ -526,18 +532,25 @@ def _download_tandem_file_unlocked(zone):
     if os.path.exists(outpath):
         return outpath
 
+    # Grab auth parameters
+    tdmauthfile = os.path.expanduser('~/.tdmdem90.creds')
+    if not os.path.isfile(tdmauthfile):
+        raise DownloadCredentialsMissingException(
+            tdmauthfile + ' does not exist.')
```
fmaussion: Add instructions on how to generate this file in the error message.

oggm/utils/_downloads.py
TimoRoth: No, that's where they already are.

oggm/utils/_downloads.py
TimoRoth: You think what I added is enough? I don't want to write too many instructions into an exception message, but what's there now should be enough to allow figuring things out.

PR #720, opened 2019-03-21
Changes in calving functions to fix water depth

bearecinos:
Hello, 
I have modified both calving functions so they can cope with a fix water depth from Bathymetry data.
Let me know if you are happy with the changes.

Cheers 
Bea


fmaussion:
Hi Bea, for some reason your PR deleted all empty lines, which is not good for the docstrings... Can you start from a fresh branch again?

bearecinos:
sorry I asked timo help ... he fixed very fast

fmaussion:
One good way to avoid all this is to use branches instead of working on master. for next time ;-)

oggm/utils/_funcs.py
```diff
@@ -753,8 +762,19 @@ def find_inversion_calving(gdir, water_depth=1, max_ite=30,
...
+            if fix_water_depth:
+                # We fix the water and calculate a pre-calving
+                # assuming the first thicknes to be:
+                cl = gdir.read_pickle('inversion_output')[-1]
+                t_altitude = cl['hgt'][-1]
+                thick = t_altitude + 1
+                out = calving_flux_from_depth(gdir,
+                                              water_depth=water_depth,
+                                              thick=thick,
+                                              fix_water_depth=True)
```
fmaussion: I don't understand this part. If we fix the water depth, shouldn't the original thickness be `water_depth + 1`? Or do we allow negative free-board? In any case, I don't think that `t_altitude` should play any role here, because this is what becomes a free parameter in the new loop.

bearecinos: Hello sorry for late reply ... took sort of Friday afternoon off.. I think you are right here I will check again, do some test and get back to you. Thanks

fmaussion:
@bearecinos I made some modifs to your code, the only significant changes are:
- the varying free board is now correctly computed and written out
- the `water_depth` kwarg is now called `initial_water_depth` 
- the `initial_water_depth` now defaults to 1/3 of the terminus elevation, clipped to a min of 10m

can you review these changes? Then you will have to re-run everything with the modifs I'm afraid...

Ideally your code (i.e. outside of OGGM) should be as minimal as possible: just setting up the run and analyze the results - does this PR help in this respect?

bearecinos:
Yes I will then re run things in case it gets delay I will just request another extension to the journal

fmaussion:
> It might take me more than a few days.

really? Then don't do it.

fmaussion:
OK Bea, I've implemented the changes we talked about yesterday. No external changes, just internal:
- moved the find_calving task to the inversion module in order to make a real task out of it (maybe you have to rename things in your sript but a find/replace should be enough)
- changed the internals of the glacier_statistics task, but this should not change anything in your code. (I also added more interesting output)

bearecinos:
Thanks I will implement this and let you know if I don't understand something.

PR #721, opened 2019-03-25
Dynamically download verification data

TimoRoth:
The new verification file is still generating and I only put a partial version in place for now.
Will update once it's done.


TimoRoth:
The file is generated completely now and remaining test issues have been fixed, merging as it is now to fix the verification issues at hand.

PR #725, opened 2019-03-27
Added task to get the pointwise catchment area and linear mass balance

MatCast:
- [x] Tests added/passed
- [x] Fully documented


fmaussion:
Thanks! This is looking good. I'll have a more detailed look next week.

fmaussion:
@MatCast this is very good thanks!

in terms of API this is not exactly what I was looking for - I want to hide this under a more general task. If you agree I'll add some changes on top of your PR, ok?

MatCast:
@fmaussion of course go ahead please.

MatCast:
I found a bug in the code which would generate a `ValueError` for some glaciers. I fixed it but let me merge the updated code before merging this into master. Or should i merge it with the corrected API that you had in mind as well?

fmaussion:
@MatCast sorry I missed your previous posts. I think I've changed too much now, and likely my version is buggy as well so I'll try to test it on more glaciers.

I'm thinking a bit bigger now because I also have interest to get more things related to ice thickness in OGGM, I'll add some documentation about this first,  then continue in another PR.

fmaussion:
Thanks @MatCast !

PR #726, opened 2019-03-29
Make plot functions work with merged glaciers

matthiasdusch:
Need this for a merged glacier tutorial.
Also added the workflow task to the API documentation.

- [x] what's new entry

### DIFF
--- a/oggm/graphics.py
+++ b/oggm/graphics.py
@@ -252,7 +252,8 @@ def plot_domain(gdirs, ax=None, smap=None):
             if isinstance(poly_pix, shpg.MultiPolygon):
                 for _poly in poly_pix:
                     for l in _poly.interiors:
-                        smap.set_geometry(l, crs=crs, color='black', linewidth=0.5)
+                        smap.set_geometry(l, crs=crs, color='black',
+                                          linewidth=0.5)
             else:
                 for l in poly_pix.interiors:
                     smap.set_geometry(l, crs=crs, color='black', linewidth=0.5)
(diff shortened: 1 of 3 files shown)


oggm/graphics.py
```diff
@@ -249,8 +249,14 @@ def plot_domain(gdirs, ax=None, smap=None):
...
-            for l in poly_pix.interiors:
-                smap.set_geometry(l, crs=crs, color='black', linewidth=0.5)
+            if isinstance(poly_pix, shpg.MultiPolygon):
+                for _poly in poly_pix:
+                    for l in _poly.interiors:
+                        smap.set_geometry(l, crs=crs, color='black',
+                                          linewidth=0.5)
+            else:
+                for l in poly_pix.interiors:
+                    smap.set_geometry(l, crs=crs, color='black', linewidth=0.5)
```
fmaussion: Would the following work? 

```python
poly_pix = utils.tolist(poly_pix)
for _poly in poly_pix:
    for l in _poly.interiors:
        smap.set_geometry(l, crs=crs, color='black',
                                         linewidth=0.5)
```

Alternatively:

```python
if not isinstance(poly_pix, shpg.MultiPolygon):
    poly_pix = [poly_pix]
```

(just to avoid duplication)

matthiasdusch: yes, I like that

matthiasdusch:
Note for future: This PR only fixes `graphics.plot_domain`, `graphics.plot_centerlines` and
  `graphics_plot_modeloutput_map` for merged glaciers. Other plot functions need more work or will never work as they for example need inversion results or catchment indices which are not processed for merged glaciers.

Issue #727, opened 2019-04-01
Versioneer is deprecated


fmaussion:
See https://github.com/pydata/xarray/issues/2853

TimoRoth:
That other version system looks super straight forward and clutter free. I'll have a look at it tomorrow.

PR #730, opened 2019-04-02
Switch to setuptools-scm

TimoRoth:
See Issue #727


fmaussion:
Nice! This implies no new dependency, right?

TimoRoth:
Only setuptools_scm and setuptools_scm_git_archive at setup time from how I understand it.

fmaussion:
If I install your branch from git, I get: 

```
>>> import oggm
>>> oggm.__version__
'0.1.dev763+gb361915'
```

This doesn't seem right?

TimoRoth:
My fork does not have any tags at all.
Or you didn't pull tags (git fetch --tags) in a very long time.

fmaussion:
> My fork does not have any tags at all.

Yeah, probably this

TimoRoth:
Pushed the tags from here to my fork as well now. Also, this is documented to not support anything else than 3 segment version schemes (major.minor.micro), but it seems to work just fine with v1.1 for all I can tell.

fmaussion:
yeah, v1.1 was a mistake. Let's get v1.1.1 out soon ;-)

can we merge this?

TimoRoth:
One advantage of omitting the micro version is that it titles master as 1.2-dev right now, which seems better if we want to release a 1.1.1 from the stable branch.

fmaussion:
Let's move this discussion in another issues, I'd like to make a clear policy about this

PR #733, opened 2019-04-03
Add a page about clusters in docs

fmaussion:
This is just a skeleton, in order not to forget I actually started to work on that.


fmaussion:
I merged the paralel computations and cluster pages in one, added some info about reproducibility and singularity containers.

Enough docs for today.... :roll_eyes:

PR #734, opened 2019-04-04
download function for JAXA's AW3D30 DEM

matthiasdusch:
- [x] Tests added/passed
- [x] Fully documented

This shall become a download function to use JAXA's ALOS World 3D 30m DEM within OGGM.
I'll put this up here as work in progress to get some comments on how to handle the files being on a ftp server.

@TimoRoth the OGGM `file_downloader` can not download files from an ftp server out of the box. Do you have any suggestions or ideas how to handle this?


oggm/utils/_downloads.py
```diff
@@ -741,6 +742,43 @@ def _download_arcticdem_from_cluster_unlocked():
...
+    outpath = os.path.join(tmpdir, zone + '_AVE_DSM.tif')
+
+    # check if extracted file exists already
+    if os.path.exists(outpath):
+        return outpath
+
+    # Did we download it yet?
+    ftpfile = ('ftp://ftp.eorc.jaxa.jp/pub/ALOS/ext1/AW3D30/release_v1804/'
+               + zone + '.tar.gz')
+    dest_file = file_downloader(ftpfile)
```
matthiasdusch: this doesn't work right now

TimoRoth: requests does not support downloading from ftp. I searched quite a bit on how to do it, and only found ugly hacks and conclusions that it's an old protocol and they want people to use https instead.

fmaussion: This is a clear incentive to solve the license issue. Since we have the files on cluster we could either:
- make them open and download from there
- hide them behind a password and use the same technique as for Tandem-X with a local credential file. Then we can control who gets the password (my preferred approach from now)

matthiasdusch:
Thanks for the comments and for #735 !
I'll finish this tomorrow and yes I agree, we (e.g. I) should tackle the licenses.

matthiasdusch:
- [x] fixed a typo in` _classic_urlretrieve`
- [x] finished the AW3D30 downloader

@OGGM/oggm-devs how is the policy on download tests? Should I add one which compares the DEMs?

aw3d30

docs/whats-new.rst
```diff
@@ -20,6 +20,11 @@ Enhancements
   mass-conservation inversion (`inversion.find_inversion_calving`). This
   is still in experimentation phase! (:pull:`720`).
   By `Beatriz Recinos <https://github.com/bearecinos>`_.
+- The ALOS Global Digital Surface Model "ALOS World 3D - 30m" DEM from JAXA can
+  now be used as alternative DEM within OGGM.
+  `See our tutorial <http://edu.oggm.org/en/latest/oggm_tuto.html>`_ on how to
+   set an alternative DEM (:pull:`734`).
+  By `Matthias Dusch <https://github.com/matthiasdusch>`_.
```
fmaussion: Nit: space missing in front of "by"

oggm/utils/_downloads.py
```diff
@@ -45,6 +45,7 @@
                              DownloadVerificationFailedException,
                              DownloadCredentialsMissingException,
                              HttpDownloadError, HttpContentTooShortError)
+from oggm import utils
```
fmaussion: There is a risk of circular dependencies here - utils is in fact three modules including this one. Remove this line and import directly in the function itself with `from oggm.utils._workflow import robust_tar_extract`

fmaussion: (reason: not so good design of a huge utils module...)

oggm/utils/_downloads.py
```diff
@@ -961,6 +1006,50 @@ def tandem_zone(lon_ex, lat_ex):
...
+    # But in N and E the lower boundary is indicated
+    # e.g. N060 contains N060 - N064
+    # e.g. E000 contains E000 - E004
+    # but S and W indicate the upper boundary:
+    # e.g. S010 contains S006 - S010
+    # e.g. W095 contains W091 - W095
+
+    # get letters
+    NS = 'S' if lat_tile < 0 else 'N'
+    EW = 'W' if lon_tile < 0 else 'E'
```
fmaussion: nit: keep `ns` and `ew` lower case

oggm/utils/_downloads.py
```diff
@@ -961,6 +1006,50 @@ def tandem_zone(lon_ex, lat_ex):
...
+
+    # get lat/lon
+    lon = abs(5 * np.floor(lon_tile/5))
+    lat = abs(5 * np.floor(lat_tile/5))
+
+    folder = '%s%.3d%s%.3d' % (NS, lat, EW, lon)
+    file = '%s%.3d%s%.3d' % (NS, abs(lat_tile), EW, abs(lon_tile))
```
fmaussion: nit: `file` is a python dedicated name - use `fpath` instead (or similar)

matthiasdusch:
Thanks for the review. And I'll add the tests.

oggm/utils/_downloads.py
```diff
@@ -341,7 +342,7 @@ def _requests_urlretrieve(url, path, reporthook, auth=None):
             raise HttpContentTooShortError()
 
 
-def _classic_urlretrieve(url, path, reporthook, auth=None):
+def _classic_urlretrieve(url, cache_path, reporthook, auth=None):
```
TimoRoth: I just fixed this in the opposite way myself, so this change here needs undone or it will introduce the same issue again, just reversed.

matthiasdusch:
decided to remove the SRTM vs. AW3D30 test again

PR #739, opened 2019-04-07
Do not fail when retrieving hashes without internet connection

fmaussion:
Add-hoc fix while traveling


TimoRoth:
Could probably just use req.raise_for_status() instead of the inner check. Avoids a bit of duplication.

fmaussion:
> Could probably just use req.raise_for_status() instead of the inner check. Avoids a bit of duplication.

I don't understand how to do this - do you want to give it a go? Can wait till tomorrow, I have internet now

fmaussion:
thanks!

PR #745, opened 2019-04-16
Consistent parameter names in run output dataset

oberrauch:
The parameter names in the run output datasets produced by different tasks are not the same. The dataset returned by `FlowlineModel.run_until_and_store()` has units as suffixes (for the parameters length, area, volume, etc.) while the dataset returned by `utils.compile_run_output()` does not. Given that the datasets are otherwise identical in structure, it seems useful to make the parameter names coherent.
I chose to delete the suffixes, since the parameters store the unit as attribute. On the other hand, most variables in the code base have their units in the name as suffix. Therefore I am not sure which method to prefer... Anyway, it should used consistently.

### DIFF
--- a/oggm/core/flowline.py
+++ b/oggm/core/flowline.py
@@ -798,24 +798,24 @@ class FlowlineModel(object):
         diag_ds['calendar_month'].attrs['description'] = 'Calendar month'
 
         # Variables and attributes
-        diag_ds['volume_m3'] = ('time', np.zeros(nm) * np.NaN)
-        diag_ds['volume_m3'].attrs['description'] = 'Total glacier volume'
-        diag_ds['volume_m3'].attrs['unit'] = 'm 3'
-        diag_ds['area_m2'] = ('time', np.zeros(nm) * np.NaN)
-        diag_ds['area_m2'].attrs['description'] = 'Total glacier area'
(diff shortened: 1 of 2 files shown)


juliaeis:
Do you also plan to rename names of the attributes of a flowline, like model.fls[-1].length_m/model.fls[-1].length_km ?  Because, I would be really annoyed, if this would also change in the future, because I would have to change all my scripts then. As I don't use the dataset itself, the changes in this pull request don't affect me :)

fmaussion:
Yeah we have to discuss this - I'll try to have a closer look when I find time

oberrauch:
> Do you also plan to rename names of the attributes of a flowline, like model.fls[-1].length_m/model.fls[-1].length_km ? Because, I would be really annoyed, if this would also change in the future, because I would have to change all my scripts then. As I don't use the dataset itself, the changes in this pull request don't affect me :)

Hey @juliaeis, I think that is not reasonable. Especially since certain parameters are stored with different units. For those parameters the unit suffixes makes perfect sense. In order to be overall consistent, we could therefore add unit suffixes to the `Dataset` produced by `utils.compile_run_output()`...
I'm well aware that this changes may cause some problems with existing scripts, but hopefully simplify future usage... But if my suggested changes cause more problems than they solve, I'm happy with the current implementation.

PR #746, opened 2019-04-18
possible bug in get_ref_mb_glaciers_candidates

matthiasdusch:
For me line 1874 in utils.`_downloads.py`:
`cfg.DATA[key] = flink['RGI{}0_ID'.format(rgi_version)].to_list() `
does not work:
`AttributeError: 'Series' object has no attribute 'to_list'`

I think it should be `tolist()`

**But**: How does this code pass the tests and also the crossvalidation on the cluster? Maybe I got something wrong here...

### DIFF
--- a/oggm/utils/_downloads.py
+++ b/oggm/utils/_downloads.py
@@ -1871,7 +1871,7 @@ def get_ref_mb_glaciers_candidates(rgi_version=None):
 
     if key not in cfg.DATA:
         flink, _ = get_wgms_files()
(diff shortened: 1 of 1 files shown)


fmaussion:
Thanks! See https://github.com/pandas-dev/pandas/pull/23398 

So since recent pandas versions both are correct, but tolist is more compatible...

matthiasdusch:
I see. I checked (aka googled)  for pandas to_list method, but got no results...

However you prefer, I also don't mind ditching this and updating my pandas :-)

fmaussion:
let's avoid someone else's future problem then!

PR #767, opened 2019-05-15
Docs: mass balance calibration with HISTALP

oberrauch:
The mass balance calibration script results in an error when calibration with HISTALP climate data. Some reference glaciers in the RGI region 11 are outside the HISTALP domain. For those glaciers no climate data is processed. This result in a `RuntimeError('Please process some climate data before call')` thrown by the task `utils.get_ref_mb_glaciers`.

To prevent this I subset the glaciers for RGI subregion 11-01.

### DIFF
--- a/docs/run_examples/_code/run_reference_mb_glaciers.py
+++ b/docs/run_examples/_code/run_reference_mb_glaciers.py
@@ -77,9 +77,12 @@ log.info('Process the climate data...')
 if baseline == 'CRU':
     execute_entity_task(tasks.process_cru_data, gdirs, print_log=False)
 elif baseline == 'HISTALP':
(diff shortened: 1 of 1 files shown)


fmaussion:
Thanks! Could you try to remove the    `cfg.PARAMS['continue_on_error'] = True  # Some glaciers are not in Alps` and later `cfg.PARAMS['continue_on_error'] = False` and see if it still works?

oberrauch:
You are right, it is a bit redundant. There are no errors occurring with the glaciers in RGI subregion 11-01.

fmaussion:
Can you make the change and update the PR then?

fmaussion:
Thanks @oberrauch !

PR #776, opened 2019-05-21
Remove dependency on boto3

TimoRoth:
It was causing a slight dependency headache, and the files it tries to download are long gone anyway.

The download function itself is still left in place, so it will continue to work if the files are in local cache for now.
Once we figured out what to do about those files, it should be replaced with a regular download function.

### DIFF
--- a/docs/installing-oggm.rst
+++ b/docs/installing-oggm.rst
@@ -58,6 +58,7 @@ GIS tools:
 
 Testing:
     - pytest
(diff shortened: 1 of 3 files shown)


fmaussion:
Thanks!

> Once we figured out what to do about those files, it should be replaced with a regular download function.

Yeah, will go for scp for now, that's not urgent

PR #798, opened 2019-06-14
Update test container

TimoRoth:
I'm pretty sure this will break things, and since I pinned pyproj to <2.0 on this one, some other update must have made a mess as well.
If not, it really is all just caused by pyproj in some ways.


fmaussion:
I'm betting 5 euros on all green!

TimoRoth:
Thing is, on Anaconda(both for 3.6 and 3.7) and Travis-Python 3.7 I'm seeing this breakage:
https://travis-ci.org/OGGM/OGGM-Anaconda/jobs/545222912#L1006
And that is with pyproj pinned to <2.0.

fmaussion:
actually we shouldn't even download these things from our tests (410.4 MiB)

I'll have a look at the VASScaling tests

fmaussion:
OK, let's hope the new test will look different - can you rebase once my pr is merged? 

The mpl test failures are probably unrleated

PR #809, opened 2019-06-20
add bias option for run_from_climate_data()

juliaeis:
This change add an optional parameter for the bias in run_from_climate_data()


fmaussion:
Thanks! Can you add the keyword argument to the docstring as well? See task above for an example.

juliaeis:
sorry, for the unnecessary commits cause to wrong changes and spelling!

fmaussion:
> sorry, for the unnecessary commits cause to wrong changes and spelling!

this is totally not important - we squash everything at merge anyway

fmaussion:
Thanks @juliaeis !

Issue #549, opened 2018-10-16
DOC: document the OGGM classes


fmaussion:
Amongst all other things the OGGM classes like Flowline, FlowlineModel, MassBalanceModel are very important to understand. It would be good to document them proper (following the model of MassBalanceModel, using a Metaclass) and have a dedicated page for them in the docs.

PR #811, opened 2019-06-21
DOC updates, mainly flowline(model)

matthiasdusch:
I wanted to tackle #549 but I am not sure about the metaclass. 
Started writing a more detailed Flowline description comparable to https://docs.oggm.org/en/stable/api.html#mass-balance 

Most important part for me is the "hierarchy of center/flowlines", always confuses me.

Will happily take feedback where to actually put this and will continue in IBK.


fmaussion:
I like it! I think it is well placed in API -> more model internals should be added here. I think it's almost enough to merge already....

matthiasdusch:
I kind of forgot about this...  and now I obviously will face conflicts... But determined to finish this today

PR #815, opened 2019-06-25
Add parameter to initialise the glacier for the random climate task. 

oberrauch:
Adding `init_area_m2` parameter to the VAS random climate task, in analogy to `init_model_fls`.


fmaussion:
Thanks @oberrauch !

PR #829, opened 2019-07-30
RGI attributes used by oggm docs

bearecinos:
Here are the additions to the documentation. Only like I said links between the docs and the notebook are missing but you can add them once things are online. 

Let me know if you need any change


fmaussion:
Thanks @bearecinos this is great!

Issue #820, opened 2019-07-01
assert in `flows_to_indice` is called on undefined variable `ind`


matthiasdusch:
https://github.com/OGGM/oggm/blob/13ca306fbfe1ce103697120bd2c0d88b2af11714/oggm/core/centerlines.py#L144-L152

I'll fix that myself

fmaussion:
> I'll fix that myself

That made my day ;-)

matthiasdusch:
by #834

PR #834, opened 2019-08-01
indices had to been treated differently

matthiasdusch:
always something: for some cases bed got trapezoidal. This is stored with indices rather than on the actual grid points. I did not realize that before.

### DIFF
--- a/oggm/core/centerlines.py
+++ b/oggm/core/centerlines.py
@@ -149,7 +149,7 @@ class Centerline(object):
         for i, p in enumerate(self.flows_to.line.coords):
             if p == tofind:
                 ind.append(i)
-        assert len(ind) == 1
+        assert len(ind) == 1, 'We expect exactly one point to be found here.'
         return ind[0]
 
     @lazy_property
@@ -161,7 +161,8 @@ class Centerline(object):
             ind = [i for (i, pi) in enumerate(self.line.coords)
                    if (p.coords[0] == pi)]
             inds.append(ind[0])
-        assert len(inds) == len(self.inflow_points)
+        assert (len(inds) == len(self.inflow_points),
+                'For every inflow point should be exactly one inflow indice')
         return inds
 
     @lazy_property
--- a/oggm/core/centerlines.py
+++ b/oggm/core/centerlines.py
@@ -144,12 +144,13 @@ class Centerline(object):
     def flows_to_indice(self):
         """Indices instead of geometry"""
 
+        ind = []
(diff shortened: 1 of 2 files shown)


fmaussion:
> always something: for some cases bed got trapezoidal.

Sorry about that :-( 

I think you noticed how dirty this initialize_present_day_glacier task really is. I wonder if we will ever find the courage to make it better...

oggm/core/flowline.py
```diff
@@ -237,7 +237,7 @@ def __init__(self, line=None, dx=None, map_dx=None, surface_h=None,
         if np.any(self._w0_m <= 0):
             raise ValueError('Trapezoid beds need to have origin widths > 0.')
 
-        self._prec = np.where(lambdas == 0)
+        self._prec = np.where(lambdas == 0)[0]
```
fmaussion: Wait, but `TrapezoidalBedFlowline` is never used operationally.

fmaussion: We use MixedBedFlowline if I recall well!

matthiasdusch: yes, I just added this on the fly to get an array and not a tuple(array). Not strictly necessary for the TrapezoidalBed but for the MixedBed. I could account for this at the time where I change the attribute values if you prefer, but I think it is nicer to have the actual array as an attribute like it is true for ['_ptrap']

fmaussion: No, that's fine. You probably are more into it than I am currently.

matthiasdusch:
also fixing #820  
which now raises an AssertionError if there is more or less than 1 `flows_to_indices`

oggm/core/centerlines.py
```diff
@@ -144,12 +144,13 @@ def set_line(self, line):
...
 
+        ind = []
         tofind = self.flows_to_point.coords[0]
         for i, p in enumerate(self.flows_to.line.coords):
             if p == tofind:
-                ind = i
-        assert ind is not None
-        return ind
+                ind.append(i)
+        assert len(ind) == 1
```
fmaussion:
```suggestion
        assert len(ind) == 1, 'We expect only one line to be found here'
```

I just learned about this [here](https://softwareengineering.stackexchange.com/questions/225956/python-assert-vs-if-return) and it just blows my mind I didn't know about this earlier.

Asserts are bad practice because their error message is obscure, but the possibility to actually add a message makes them much more useful in such cases!!!!

PR #836, opened 2019-08-02
Fix for numpy change in clip

fmaussion:
This was a really hard one

Numpy changed the `clip` function internally (https://docs.scipy.org/doc/numpy/release.html#clip-now-uses-a-ufunc-under-the-hood), and now clip(-1, 0, -2) returns -2!

This was a kind of bug in OGGM, but still. I hope to have properly checked all occurrences of clip now...

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -7,7 +7,7 @@ dependencies:
   - scipy
   - pandas
   - shapely
-  - matplotlib<3.0.0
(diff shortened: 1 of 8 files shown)


fmaussion:
@TimoRoth this should solve the remaining conda issues. This was a bad bug and I'm glad the tests took it...

PR #848, opened 2019-08-12
Remove fiona and GDAL from setup.py

fmaussion:
These two are never imported in OGGM - they are an indirect dependency. 

Testing out pip install based on wheels only that *might* actually work out of the box


fmaussion:
cc @TimoRoth who wrote this setup

Issue #777, opened 2019-05-21
Alternatives to ReadTheDocs?


fmaussion:
ReadTheDocs is great, but their build system is a real pain. It's always causing me trouble, either here or on rgitools or elsewhere. The main issue is that they want to install the build environment *each time*, which creates dependency problems, takes time, and gets killed because it takes time.

An alternative would be to use travis to build the docs via sphinx in our prepared containers. I thought there would be more tools online to do this, but a quick search yielded:

https://github.com/Syntaf/travis-sphinx

We would then source the built documentation in a dedicated github repo and use github pages to serve them.

The main disadvantage is that we wouldn't have the nice features of ReadTheDocs, like the ability to browse older versions of the documentation.

PR #851, opened 2019-08-12
Use pip on ReadTheDocs

fmaussion:
Closes https://github.com/OGGM/oggm/issues/777

As unbelievable as it may sound, it is now possible to use `pip` on readthedocs, without installing conda.


fmaussion:
xref: https://github.com/readthedocs/readthedocs.org/issues/6025

Issue #420, opened 2018-03-05
Remove some of OGGM's hard dependencies


fmaussion:
Currently one needs to install a LOT of packages to get OGGM to be simply *imported*. Some users (e.g. https://github.com/OGGM/oggm/issues/418) could use a much lighter version. 

Very heavy packages (only useful for the operational workflow and traditionally hard to install) include:
- rasterio 
- gdal
- fiona

Getting those as an optional dependency would be very useful.

fmaussion:
There are at least two ways to do this I know about:
- keeping the import statements at the top of the module but put them in a try .. catch statement
- put the imports in the function where there are used, thus raising an import error at runtime only

I think option 2 is better, and should not be too hard to implement.

fmaussion:
This becomes more and more important, now that OGGM-Edu is gaining momentum. For plenty of applications we don't need all the GDAL/RASTERIO things...

PR #852, opened 2019-08-13
Make OGGM flowline run without rasterio/geopandas

fmaussion:
Closes https://github.com/OGGM/oggm/issues/420

This is going to be interesting. Not sure if the travis config is the most clever ever, but maybe @TimoRoth can have a look as well.

- [x] Tests added/passed
- [x] Fully documented


fmaussion:
This is the new list of *required* packages:

```
req_packages = ['numpy',
                'scipy',
                'pandas',
                'matplotlib>=2.0.0',
                'shapely',
                'requests',
                'configobj',
                'netcdf4',
                'xarray',
                ]
```

I think this is quite sustainable for minimal configs (getting rid of shapely would be good as well but more involved).

Issue #875, opened 2019-10-01
Retain calving front width from inversion.find_inversion_calving() output


bearecinos:
Hello @fmaussion ,

I noticed that in this version the width of the calving front is not retain in the output of the function. This variable is needed for Marco's paper where he estimates calving fluxes with velocity data and OGGM glacier cross-sections at the terminus. 

Can we add the calving front width in the output of the function so it gets save in the glacier characteristics file:

Width is not there anymore: 
https://github.com/OGGM/oggm/blob/1d82e25cf30fa8a7af38b2b9bdfafa74bacb54fe/oggm/core/inversion.py#L933

Its in the old output of the function: `df = pd.read_csv(gdir.get_filepath('calving_loop'), index_col=0),` but not on the new one.
https://github.com/OGGM/oggm/blob/1d82e25cf30fa8a7af38b2b9bdfafa74bacb54fe/oggm/utils/_workflow.py#L1046

Let me know if there is a easy way to fix it.
Cheers 
Bea

PR #876, opened 2019-10-02
Adding width to calving output

bearecinos:
Hello, 

Following the issue raised yesterday: https://github.com/OGGM/oggm/issues/875
I would like that in the glacier statistics function, we can also write the width of the calving front. I have now added a line in the inversion.find_inversion_calving() function so this variable can be included in the diagnostics.

Let me know if this is the way of solving this.
Thanks


bearecinos:
@fmaussion  I have added the velocity function to the inversion.py and added that width parameter to the output of the new calving fucntion. Let me know if I need to change anything else so the tests pass.

fmaussion:
Thanks! For the sake of time I did some edits myself: note that I changed some names and that you now have to run `prepare_for_inversion` with `add_debug_var=True` for this to work. Please have a look and let me know.

bearecinos:
Thanks for merging !

PR #880, opened 2019-10-03
Switch to Python 3.7 as main test environment

TimoRoth:
Should be good to go assuming all tests pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,53 +24,53 @@ env:
 matrix:
   fast_finish: true
   allow_failures:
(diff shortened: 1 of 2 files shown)


fmaussion:
I restarted the failing build - but the 3.7 runs seem slower than 3.6. maybe I didn't solve the numpy 1.17 problem in the end? I really thought I would

TimoRoth:
I downgraded numpy back to 1.16 and now the times are back to normal.

fmaussion:
Thanks! I'm really confused as to this numpy 1.17 problem, but I don't have time to have a look at the moment...

PR #884, opened 2019-10-10
switch to ASTER v3 DEM

matthiasdusch:
ASTER v3 was released in June 2019:
https://lpdaac.usgs.gov/products/astgtmv003/

They filled some voids with newer data. But overall I think it is still of little use to us because it has artefacts which are not flagged.
But it is a global DEM in a good resolution and it was mostly produced between 2000-2013 which is close to the RGIv6 date. It might be useful for individual cases if checkt thoroughly.

Also: ASTER v3 is (now?) available via NASA Earthdata. To access one needs to register at https://urs.earthdata.nasa.gov and store username and password locally to allow OGGM to automatically download the necessary DEM tiles. This is similar to how we deal with TanDEM-X. And might also be the case for future DEMs or other data.
As I don't think having one credentials file per host is a sleek solution I turned to Python's netrc library which uses a `.netrc` file to store and parse credentials to different hosts. This file is mainly used for automatic FTP connections or other remote services.
https://www.gnu.org/software/inetutils/manual/html_node/The-_002enetrc-file.html
@TimoRoth  and @fmaussion I am interested if you agree with that change.

I also put the part of the actual download code where the cache is checked first without credentials into a separate function to avoid duplicate code.

I might have to think about the `_test_credentials` again. I don't like to download actual data for this but the NASA server will return a valid header information even with wrong credentials.
But right now all connection requests get rejected as I might have poked the NASA-bear a bit too often with wrong credentials :-D

- [x] Tests added/passed
- [x] Fully documented


TimoRoth:
Looks generally sensible to me. Just a bit annoying that users who already use the current credentials file will have to re-login, but I think that's such a small amount of people that it hardly matters.

matthiasdusch:
I tried to write some tests, but had some problems with the password prompt... I'll probably merge that after travis passes and maybe come back to the tests later.

fmaussion:
> I tried to write some tests, but had problems with the password prompt...

SWorry I was not clear enough - these tests are hard indeed. But sometimes you can test other parts of the CLI, such as "is it behaving well" etc. I haven't looked in detail if this possible or not here.

But as a general rule, any internet test is not reliable and won't be done on Travis anyway

matthiasdusch:
I was unclear as well: I only wanted to test the behavior of the CLI. But almost every call (especially in a test environment without existing .netrc file) will trigger a username/password prompt. And while username is no problem, the `getpass.getpass` password prompt just behaves differently than standard input. I did not manage to automatically pass a (fake) password during the test. Of course it's possible somehow I just didn't find a solution.

Issue #850, opened 2019-08-12
Adding a dem_date attribute to GlacierDirectory


juliaeis:
Similar to the gdir.rgi_date atribute, it would be good to have a gdir.dem_date attribute

PR #885, opened 2019-10-11
Add DEM source information

matthiasdusch:
Every GlacierDirectory gets a file `dem_source.txt` where we specify the used DEM and advise the user how to cite it in publications. There were still some DEMs missing.

Also added a "Date of acquisition" to the source text. Could be used for #850 but for most DEMs this is a quite broad range of a decade or so...

### DIFF
--- a/docs/input-data.rst
+++ b/docs/input-data.rst
@@ -385,6 +385,12 @@ reproduce this information
     Visit `rgitools <https://rgitools.readthedocs.io/en/latest/dems.html>`_
     for a discussion about our current efforts to find "the best" DEMs.
 
+.. note::
+
+    `In this blogpost <https://oggm.org/2019/10/08/dems/>`_ we talk about which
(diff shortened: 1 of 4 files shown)


matthiasdusch:
also this lead to the blog post PR here https://github.com/OGGM/oggm.github.io/pull/78 which should be merged first to make sure the links in the source text work.

fmaussion:
Thanks! I think we should tackle https://github.com/OGGM/oggm/issues/850 in this PR. Let's chat tomorrow about the way to go

matthiasdusch:
> I think we should tackle #850 in this PR. 

Had an first attempt at this, but still WIP or at least thinking in progress...

matthiasdusch:
in the previous commit I read the daterange and info from the dem_sources.txt template. Now I go for the actual dem_source.txt in the GDir which is more sensible I think.

PR #886, opened 2019-10-17
Some DEM download fixes and logging

matthiasdusch:
- AW3D30 raises a specific URL error if tile is not available. Catch this and return `None`. This could be an ocean tile.
- AW3D30 is really bad for Antarctica ([see map](https://www.eorc.jaxa.jp/ALOS/en/aw3d30/index.htm)) I suggest to not even try getting a DEM here and implemented this.
- There was a bug in REMA, which means nobody ever used this o_O
- Added a log-entry to the `_rename_dem_folder` function: This should allow to make nice statistics for the RGI DEM tests.

### DIFF
--- a/oggm/cli/prepro_levels.py
+++ b/oggm/cli/prepro_levels.py
@@ -62,6 +62,9 @@ def _rename_dem_folder(gdir, source=''):
         f = gdir.get_filepath(fname)
         os.rename(f, os.path.join(out, os.path.basename(f)))
 
+    # log SUCCESS for this DEM source
+    gdir.log('{},DEM SOURCE,{}'.format(gdir.rgi_id, source))
+
 
 def run_prepro_levels(rgi_version=None, rgi_reg=None, border=None,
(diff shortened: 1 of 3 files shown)


fmaussion:
THANKS!

> There was a bug in REMA, which means nobody ever used this o_O

yes this is a drawback of the current CLI, which just ignores a DEM if not working. I don't know if there is a better way...

matthiasdusch:
Also, I don't think there is a proper error raised if one of multiple tiles is missing (returned as None by the downloader). None tiles are ignored as this could be just ocean at the border. But I think creating the glacier specific DEM from several tiles still 'works' even if one necessary tile is missing.

I have to double check that though.

fmaussion:
> But I think creating the glacier specific DEM from several tiles still 'works' even if one necessary tile is missing.

Yes I think so too

fmaussion:
Excellent thanks!

Issue #887, opened 2019-10-18
missing DEM tile might go without error


matthiasdusch:
I discussed this yesterday with @fmaussion and we came to the conclusion that a missing DEM tile will go unnoticed by `define_glacier_region` but should be detected during `glacier_masks` as the DEM is smoothed and interpolated.

But now I think it still might go undetected as areas where the DEM is missing are not loaded as NaNs or nodata values but 0. 

After `define_glacier_region`:
![define_glacier_region](https://user-images.githubusercontent.com/37663627/67094871-f21e7380-f1b4-11e9-8303-99df010b6cfe.png)

After `glacier_masks`, slightly interpolated:
![glacier_mask](https://user-images.githubusercontent.com/37663627/67094780-c3080200-f1b4-11e9-91b3-44dc0ae4d356.png)

Notebook:
https://nbviewer.jupyter.org/gist/matthiasdusch/f8cb20222d3c974aa37bd928c489e561

There is still the possibility that this will raise an error in a later processing step. I'll investigate this next week with a fresh brain...

Oh, and the reason why this tile is missing in the first place there: A bug in my ASTER implementation. Who would have thought glaciers can span more than two degree in longitude or latitude...

fmaussion:
Nice catch! I'm still worried that the merge_raster tool is actually filling with zeros instead of NaNs. Will have a closer look.

PR #890, opened 2019-10-18
fix silent zero fill when DEM are missing

fmaussion:
Closes https://github.com/OGGM/oggm/issues/887

- [x] Tests added/passed
- [ ] Fully documented

Still WIP:
- I need to check simple_glacier_masks too
- the check on how many percent data is missing should also happen on the glacier mask, not on the full DEM: the reason is that many coastal glaciers will have missing data but its not a big deal there: need some more thinking

cc @matthiasdusch

### DIFF
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -28,6 +28,11 @@ Enhancements
 Bug fixes
 ~~~~~~~~~
 
+- Small bugs in DEM processing fixed - internal refactoring followed,
+  hopefully for the good (:pull:`890` and :pull:`886`).
+  By `Fabien Maussion <https://github.com/fmaussion>`_ and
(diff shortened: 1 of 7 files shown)


fmaussion:
With the update in plot:

from geotiff:

tiff

after glacier masks:

after_mask

fmaussion:
OK so this has become more involved than originally intended. The DEM processing is now more robust.

On the way I found several inconsistencies between the GIS tasks and actually planned to make glacier_masks two task instead of one:  `process_dem`, which filters the DEM and smoothes it, and `glacier_masks` which computes the mask. I decided against a deprecation cycle enforcing the use of the two tasks for now and just silently call `process_dem` from within `glacier_masks`.

There is some work to do for RGI tools, will open a separate issue for that.

PR #897, opened 2019-11-07
add rasterio_glacier_mask

matthiasdusch:
- [x] Tests added/passed

This function will read the GDir DEM and the corresponding RGI outlines and write a `dem_mask.tif` file It has the same extent and projection as the DEM but only consists of zeros (unglaciated grid points) and ones (glaciated grid points).

It does *not* contain NaNs or other nodata values, as this mask is independent of the original source DEM.


fmaussion:
Great! Can we add a quick test for this? Just in test_prepro, check that the mask area is more or less close to RGI area

fmaussion:
Or maybe based on this: https://github.com/OGGM/oggm/blob/master/oggm/tests/test_prepro.py#L248

PR #900, opened 2019-11-08
Remove Alex's SIA fluxlimiter from the codebase and import it instead

fmaussion:
This was needed for license issues. Once https://github.com/alexjarosch/sia-fluxlim/pull/2 is merged one needs to install Alex's package from github before the tests here can pass.

cc @TimoRoth fyi


fmaussion:
@TimoRoth https://github.com/alexjarosch/sia-fluxlim/pull/2 is merged, so now we need to add it as dependency and update the docker image here - there is no rush though, that's not urgent.

TimoRoth:
Is it on pypi and conda-forge?

fmaussion:
> Is it on pypi and conda-forge?

No... I could ask Alex to do so though (it's not my package). This will have to wait to next week so that I can tell him how to do that.

fmaussion:
Since it's really just **one class** and we use it exclusively for testing, and it was originally in the OGGM codebase, an alternative would be to ship one GPL licensed module with OGGM (just [this one](https://github.com/alexjarosch/sia-fluxlim/blob/master/sia_fluxlim/oggm_flowline.py)) and don't care about an additional dependency. I think I'd prefer that.

Alternatively, we can make the tests pass when users don't have it installed (relatively easy to do). 

I don't know what's best. Right now I'd tend to ship the class as GPL licensed within the testing module in fact...

fmaussion:
yeah I went for it now

PR #902, opened 2019-11-11
don't use actual DEM for rasterio_glacier_mask

matthiasdusch:
If the glacier DEM is covered in NaNs, the mask computation does not work. But wo do not really need the actual DEM values, we just want the DEM profile (projection, extents, ...). So just use an array of the respective size instead.


fmaussion:
Looks sound to me!

PR #904, opened 2019-11-12
change dem_mask name and remove .properties file

matthiasdusch:
glacier_mask might be a better name than dem_mask

also, if oggm_prepro is run with the docker container in singularity, it creates a weird `outlines.tar.gz.properties`. Let's remove that before packing up because that are 200k less files.

### DIFF
--- a/oggm/cfg.py
+++ b/oggm/cfg.py
@@ -116,7 +116,7 @@ BASENAMES['dem'] = ('dem.tif', _doc)
 _doc = ('A glacier mask geotiff file with the same extend and projection as '
         'the `dem.tif`. This geotiff has value 1 at glaciated grid points and '
         ' value 0 at unglaciated points.')
-BASENAMES['dem_mask'] = ('dem_mask.tif', _doc)
+BASENAMES['glacier_mask'] = ('glacier_mask.tif', _doc)
 
 _doc = ('The glacier outlines in the local map projection (Transverse '
         'Mercator).')
--- a/oggm/cli/prepro_levels.py
+++ b/oggm/cli/prepro_levels.py
@@ -210,6 +210,9 @@ def run_prepro_levels(rgi_version=None, rgi_reg=None, border=None,
         workflow.execute_entity_task(gis.rasterio_glacier_mask,
                                      gdirs, source='ALL')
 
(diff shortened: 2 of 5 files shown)


matthiasdusch:
document for future:
I first thought the tar.gz.properties file is from compressing while creating it. But I am pretty sure now it's from reading the compressed shapefile. And it is most likely created by either Geopandas or Fiona. A bit hard to track as this only happens with the packages used in the current docker-image.

fmaussion:
> I first thought the tar.gz.properties file is from compressing while creating it.

OK then maybe we should suppress them at the tar creation step (for all tar.gz shapefiles)

matthiasdusch:
> > I first thought the tar.gz.properties file is from compressing while creating it.
> 
> OK then maybe we should suppress them at the tar creation step (for all tar.gz shapefiles)

No, the file is not created while writing or compressing but later while reading the compressed shapefile. Here:
```
geometry = gdir.read_shapefile('outlines').geometry[0]
```

So this is done by either Geopandas or Fiona. Or most likely and more specific by some module which is used withing gpd/fiona to open the tar.gz file...

fmaussion:
Yes, but it still ends up in the final tar file, because the outlines (but not the intersects) are read by your rasterio_glacier_masks task -> my point is that we should not write them down in tar.

OR, we suppress them at the gdir level which is the only point where we read these files.

fmaussion:
> we suppress them at the gdir level which is the only point where we read these files.

Here:

https://github.com/OGGM/oggm/blob/75e2c9494d03e0c2fcfd4f835fff72a675c7c868/oggm/utils/_workflow.py#L1954

matthiasdusch:
ah ok, now I get it. 

Right now I delete this file at the end of `rasterio_glacier_mask`. Suppressing it in the first place would be sensible but needs some digging. 

We could also check for it and delete it in the read_shapefile funktion (like you suggested, while I am writing this)
I didn't want to mess with these internals but I am happy to move the 4 lines of code to `_read_shapefile_from_path`

fmaussion:
> I didn't want to mess with these internals but I am happy to move the 4 lines of code to _read_shapefile_from_path

Yes, let's do that

Issue #909, opened 2019-11-23
Timestepping misconceptions and flaws in OGGM


alexjarosch:
Dear OGGM developers,

as OGGM gains momentum and is/will be used widely it is paramount to have basic yet vital numerical details correctly implemented. Else OGGM risks its own reputation, the plausibility of scientific results produced with OGGM and might in an extreme scenario (cf. climate-gate) negatively influence the reputation of glacier simulations as a whole.

Thus I urge the developers to seriously consider my suggestions below:

The basic idea of a CFL criteria for advection type problems (this is how OGGM views ice-flow) https://github.com/OGGM/oggm/blob/1dc07f57cf8f7d7708585a6f7ab972890924892e/oggm/core/flowline.py#L935 is correctly implemented: https://github.com/OGGM/oggm/blob/1dc07f57cf8f7d7708585a6f7ab972890924892e/oggm/core/flowline.py#L1113

Nevertheless how this criteria is subsequently treated raises some problems:

1. The concept of counter-limiting a time step calculated from a given CFL conditions is numerically not justifiable, nor sane. In case of OGGM this is the current argument https://github.com/OGGM/oggm/blob/1dc07f57cf8f7d7708585a6f7ab972890924892e/oggm/core/flowline.py#L939-L942
However this leaves the OGGM model on "very thin ice", so to say. Numerically such a strategy is worse than just setting a fixed, rather small time step as it deludes the user to think that OGGM applies a CFL to avoid instabilities and mass conservation issues.

2. Looking at the current choices of time stepping strategies https://github.com/OGGM/oggm/blob/1dc07f57cf8f7d7708585a6f7ab972890924892e/oggm/core/flowline.py#L965-L983 results in the realization that none of them can be justified from even the most basic time stepping stability theory. This is pure numerical engineering and would **at least** require some numerical test cases to prove that any of those choices make sense. Just to be clear, the existence and values of the variable "min_dt" **are** the issue.

3. The concept of a CFL condition in general and more so the possibility of deriving a generally applicable value for a given type of partial differential equation in combination with an applied numerical time stepping scheme seems to have escaped the attention of the OGGM developers:
https://github.com/OGGM/oggm/blob/1dc07f57cf8f7d7708585a6f7ab972890924892e/oggm/core/flowline.py#L936-L938
[...]

PR #910, opened 2019-11-23
Get ready for numerical tests

fmaussion:
A follow up to https://github.com/OGGM/oggm/issues/909 

With these tools, we can now compare various numerical schemes against each other in realistic settings.

For Alex's scheme, we need to make all bed shapes rectangular (which is OK), but we also need to make all sections the same widths (this is a bit more annoying) -> @alexjarosch , could your scheme accept varying widths along the flowline?

### DIFF
--- a/oggm/core/flowline.py
+++ b/oggm/core/flowline.py
@@ -909,7 +909,8 @@ class FluxBasedModel(FlowlineModel):
     def __init__(self, flowlines, mb_model=None, y0=0., glen_a=None,
                  fs=0., inplace=False, fixed_dt=None, cfl_number=0.05,
                  min_dt=1*SEC_IN_HOUR, max_dt=10*SEC_IN_DAY,
-                 time_stepping='user', flux_limiter=False, **kwargs):
+                 time_stepping='user', flux_limiter=False,
+                 raise_on_cfl=False, **kwargs):
         """Instanciate the model.
 
         Parameters
@@ -993,6 +994,7 @@ class FluxBasedModel(FlowlineModel):
         self.cfl_number = cfl_number
         self.calving_m3_since_y0 = 0.  # total calving since time y0
         self.flux_limiter = flux_limiter
+        self.raise_on_cfl = raise_on_cfl
 
         # Do we want to use shape factors?
         self.sf_func = None
@@ -1127,6 +1129,8 @@ class FluxBasedModel(FlowlineModel):
 
         # Time step
         self.dt_warning = dt < min_dt
+        if self.raise_on_cfl and self.dt_warning:
+            raise RuntimeError('CFL error: required time step too small.')
         dt = utils.clip_scalar(dt, min_dt, self.max_dt)
 
         # A second loop for the mass exchange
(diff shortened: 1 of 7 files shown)


fmaussion:
(I'm not going to merge this btw, this is just to get the ball rolling).

alexjarosch:
> For Alex's scheme, we need to make all bed shapes rectangular (which is OK), but we also need to make all sections the same widths (this is a bit more annoying) -> @alexjarosch , could your scheme accept varying widths along the flowline?

My scheme, as all diffusion type flowline schemes computes ice fluxes (vertically integrated velocities and knows flux gate geometries) on grid cell boundaries along the flow line geometry. You can of course recover the actual average velocity on each grid cell boundary and use this to estimate a total ice flux across a flow perpendicular cross section with methods similar to the ones OGGM uses already.

However be aware that flowline models like my scheme only see the drag (or better shear stress) from the basal boundary, they assume laterally infinite domains. Your scheme ignores the lateral shear stresses due to side walls as well if I remember correctly. This will make a comparison more difficult.

Something we should discuss in more detail.

fmaussion:
Could we put the width in the diffusion equation as Oerlemans 97 does? 

> Your scheme ignores the lateral shear stresses due to side walls as well if I remember correctly. 

In default configuration yes, but Philipp implemented a parameterization after [Adhikari and Marshall](https://www.cambridge.org/core/journals/journal-of-glaciology/article/parameterization-of-lateral-drag-in-flowline-models-of-glacier-dynamics/9C5836F192CE16F10F08DE750B1BFCAE). But people usually don't turn this on.

> Something we should discuss in more detail.

I'd be glad to chat shortly about this yes

alexjarosch:
> Could we put the width in the diffusion equation as Oerlemans 97 does?
 
Yes you can. However if it actually will make sense is a **serious** question here.

The misunderstanding lies in what the variable "S" means in OGGM and in the MUSCL-Superbee scheme.

- For OGGM "S" is the cross sectional area of the whole glacier, following Oerlemans ideas...
- In MUSCL-Superbee (as in all diffusion type SIA models), "S" or "s" as I wrote in my paper is the surface elevation. This stems from the fact that the mass conservation equation (e.g. [OGGM docs, frist equation](https://docs.oggm.org/en/latest/ice-dynamics.html) is valid for a "control volume", in which "U" (the velocity) then gets integrated vertically and so... (Derivation of the sallow ice equations, or shallow water equations for that matter). Actually the derivation is for &part; h / &part; t and one can make &part; s/ &part; t out of it by assuming a constant bed elevation b and h = s-b.

Now if you use just one control volume across the glacier, as OGGM and Oerlemans does/did, you run into issues with the side walls, as they create boundary conditions. Given one control volume across the glacier for this case means that you want to calculate on one point the average velocity "u" (in the OGGM sense) where the lateral side walls of that control volume are at u=0 (or prescribed sliding velocity), however the average should be different from u=0. You see the issue here, right?

So, sparing us a long discussion on the numerics of this setup, I would suggest you go the other way for a comparison. Why not recover the central flow line ice flux or centerline vertical ice velocity from your "flux-based" scheme and then compare to the MUSCLSuperbee. This way you skip the parametrization of any lateral boundary conditions and actually compare two results that are comparable. 
 
> I'd be glad to chat shortly about this yes

You let me know.

Issue #791, opened 2019-05-30
Refactor the test suite to use pytest instead of UnitTest


fmaussion:
Out of consistency (and because this is all I knew when we started), all our tests use the UnitTest framework (classes with test methods). 

This is outdated and inefficient. In the process of a major test refactoring our tests could be simplified and moved over to a pure functional pytest style. This is going to be quite a lot of work for not much reward, but should happen one day...

PR #922, opened 2020-01-05
Refactor tests

C-Merrill:

- [ ] Tests added/passed
- [ ] Fully documented

### DIFF
--- a/oggm/tests/test_models.py
+++ b/oggm/tests/test_models.py
@@ -420,8 +420,8 @@ def test_past_mb_model(hef_gdir):
     assert_allclose(mb, mb_gw)
 
 
-def test_glacierwide_mb_model(hef_gdir):
-
+@pytest.mark.parametrize("cl", [massbalance.PastMassBalance, massbalance.ConstantMassBalance, massbalance.RandomMassBalance])
+def test_glacierwide_mb_model(hef_gdir, cl):
     gdir = hef_gdir
     init_present_time_glacier(gdir)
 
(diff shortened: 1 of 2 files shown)


C-Merrill:
I'm new to the industry (scientific python libraries), but thought this project looked interesting. Thinking a close look at the tests would give me a good idea of the project structure, I decided to do some work related to #791. Looking for some feedback at this stage to see if this is moving in the right direction.

Next steps would be finishing up whatever test parametrization is left in test_models.py, then doing a similar flow for the other test modules. I also have some ideas about pulling the `dummy_` functions out of `funcs.py` and making them cached generator fixtures in `conftest.py` if we think that's a significant save.

fmaussion:
Wow, thanks! This is definitely a surprise, such a significant contribution out of nowhere like this - I'm very grateful!

> Looking for some feedback at this stage to see if this is moving in the right direction.

I believe this is what I had in mind when I suggested to refactor the tests - and this PR probably already goes beyond that, as you are using tools I wasn't yet aware of. As you've probably noticed by now, we are scientists by training and no programmers, and some parts of OGGM are messy.

There are a couple of things to keep in mind in this endeavor:
- we obviously have to keep or increase code coverage, and for some obscure reason that only coveralls might know, the test coverage did not run for this PR (nothing you can do about it, but annoying)
- regardless of coverage, I'll have to make a thorough review to check if the tests continue to test what they intended to. After a quick look at your code I'm not really worried, but it'll still be necessary and might take some time on my side
- we'll have to be careful in not being *too* fancy - although you might find the new structure easier, I'm concerned about present and future contributors who might get lost with pytest shenanigans - I'll assess this point myself when I'll do the review (I'm not too concerned here either)
- finally (and this is important):  `test_models` is among the "nicely coded" tests of the test suite. Other tests (mostly: `test_prepro` and `test_workflow`) are a mess of incremental tests with a lot of obsolete duplication. In the process of https://github.com/OGGM/oggm/issues/791, I also aimed to target some refactoring of the tests themselves. Here, I wonder what's the best strategy (i.e. can someone else than me assess why the tests were written this way and how to simplify them?

In all cases, I would like to go forward with this as much as your time allows yes!

Out of curiosity, what brought you to us? There are plenty of other scientific packages out there!?

C-Merrill:
Glad to hear this is more or less what was intended! I had a lot of fun doing it and learned a lot about pytest myself. As an introduction, I'm a software engineer at a company that delivers a self-serve productivity tool on the web called Lucidchart. As great as that has been, I'm looking to use the skills I've gained to develop technology that I would find more fulfilling. So, I went looking on github for some environmental science projects to contribute to and found this knowledge repository: https://github.com/pangeo-data/awesome-open-climate-science. You were the sole entry under Glaciology and I think glaciers are really interesting. So that's the story of me finding your project.

As far as the work here, I appreciate all the feedback. Ease of use was certainly on my mind through a lot of it, which is why I'm a little unhappy (as stated in the commit message) with one of the commits here. I think I went a little overboard in the name of parametrization and just made things unnecessarily messy. That case aside, though, I think that there is a lot of power to pytest fixtures that you should take advantage of. We use dependency injection pretty heavily where I work and I actually think it makes development easier as long as people are comfortable with it (it kind of seems like black magic at first). And that's to say nothing of the performance implications of reusing the same injected test resource rather than constructing a new one every time.

It's funny you bring up the final point there, because `test_models.py` was not the first module I looked at. I chose it because the conversion looked to be a bit easier than others I looked at first and I was still getting a feel for pytest best practices. I definitely have time and would be willing to start looking at those other more troublesome modules, starting with the auditing you're proposing.

Finally, not sure why things seem to not be running. I'm not totally familiar with the tools you're using, but I suppose it's possible back-to-pack force pushes of amended commits for formatting may have tripped Travis up or something like that. I'll try to look into it.

fmaussion:
@C-Merrill thanks for the info! And welcome to the team (write me an email if you'd like to have access to our slack where some discussion happens as well).

Is this PR ready for review? I'll give it a go over the week-end if this is the case

C-Merrill:
@fmaussion This could/probably should be reviewed now. I do need to make one quick change to add back part of a test I parametrized, but once that's done, it should be ready. Won't close out the issue since there's still a lot more to do.

Also, as I'm writing this, I'm wondering if I should add some inline documentation to some of the fixtures. I think I'll start working on that and addressing what I've already made a not of in the diff, but you should be fine reviewing it if I don't get that done by the weekend.

Email is cmerrill99@gmail.com  I'd love too join your slack

oggm/tests/test_models.py
```diff
@@ -49,13 +48,12 @@
 pytest.importorskip('salem')
 
 
-def setup_module(module):
-    module._url_retrieve = utils.oggm_urlretrieve
+@pytest.fixture(autouse=True, scope='module')
+def init_url_retrieve(request):
```
fmaussion: Whether this is really much better than what we had before is a bit of matter of taste. The previous (with setup and teardown) is a bit more explicit on the first sight. 

Let's keep your version, but rename to `patch_url_retrieve` and add a comment specifying that the "teardown" happens after the yield.

oggm/tests/test_models.py
```diff
@@ -64,406 +62,434 @@ def teardown_module(module):
...
-        utils.mkdir(self.testdir, reset=True)
-        self.gdir = tasks.copy_to_basedir(gdir, base_dir=self.testdir,
-                                          setup='all')
-
-    def tearDown(self):
-        self.rm_dir()
+@pytest.fixture(scope='module')
+def io_model_factory():
+    glen_a = 2.4e-24
+    return lambda fls, **kwargs: FluxBasedModel(fls, glen_a=glen_a, **kwargs)
```
fmaussion: Is this fixture *really* necessary?

oggm/tests/test_models.py
```diff
@@ -64,406 +62,434 @@ def teardown_module(module):
...
-    def rm_dir(self):
-        if os.path.exists(self.testdir):
-            shutil.rmtree(self.testdir)
 
-    def test_init_present_time_glacier(self):
-
-        gdir = self.gdir
-        init_present_time_glacier(gdir)
+@pytest.fixture(scope='module')
+def backwards_idealized_fls(backwards_model_factory):
```
fmaussion: I think we should remove the backwards tests altogether

oggm/tests/test_models.py
C-Merrill: While it feels essentially the same as using the `setup_module` and `teardown_module` calls, those are still very much xunit-style ways of doing things. One of the nicest things of doing this as a fixture is we only need to write it once and every test module can use it.

oggm/tests/test_models.py
C-Merrill: It really isn't. This is an artifact of a straight conversion from xunit class setups to analogous fixtures. Ideally, fixtures would define a certain test resource, rather than an entire setup specific to certain tests. This should be readdressed.

C-Merrill:
@fmaussion Thank you for taking a look at this. You've brought up some excellent points and concerns, some of which addressing questions I had in doing this work. It seems the biggest concern is: does this try to be too fancy, jeopardizing readability and maintainability. I agree that it does. I wasn't sure to what level you wanted refactoring done, so if you go commit by commit, you can see me adding incrementally more "fanciness" and part of what I was hoping out of a review is: what parts of this is necessary?

As far as a big picture view, I agree that it would be very tricky and take a lot of time for me to get the level of familiarity necessary to make decisions there. And being new to the project, there's a certain level of credibility that I would need to make those decisions that would be impossible to gain quickly. So, it would absolutely be best if you or one of the more experienced contributors could spearhead that effort. I understand time constraints and am in no rush to push this through in any way.

Responding to your itemized questions/suggestions:

1. Re-evalutating and restructuring the fixtures considering what is common enough functionality to be necessary as a fixture and what can be test or test-class specific setup/teardown is work that should be done. I'll start on that right away.

2. I'll add documentation to the fixtures as well. Already intended to, so that less-familiar contributors can still navigate the test suites. If you would like, I can also add a write-up to the testing section of the contributors page of the project explaining how we use pytest and best practices.

3. This is a question I intended to bring up earlier in the review process and forgot. It is permissable in the pytest framework to group tests in classes. It's a little strange because the class offers little outside of structural sectioning of tests (any class with `__init__` defined is skipped, so class setup and teardown is done by fixtures), but doing so allows for test case naming as you mentioned. It also allows for class-scoped fixtures so that a resource could be reconstructed anew for each class (and potentially differently based on static class properties). I removed the test classes thinking that that may have been part of what was intended on the issue description, but they can easily be added back.

I'll never take PR feedback as negativity, so no concern there. I fully expected to receive such a response and am glad we're ironing this out. Let's just always make sure we're on the same page!

fmaussion:
@C-Merrill , sorry I forgot to comment on this - I fully agree with all your suggestions, including the organization in classes - I know it is a bit silly, but the structure still makes sense (testing mass-balance is not the same as testing the numerical model). Also, in PyCharm, you can test either an individual unit test, a test class, of the entire module: it is a nice organizational feature I believe.

Also, you can safely remove the `Backwards...` test class.

Do you think you can have another go at your PR in the next 7 days? I have a PR open which will change some tests as well (https://github.com/OGGM/oggm/pull/931) and I'd like to merge yours first.

C-Merrill:
I'll have a revised version of this by the weekend

fmaussion:
superseded by https://github.com/OGGM/oggm/pull/934

Issue #860, opened 2019-08-27
Default time stepping parameters maybe too ambitious?


fmaussion:
In certain cases (a handfull of glaciers for very large growing glacier scenarios in the Alps, cc @juliaeis ), the simulations can become unstable.

Here is a script to reproduce the issue:

```python

def test_constant():

    rid = 'RGI60-11.00280' # RGI60-11.01450  RGI60-11.00852

    WORKING_DIR = '/home/mowglie/Downloads/constant_climate_error/wd/'

    cfg.initialize()
    cfg.PATHS['working_dir'] = WORKING_DIR
    utils.mkdir(WORKING_DIR, reset=True)

    cfg.PATHS['plot_dir'] = os.path.join(cfg.PATHS['working_dir'], 'plots')
    utils.mkdir(cfg.PATHS['plot_dir'], reset=False)

    # Use multiprocessing?
    cfg.PARAMS['use_multiprocessing'] = False

    # How many grid points around the glacier?
    cfg.PARAMS['border'] = 200

    # Set to True for operational runs
    cfg.PARAMS['continue_on_error'] = False

    # Use HISTALP climate file
    cfg.PARAMS['baseline_climate'] = 'HISTALP'
    cfg.PARAMS['prcp_scaling_factor'] = 1.75
    cfg.PARAMS['temp_all_liq'] = 2.0
    cfg.PARAMS['temp_default_gradient'] = -0.0065
    cfg.PARAMS['temp_melt'] = -1.75
    cfg.PARAMS['temp_all_solid'] = 0.0

    # We use intersects
    db = utils.get_rgi_intersects_region_file(version='61', region='11')
    cfg.set_intersects_db(db)

    # RGI file
    path = utils.get_rgi_region_file('11', version='61')
    rgidf = gpd.read_file(path)

    # import pd.DataFrame with experiement FileModels

    rgidf = rgidf[rgidf.RGIId.isin([rid])]
    gdirs = workflow.init_glacier_regions(rgidf)

    workflow.gis_prepro_tasks(gdirs)
    workflow.climate_tasks(gdirs)
    workflow.inversion_tasks(gdirs)
    workflow.execute_entity_task(tasks.init_present_time_glacier, gdirs)

    gdir = gdirs[0]

    v0 = tasks.run_random_climate(gdir, nyears=600, y0=1850, seed=1,
                                  temperature_bias=-1,
                                  store_monthly_step=False,
                                  output_filesuffix='_start')

[...]

PR #931, opened 2020-01-17
Fix OGGM's time stepping scheme

fmaussion:
Closes https://github.com/OGGM/oggm/issues/909

- [x] Tests added/passed
- [x] Fully documented

This is an implementation of the ideas discussed in https://github.com/OGGM/oggm/issues/909 and https://github.com/OGGM/oggm/pull/910 

I still have to do a couple more tests (regarding accuracy and preformance) and write a report before we can merge this.


fmaussion:
Also:
- closes https://github.com/OGGM/oggm/issues/860 
- closes https://github.com/OGGM/oggm/pull/910

PR #933, opened 2020-01-21
Use spawn-multiprocessing

TimoRoth:
This avoids issues with pyproj corrupting its database due to passing
the already opened handle to forked child processes, which then modify
it in parallel.


fmaussion:
Thanks! Looks like there is going to be some performance drop but I don't think this is relevant here...

PR #934, opened 2020-01-25
Refactor tests, revised

C-Merrill:

- [ ] Tests added/passed
- [ ] Fully documented


C-Merrill:
@fmaussion I walked back a lot of what I did in the previous pull request, hoping that in aligned more with what you're intending. There are still some class-specific fixtures that seem a little funky, but I've left them close to the class they are used in to have a sort of class setup fixture before each class.

I'm willing to work on those more if you want some of it cleaned up, but I've run out of time tonight and wanted to get you some followup work from our last conversation.

Cheers,
Chris

fmaussion:
Thanks @C-Merrill ! It looks great - I've made a quick check and it looks good. Indeed, the current mix of classes and pytest is a bit funky, but I guess this is the price to pay for having started with Unittest in the first place. New test modules (or more ambitions refactoring endeavors) should become pytest only.

I've learned a lot in the process - also seeing that rewriting the tests to be more efficient (if this ever happens one day) is going to be hard...

C-Merrill:
@fmaussion you shouldn't feel too bad about the current state of that test suite. It seems that classes in a pytest module is a perfectly normal and correct way to group tests. Were it not so, they would not have implemented class scope for pytest fixtures. There is a little weirdness in some of the class-specific fixtures, but it's not necessarily bad, and as this moves forward, we may start to see patterns in those fixtures that would help us realize what the common resource would be. It's really difficult to determine what's common after only doing one of these.

It's not an insurmountable task by any means. If you'd like me to start on another of the test modules, just let me know which, and I'll get started.

Issue #936, opened 2020-02-02
New multiproc causes MyBinder to freeze


fmaussion:
I definitely don't know why, but the new way we initialize the mp pool causes machines on Mybinder to freeze.

To reproduce, open [this link](https://mybinder.org/v2/gh/OGGM/binder/master?urlpath=git-pull?repo=https://github.com/OGGM/oggm-edu-notebooks%26amp%3Bbranch=master%26amp%3Burlpath=lab/tree/oggm-edu-notebooks/oggm-tuto/welcome.ipynb%3Fautodecode), open a notebook or a terminal, and run:

```python
from oggm.workflow import init_mp_pool 
from oggm import cfg 
cfg.initialize()  
init_mp_pool()
```

cc @TimoRoth

PR #937, opened 2020-02-03
Use environment variable to override multiproc

fmaussion:
Closes https://github.com/OGGM/oggm/issues/936

- [x] Tests added/passed
- [ ] Fully documented

### DIFF
--- a/oggm/workflow.py
+++ b/oggm/workflow.py
@@ -5,7 +5,7 @@ import os
 from shutil import rmtree
 from collections.abc import Sequence
 # External libs
-import multiprocessing as mp
+import multiprocessing
 import numpy as np
 
 # Locals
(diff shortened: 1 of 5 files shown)


oggm/workflow.py
```diff
@@ -147,7 +147,7 @@ def execute_entity_task(task, gdirs, **kwargs):
         if ogmpi.OGGM_MPI_COMM is not None:
             return ogmpi.mpi_master_spin_tasks(pc, gdirs)
 
-    if cfg.PARAMS['use_multiprocessing']:
+    if cfg.PARAMS['use_multiprocessing'] and cfg.PARAMS['mp_processes'] > 1:
```
TimoRoth: The issue here is that there are two other ways it could be 1.
Once if the system actually only has a single CPU, and one if the slurm env var is 1.

fmaussion: > Once if the system actually only has a single CPU, and one if the slurm env var is 1.

But in this cases, we also would like to turn off multiprocessing altogether right? I thought this was the point of your comment on slack yesterday

TimoRoth: Yes, that's what I'm saying.
If the CPU count is set to 1, there is no reason to even bother with multiprocessing.
And mp_processes is not the only way it can be set to 1.

fmaussion:
I have now moved the logic to cfg.initialize, so that mp_processes is decided at the very beginning. 

One question though: do you know how many CPUs will be given by `cpu_count` on Travis?

fmaussion:
I can't see from the travis log if multiproc was used or not, I think I need to add a check to the tests or something...

fmaussion:
OK I removed the "less than 1 process" magic, I think users should simply decide whether to use multiprocessing or not. 

I have however removed the spawn context per default, and added an env variable to switch it on.

@TimoRoth thoughts?

TimoRoth:
Looks good to me. I will see if I can somehow hot-patch this in the Anaconda-Path of things, because on Anaconda this is for some reason way more likely to cause problems.

PR #938, opened 2020-02-04
The dynamical model now shares glacier flowlines state with MB model

fmaussion:

- [ ] Tests added/passed
- [ ] Fully documented

### DIFF
--- a/oggm/core/flowline.py
+++ b/oggm/core/flowline.py
@@ -571,7 +571,7 @@ class FlowlineModel(object):
     def length_m(self):
         return self.fls[-1].length_m
 
-    def get_mb(self, heights, year=None, fl_id=None):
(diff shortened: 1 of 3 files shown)


oggm/core/flowline.py
```diff
@@ -983,6 +987,7 @@ def __init__(self, flowlines, mb_model=None, y0=0., glen_a=None,
         self.shapefac_stag = []
         self.flux_stag = []
         self.trib_flux = []
+        self.mb_on_surfaceh = []
```
matthiasdusch: that is never used. was that intended for `mbs` in line 1014?

fmaussion: Thanks! I overlooked this yes, needs removal

Issue #857, opened 2019-08-16
Add a flux gate to numerical model


fmaussion:
I did so for some test case in the WIP calving module (https://github.com/ehultee/chakra), and it allows quite nice and clean mass conservation checks and sensitivity experiments.

PR #944, opened 2020-02-08
Add flux gate to numerical model for mass-conservation tests

fmaussion:
Closes https://github.com/OGGM/oggm/issues/857

- [ ] Tests added/passed
- [ ] Fully documented

cc @drounce


fmaussion:
One nice side effect of this tool: one can check mass-conservation of the dynamical model at the floating point precision.

Current checks are done for simple beds and for Hintereisferner, and mass-conservation is perfect (yay!). However, we slightly loose mass via the simple calving param (not much but still), but I decided not to care too much as we'll try another calving param soon.

fmaussion:
Correction: I discoverd a small bug in "calving", this is now also 100% mass-conservative

PR #957, opened 2020-02-21
Add support for geopandas >0.7 and latest salem

fmaussion:
Once available, we will have to test on both an older and a new image with newest salem and geopandas in parallel for a while.

Currently the latest image is from 6 days ago, I'll wait a little.


fmaussion:
This also adds a better check if the glacier directory exists or not, so that one can start from a DEM gdir

PR #961, opened 2020-02-25
add Copernicus DEM

matthiasdusch:
Closes https://github.com/GLIMS-RGI/rgitools/issues/35
work in progress, right now my account is locked for entering a wrong password.

unfortunately this will increase the complexity of the download module even more. Copernicus not only uses FTP**S** for their data, they use the apparently deprecated *implicit* FTPS...

Also they don't name their files in a meaningful way so this will need a lookup table of some sorts.

And the download is rather slow. Not the actual download but something along the way, SSL handshake or such...

- [x] Tests added/passed
- [x] Fully documented
- [x] update `SAMPLE_DATA_COMMIT` if lookup table ends up there

### DIFF
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -17,6 +17,12 @@ Breaking changes
 Enhancements
 ~~~~~~~~~~~~
 
+- Added Copernicus DEM GLO-90 as optional DEM. Requires credentials to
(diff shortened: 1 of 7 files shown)


fmaussion:
this `_downloads.py` is becoming a piece of abstract art :heart_eyes: 

Thanks for looking into this and let's hope they have a way to find the files with a machine

PR #971, opened 2020-03-05
moar dems! this time: NASADEM

matthiasdusch:
- [x] Tests added/passed
 
There is a new SRTM in town:
https://lpdaac.usgs.gov/products/nasadem_hgtv001/
This brings higher resolution (30m) and other (for us probably more relevant) improvements like improved geolocation accuracy and usage of other sources for void filling.


fmaussion:
Thanks!

Issue #893, opened 2019-11-05
Nonexistent TanDEM-X files are "downloadable" without credentials


matthiasdusch:
If a nonexistent TanDEM-X tile url is given to `requests` without credentials (which we do in OGGM to check the download cache first) it returns a valid status_code and a local file gets created. Beside the hash-warning this local file is now treated as valid.

Minimal example:
https://nbviewer.jupyter.org/gist/matthiasdusch/0b0444ee0b0ec743c466f73cb7f526ec

A first serious error is raised when trying to unpack this file and the respective glacier will fail, even if this particular tile is not necessary. A glacier where this happens is `RGI60-01.12886` and the reason why the tile is not necessary is that at this latitude there are two lons/lats in one tile. So W144 and W146 are required but we also try to download W145 which does not exist.

Possible ways to fix:
- Be more precise at the zone selection and do not select W145 in the first place: Quite error prone as we would rather download 1-2 unnecessary tiles than skipping existing ones.
- Test files for size or valid zip format in the TanDEMx download function and return None in that case.
- Catch this in `_requests_urlretrieve` and raise a HttpDownloadError. I'd say the best solution but not really sure how to do this without breaking other downloads. Maybe @TimoRoth can help here.

https://github.com/OGGM/oggm/blob/91b878b7119248787a2e1bede481b9fce510ab22/oggm/utils/_downloads.py#L356-L384

Line 368: The fallback to `size=-1` if no content-length is given is necessary I guess?

TimoRoth:
There's a lot of files which do not give a proper content-length. Without setting it to -1, downloading them would fail.
I'm not sure how to properly catch this in the download function. If the server returns a 200 status code, and some data, really everything is OK for what it's concerned.
It's primarily an issue of the server to behave like this in this case.

As a workaround I'd say checking the file for an obviously too small size for that specific dataset is the cleanest way.

fmaussion:
> Test files for size or valid zip format in the TanDEMx download function and return None in that case.

if this is possible yes -> maybe we should do this for all other datasets as well if this is somehow at the same location in code? (I don't have the overview right now)

TimoRoth:
On 05.11.2019 17:39, Fabien Maussion wrote:
>     Test files for size or valid zip format in the TanDEMx download
>     function and return None in that case.
> 
> if this is possible yes -> maybe we should do this for all other 
> datasets as well if this is somehow at the same location in code? (I 
> don't have the overview right now)

[...]

PR #976, opened 2020-03-16
avoid downloads without credentials

matthiasdusch:
the DLR server had this annoying habit to return a small/broken file to the OGGM download function even if it did not exist. See #893  
But *I think* until now this only happened if authentication credentials were provided.

Now this also happens to existing files if the download function is called without credentials which we did to check the cache first. So all files not in the cache just returned broken.

Fixed this by not calling the entire file downloader but only parts of it. This should check the cache and verify the file with the hash. But it avoids trying to establish a connection without the credentials by using a dummy function.

Maybe @TimoRoth can have a look if I missed something. 

I'll add 1-2 tests before merging

### DIFF
--- a/oggm/utils/_downloads.py
+++ b/oggm/utils/_downloads.py
@@ -657,7 +657,7 @@ def file_downloader(www_path, retry_max=5, cache_name=None,
     return local_path
 
 
-def download_with_authentification(wwwfile, key):
+def download_with_authentication(wwwfile, key):
     """ Uses credentials from a local .netrc file to download files
 
     This is function is currently used for TanDEM-X and ASTER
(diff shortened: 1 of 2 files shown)


matthiasdusch:
The already existing `test_tandemdownload` should and does fail when run without these commits here. So pretty sure something must have changed on DLR's side. Not sure when though as download tests aren't run regularly as far as I know.

Still added another test which is supposed to fail.

fmaussion:
>  download tests aren't run regularly as far as I know.

the download tests are basically never run, unless someone does so on a local machine.

oggm/utils/_downloads.py
```diff
@@ -657,7 +657,7 @@ def file_downloader(www_path, retry_max=5, cache_name=None,
     return local_path
 
 
-def download_with_authentification(wwwfile, key):
+def download_with_authentication(wwwfile, key):
```
fmaussion: "f" is missing here ;-)

fmaussion: ah no: "fi"

matthiasdusch: well, I decided to go with the correct spelling :-)
https://english.stackexchange.com/questions/7844/is-authentification-a-real-word

fmaussion: Ha! Learned something new, but beh, "authentication" is ugly...

PR #977, opened 2020-03-19
increase LRU file handler size

matthiasdusch:
I noticed some of the DEM runs to fail because they couldn't find/read the downloaded and unpacked .tif file. e.g.:
`RasterioIOError occurred during task define_glacier_region on RGI60-07.00608: /work/mdusch/91650/oggm_tmp/tmp/Copernicus_DSM_30_N79_00_E023_00_DEM.tif: No such file or directory`

This happens only for DEMs with smaller tiles (like the 1x1 degree of Copernicus and TanDEM-X) and probably only on a cluster where there multiple glaciers processed at the same time: Unpacked tiles in the temp directory will be deleted by the file handler before they could actually be used by the define_glacier task...

So far I just checked what happens if I increase the maxsize within the Copernicus download function. And it works/runs without the above error.

But what's the best solution? 
i) Introduce a cfg.PARAMS variable 
ii) Increase the maxsize for the DEM sources where this might happen?
iii) ?

Happy to get some feedback by @TimoRoth and @fmaussion 
Otherwise I will implement variant i)

### DIFF
--- a/oggm/params.cfg
+++ b/oggm/params.cfg
@@ -59,7 +59,7 @@ mpi_recv_buf_size = 131072
 dl_verify = True
 
 # Default number of files to be cached in the temporary directory
-LRU_maxsize = 100
+lru_maxsize = 100
 
 ### CENTERLINE determination
 
--- a/oggm/params.cfg
+++ b/oggm/params.cfg
@@ -58,6 +58,9 @@ mpi_recv_buf_size = 131072
 # Check for the integrity of the files OGGM downloads at run time
 dl_verify = True
 
+# Default number of files to be cached in the temporary directory
+LRU_maxsize = 100
+
 ### CENTERLINE determination
 
 # Decision on grid spatial resolution for each glacier
--- a/oggm/cfg.py
+++ b/oggm/cfg.py
@@ -365,6 +365,15 @@ def initialize_minimal(file=None, logging_level='INFO'):
                      'processors (N={})'.format(mpp))
     PARAMS['mp_processes'] = mpp
 
+    # Size of LRU cache
+    try:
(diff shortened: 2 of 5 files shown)


fmaussion:
yes, variant i) is fine. An ENV variable would also do it, usually we have a combination of both (the ENV variable overrides whatever cfg.PARAMS says

matthiasdusch:
Is there a reason to use directory **and** maxsize as key to the LRU handler dict?
Directory *and* ending seems like a more likely usecase.

https://github.com/OGGM/oggm/blob/3eeaf6e5d9645effb6a1ff6959d83745beeff526/oggm/cfg.py#L580-L615

fmaussion:
Yes, it looks like you are right! Thanks!

oggm/params.cfg
```diff
@@ -58,6 +58,9 @@ mpi_recv_buf_size = 131072
 # Check for the integrity of the files OGGM downloads at run time
 dl_verify = True
 
+# Default number of files to be cached in the temporary directory
+LRU_maxsize = 100
```
fmaussion: Can you change this to all low-caps? Also in cfg.PARAMS - I think we only have low-caps in there

matthiasdusch:
> Yes, it looks like you are right! Thanks!

While implementing I found a usecase for having the maxsize as part of the key: It allows to in-/decrease the LRU cache during runtime. Not sure if anybody will ever do this, but an existing test was covering this. So I thought I keep this possibility but still removed it from the dict-key.

fmaussion:
yes, I'm not even sure we are using more than one handler tbh :thinking:

fmaussion:
Got hit by that recently, took me quite some time to figure out. Increasing to 1000 did help - maybe we should change the default?

PR #979, opened 2020-03-20
Optimize hash verify file IO

fmaussion:
On my machine, reading the downloads.sha256.xz into a dict takes about 11-12 seconds. Since we are now reading the hash file in init_glacier_regions (i.e. to avoid a read per process), this is just too long because the simplest test script will read it as well. 

This is a quick hack to store the data in an intermediate binary file the first time it is read. This means that the first call on a new machine will trigger the download of the file and the creation of the intermediate file (ie slilghtly longer than the 12 secs mentioned above), but < 2 secs on all subsequent ones. 

This is already better than what we have, but I still think that the proper solution would be to not have to read the data unless we really want to download something, and this requires communication between processes. I might give a go at th e Manager object this week-end...

- [ ] Tests added/passed
- [ ] Fully documented

### DIFF
--- a/oggm/tests/test_utils.py
+++ b/oggm/tests/test_utils.py
@@ -1202,7 +1202,8 @@ class TestFakeDownloads(unittest.TestCase):
         file_sha256 = file_sha256.digest()
 
         data = utils.get_dl_verify_data()
-        data['test.com/test.txt'] = (file_size, file_sha256)
+        data.loc['test.com/test.txt', 'size'] = file_size
+        data.loc['test.com/test.txt', 'sha256'] = file_sha256.hex()
 
         return 'https://test.com/test.txt'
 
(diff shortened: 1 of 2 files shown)


fmaussion:
cc @TimoRoth @matthiasdusch

TimoRoth:
I can also easily generate the hash file in this format on the server side in the first place. That should simplify this code a lot.

fmaussion:
> I can also easily generate the hash file in this format on the server side in the first place.

Yes, but wait before doing this. We need some better strategy altogether first. I've made some tests and I am in the process of writing this down.

TimoRoth:
This seems obsolete now?

fmaussion:
Yes! Will optimize a couple of other things though

PR #982, opened 2020-03-23
Switch download verification to hdf file format

TimoRoth:
This makes use of the new HDF file format.
It is, as of yet, untested. So don't merge yet.

### DIFF
--- a/oggm/workflow.py
+++ b/oggm/workflow.py
@@ -283,6 +283,9 @@ def init_glacier_regions(rgidf=None, *, reset=False, force=False,
         if os.path.exists(fpath):
             rmtree(fpath)
 
+    # Read the hash dictionary before we use multiproc
+    utils.get_dl_verify_data('cluster.klima.uni-bremen.de')
+
     gdirs = []
     new_gdirs = []
     if rgidf is None:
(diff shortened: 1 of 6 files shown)


TimoRoth:
The one remaining failing test is unrelated to the download changes, and is getting uncovered by the base image update.

fmaussion:
It's just a numerical issue. Change the failing line from

```np.testing.assert_allclose(df.mu_star.iloc[-1], 0)```

to

```np.testing.assert_allclose(df.mu_star.iloc[-1], 0, atol=1e-7)```

File: oggm/workflow.py [hunk 1]
```diff
@@ -283,6 +283,9 @@ def init_glacier_regions(rgidf=None, *, reset=False, force=False,
         if os.path.exists(fpath):
             rmtree(fpath)
 
+    # Read the hash dictionary before we use multiproc
+    utils.get_dl_verify_data('cluster.klima.uni-bremen.de')
```
fmaussion: I'm not 100% sure yet, but it's likely I'll remove all caching of the hash file(s) altogether: i.e. each potential verify will trigger a new read. The reason it that I'd like to avoid memory pickling too much data around. What do you think?

fmaussion: You can leave it as is for now: we can discuss on slack

fmaussion:
Travis seems to be stalled and did not report back to github. But on the travis website all is green. Thanks!

Issue #965, opened 2020-02-27
init_glacier_regions considered harmful


fmaussion:
See

https://github.com/OGGM/oggm/issues/964

https://github.com/OGGM/oggm/issues/754

https://github.com/OGGM/oggm/issues/366

Issue #964, opened 2020-02-27
run_prepro_levels is broken


matthiasdusch:
#957 broke `run_prepro_levels` with:
https://github.com/OGGM/oggm/blob/be4f1186ecb408e101e9a2aa96d7e560b1fa01ec/oggm/workflow.py#L331-L335

Outlines were kept between different DEMs and now there is no new GDir initiated.

Probably need to delete the outlines between every DEM

fmaussion:
I'm not very happy with #957 either.

I did it so because it allows to start from a DEM prepro dir, and then re-open the gdirs with `init_glacier_regions` without triggering a DEM generation. 

Now I believe that the *true* problem is that `init_glacier_regions` does two things: 
- preparing the glacier directories
- processing the DEMs (call `define_glacier_region` on the directories that are not OK)

It was a design error. In hindcast, #957 was probably  a design error since `init_glacier_regions` should actually deliver gdirs with proper DEMs in them...

The best way forward would be to rethink the `init_glacier_regions` logic, but it will be very hard to do so in a backwards compatible manner...

fmaussion:
> Probably need to delete the outlines between every DEM

This, or bypass `init_glacier_regions` entirely. It's only a wrapper anyway

PR #983, opened 2020-03-23
init_glacier_directories to replace init_glacier_regions

fmaussion:
Closes https://github.com/OGGM/oggm/issues/965


fmaussion:
I'm going to merge this now - without review. hopefully I didn't brake anything. @matthiasdusch since you are always on the frontline any feedback is appreciated

fmaussion:
Just some more info:
- init_glacier_regions is back to previous (and consistent) behavior before https://github.com/OGGM/oggm/issues/964
- I tested it in the middle of this PR to be sure I didn't brake anything, but init_glacier_regions is now only minimally tested and not for full functionality (it would have meant duplicating all the tests). So in the future it might get broken without noticing (which is not a biggie since it's supposed to be deprectated)
- so this shouldn't change anything for users in theory. Later we might start to send deprecation warnings (maybe in a couple of months)

PR #993, opened 2020-03-26
source argument for define_glacier_region

### LINKED ISSUE #992: Problems changing DEM_SOURCE
I am a bit late to the party, but #983 broke the prepro task.

I am working on a PR to fix right now. But I am confused about the following:
https://github.com/OGGM/oggm/blob/4436fa061d76b1b113412f220406fc89fdc93766/oggm/utils/_workflow.py#L1762-L1767

What is the reason to delete the DEM_SOURCE from the outlines?

Especially since it is later used to specify a custom DEM:

https://github.com/OGGM/oggm/blob/4436fa061d76b1b113412f220406fc89fdc93766/oggm/core/gis.py#L291-L293

matthiasdusch:
If using an alternative DEM it was common until now to declare this as an entry of the GeoPandasDataFrame.
Due to some recent changes (#983 ) it is necessary and also more appropriate to change this.
Following this PR one has to specify an alternative DEM source as parameter to `define_glacier_region`. 

Closes #992 

Unrelated: This PR also changes the sorting of the LRU file handler from `getmtime` to `getctime`. ctime is updated when the files are moved/extracted on the disc while mtime only shows the last modification which usually happened months/years ago at the original file (DEM) creation.

### DIFF
--- a/oggm/cfg.py
+++ b/oggm/cfg.py
@@ -680,7 +680,7 @@ def get_lru_handler(tmpdir=None, maxsize=None, ending='.tif'):
         from oggm.utils import LRUFileCache
         # the files already present have to be counted, too
         l0 = list(glob.glob(os.path.join(tmpdir, '*' + ending)))
-        l0.sort(key=os.path.getmtime)
+        l0.sort(key=os.path.getctime)
         lru = LRUFileCache(l0, maxsize=maxsize)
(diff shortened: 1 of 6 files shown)


fmaussion:
Thanks! I shortly considerer if `dem_source` is a better name than `source`, but then I though bah, whatev ;-) 

Thanks!

Issue #954, opened 2020-02-16
Some ideas / questions about the operational use of the calving param for global runs


fmaussion:
Once https://github.com/OGGM/oggm/pull/945 is merged, we will have a pet model for calving in OGGM. Here are some thoughts/ideas to make everything a bit more robust:

- **Water level and DEMs**. In https://github.com/OGGM/oggm/pull/945, we introduce a variable ("water_level") which doesn't necessarily have to be zero. With this, we can make lake terminating glaciers calve, requiring an assumption only about their free board. Similarly, we could use this water level to make arbitrary "corrections" to the DEMs. I see the following situations:
     - elevation of terminus grid point is too high (e.g. 250 m for columbia). In this case, we arbitrarily raise the water level until the free board is reasonable (what's reasonable?)
     - elevation of terminus grid point is too low (e.g. 0).  In this case, we arbitrarily raise the grid point or lower the water level (both are very similar, as the outcome will be that the slope at the terminus will be set to the minimal slop allowed in OGGM anyway.
     - the glacier is lake terminating. In this case we set the water level (free board) to an arbitrary value (what's reasonable?)

- **Inversion**. In Recinos et al 2019, the assumption "ice flux by deformation = ice flux by calving" is a good one. However, I think now that taking the slope of the last grid point only is a bit too noisy and arbitrary. We could use the same method, but average the slope (and maybe even to width) to the N last grid points of the flowine for more robust results. For example, 3 or 5?

Issue #869, opened 2019-08-31
Report volume loss happening below sea level


fmaussion:
I was thinking about this after doing some data analysis of [Farinotti et al., 2019](https://www.nature.com/articles/s41561-019-0300-3).

Currently the model looses mass via SMB everywhere, and via "calving" only in very special cases. It is actually not so trivial to know which mass actually contributes to SLR and not.

Also, currently the model doesn't care about sea level (i.e.: water) at all, i.e. SMB is still computed below 0 m a.s.l.

PR #996, opened 2020-03-31
Add sanity checks for water level of tidewater glaciers

fmaussion:
Closes https://github.com/OGGM/oggm/issues/954

This adds the sanity checks discussed in https://github.com/OGGM/oggm/issues/954

WIP:

- [x] Make the inversion calving a *default* workflow for the inversion
- [ ] Use 5 last pixels instead of last for inversion with calving (although I'm not sure about this one)
- [x] Tests added/passed
- [ ] Fully documented

### DIFF
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -15,6 +15,10 @@ Breaking changes
   `on the website <https://oggm.org/2020/02/16/calving-param/>`_
   (:pull:`945`).
   By `Fabien Maussion <https://github.com/fmaussion>`_
+- The calving parameterization has been extended and made more useful
+  by logging its output to compiled files. See :pull:`996` for code changes
+  and watch out for upcoming changes in the documentation and notebooks.
+  By `Fabien Maussion <https://github.com/fmaussion>`_
 - The new default when applying GCM data to a glacier is to correct for
   temperature standard deviation (:pull:`978`). The previous default was
   wrong and should not be the default.
--- a/oggm/cfg.py
+++ b/oggm/cfg.py
@@ -24,6 +24,8 @@ try:
 except ImportError:
     pass
 
+from oggm.exceptions import InvalidParamsError
+
 # Local logger
 log = logging.getLogger(__name__)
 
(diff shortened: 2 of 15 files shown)


fmaussion:
A nice side effect of this change: Coxe glacier (which had an "out-of-the box" free board of 95 meters) now has the max arbitrary free board of 50 meters. Before, the calving inversion was failing (no calving, meaning unphysical values), and with this change it now "calves"

This could be by chance (and other glaciers might get worse), but I'm quite confident that forcing the free-board within a given range (currently: 10 to 50 m) is a good thing.

cc @bearecinos @ehultee

ehultee:
Nice.  This might burn you on really big glaciers, but it seems a good baseline.  What happens if the front retreats and exposes a freeboard >50 m?  More calving, dynamic advance and thinning, total implosion?  Or is this only for static inversion now?

Also, fwiw: looks like you implemented a new default lake-terminating freeboard of 10 m (in params.cfg) which is not reflected in the comments in flowline.py L1053-1055.

fmaussion:
>  What happens if the front retreats and exposes a freeboard >50 m? More calving, dynamic advance and thinning, total implosion? Or is this only for static inversion now?

The latter - although the artifial initial water level which is used for the inversion (e.g. about 40 m for Coxe) is also used for the dynamical runs (this is consistent), but there is not physical/numerical limit on the free board anymore.

> Also, fwiw: looks like you implemented a new default lake-terminating freeboard of 10 m (in params.cfg) which is not reflected in the comments in flowline.py L1053-1055.

Thanks! I will check this + other aspects of the documentation today

fmaussion:
also closes https://github.com/OGGM/oggm/issues/869

fmaussion:
cc @jmalles if the tests pass I'll merge this as is. 

I don't have time to write a comprehensive documentation yet but I'll revisit the example notebook to take these changes into account. Basically the main changes are:
- there is now a water_level variable used here and there to make calving less crazy when DEMs our outlines are bad 
- we report volume below water and other calving related diagnostics in the compile_* files
- calving is switched on per default, but you still have to manually run the calving inversion to use it in the inversion

fmaussion:
OK, in it is. Let's cross fingers that not too many bugs are hidden in there

PR #1002, opened 2020-04-09
A concept for OGGM-Shop

fmaussion:
This is really a WIP 

But I already like it ;-)


fmaussion:
This is a first step toward a new module in oggm: `oggm.shop`. The functionality is quite poor at the moment but it will be extended.

Most use case atm is to start from an RGITOPO dem (cc @matthiasdusch). The rest will follow, mostly about adding climate and validation data to the folders, etc.

PR #1003, opened 2020-04-15
Update WGMS ref data

fmaussion:
This updates to the latest WGMS data (links + MB time series). 

(solved a new funny bug with bad DEMs on the way)

The results of the cross-val are slightly different, but not much:

image

cc @matthiasdusch  to monitor on crossval, but my understanding is that you shouldn't change anything (maybe)


fmaussion:
@matthiasdusch  I spoke too fast - for my scripts I've updated to using the RGITOPO data; you don't have to do that yet, but probably one day.

fmaussion:
The scores seem to have worsen as a result of the new additions. 

I won't spend too much time on this because this will all need a revamp anyways.

HISTALP:

```
Median bias: -27.04
Mean bias: -50.68
RMS: 442.02
Sigma bias: 0.90
```
Before: https://cluster.klima.uni-bremen.de/~mdusch/ci/1.0.2+104.g3b58edf-histalp/histalp_short.html

CRU:

```
Median bias: 3.66
Mean bias: -20.52
RMS: 517.82
Sigma bias: 0.85
```
Before: https://cluster.klima.uni-bremen.de/~mdusch/ci/1.1+3.g64245a4/cru_short.html

PR #1008, opened 2020-04-24
Move the cru and histalp funcs from climate to oggm.shop

fmaussion:
This is not a big change from outside, but much cleaner code in many aspects. 

This is backwards incompatible. "CRU_DIR" is not a thing anymore, and a couple of functions moved around -> to add CRU data you'll have to import from shop now

- [ ] Tests added/passed
- [ ] Fully documented


fmaussion:
Merging as is - without docs and whats new - because this will be part of a larger (upcoming) change anyway

PR #1015, opened 2020-05-12
Add ECMWF to the shop

fmaussion:
Still WIP - we need some functionality to apply CERA on top of ERA5 and ERA5L for the MB calibration.

- [ ] Tests added/passed
- [ ] Fully documented


fmaussion:
OK - let's go in with that. I still plenty of work to do for the documentation but this will all come in one batch later.

PR #1023, opened 2020-05-19
make process_climate entity task and DEM3 path fix

matthiasdusch:
some small fixes along the way while working on updating the crossvalidation tool.

WIP as there could pop up other stuff as well

### DIFF
--- a/oggm/core/climate.py
+++ b/oggm/core/climate.py
@@ -853,7 +853,7 @@ def local_t_star(gdir, *, ref_df=None, tstar=None, bias=None):
             tstar = amin.tstar.iloc[0]
             bias = amin.bias.iloc[0]
         else:
-            tstar = int(np.average(amin.tstar, weights=1./distances))
+            tstar = int(np.average(amin.tstar, weights=1./distances).round())
             bias = np.average(amin.bias, weights=1./distances)
 
     # Add the climate related params to the GlacierDir to make sure
--- a/oggm/core/climate.py
+++ b/oggm/core/climate.py
@@ -134,6 +134,7 @@ def process_custom_climate_data(gdir, y0=None, y1=None,
                                     source=fpath)
 
 
+@entity_task(log)
 def process_climate_data(gdir, y0=None, y1=None, output_filesuffix=None,
                          **kwargs):
     """Adds the selected climate data to this glacier directory.
@@ -201,7 +202,7 @@ def process_climate_data(gdir, y0=None, y1=None, output_filesuffix=None,
                                     output_filesuffix=output_filesuffix,
                                     **kwargs)
     else:
-        raise ValueError("cfg.['baseline_climate'] not understood")
+        raise ValueError("cfg.PARAMS['baseline_climate'] not understood")
 
 
 @entity_task(log, writes=['climate_historical'])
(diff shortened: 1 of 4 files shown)


matthiasdusch:
https://github.com/OGGM/oggm/blob/75aa05d0aca1348dc48fbc1bfc499949b5f9aec2/oggm/core/climate.py#L855

Averaging e.g. 10 glaciers with tstar==1997 resulted in 1996.9999 for me and got transformed to int 1996. Noticed this as tstar=1996 did not work with the ERA5 data anymore. Should check if this is behaviour is possible somewhere else as well.

fmaussion:
Thanks! Sorry about the ERA5 issue - I didn't test this, as it is quite a misuse of the calibration procedure - does it even work with such a short time period?

oggm/utils/_downloads.py
```diff
@@ -981,7 +981,7 @@ def _download_dem3_viewpano_unlocked(zone):
         ifile = 'http://viewfinderpanoramas.org/dem3/' + zone + 'v2.zip'
     elif zone in DEM3REG.keys():
         # We prepared these files as tif already
-        ifile = ('https://cluster.klima.uni-bremen.de/~oggm/DEM/'
+        ifile = ('https://cluster.klima.uni-bremen.de/~oggm/dem/'
```
fmaussion, line 984: How was it possible to create the RGI-TOPO files? Is there still the old path on cluster maybe?

matthiasdusch: just checked: I did the RGI topo runs 2-3 days after the paths changed in #981 
But by chance I still manually installed the commit 1 before the change in singularity. And the old path with your name is still a symlink to the new path :-)

matthiasdusch:
> Thanks! Sorry about the ERA5 issue - I didn't test this, as it is quite a misuse of the calibration procedure - does it even work with such a short time period?

Well not really a ERA5 issue. This rounding should change 50% of all t*. But I assume a 1-year change should have little effects.

The interesting ERA5 part though (regarding, does it even work?) is that most of the reference glaciers have a t*=1997 which is the lowest possible year and likely the coldest 30-year period. But this is still for the CRU parameters, so we should run a crossvalidation with different parameter combinations and decide upon a ERA5 default.

fmaussion:
>  the coldest 30-year period

if we had time for such things, playing around with a 20 or 10 years window would be something to try out

fmaussion:
Is this ready to merge?

matthiasdusch:
> Is this ready to merge?

hmmm, still a bit unsure how to handle errors from `t_star_from_refmb` propagating to `compute_ref_t_stars`. Right now I just continue on these glaciers so they are not used in the ref_tstar.csv

oggm/core/climate.py
```diff
@@ -1275,6 +1276,15 @@ def compute_ref_t_stars(gdirs):
     # Loop write
     df = pd.DataFrame()
     for gdir, res in zip(ref_gdirs, out):
+        if res is None:
+            # For certain parameters there is no valid mu candidate on certain
+            # glaciers. E.g. if temp is to low for melt. This will raise an
+            # error in t_star_from_refmb and should only get here if
+            # continue_on_error = True
+            # Do not add this glacier to the ref_tstar.csv
```
fmaussion, line 1284: I think not adding these to the ref_tstar list is OK

PR #1027, opened 2020-05-22
Verify downloaded files only once per run

fmaussion:
This was so easy to implement that it is almost embarrassing that we lost so much time because of that.

cc @matthiasdusch @TimoRoth 

I will also add a pytest option to disable it entirely when doing tests, this will make my life a bit easier as I always develop in tests and the verification takes time

### DIFF
--- a/oggm/cfg.py
+++ b/oggm/cfg.py
@@ -9,6 +9,7 @@ import sys
 import glob
 import json
 from collections import OrderedDict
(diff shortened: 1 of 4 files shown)


fmaussion:
I could write a test for this but it is not worth it.

Also the pytest option was more complex than I thought, so I'll leave it here once the tests pass

PR #1029, opened 2020-05-24
Do not replace volume in filter_inversion

fmaussion:
This is more "consistent" like this.


fmaussion:
overriden by https://github.com/OGGM/oggm/pull/1045

PR #1033, opened 2020-06-09
Migrate to Github Actions for tests

TimoRoth:
Travis has been pretty annoying more than once, and Github actions seems to be in decent enough shape now to run our tests.
Plus, it appears to be massively faster, and gets rid of quite a bit of complex shell scripting.

I'm sure there will be a few rough edges still, specially when testing pull-requests that I cannot test without it being merged.


fmaussion:
Thanks! This will need to get used to, definitely - but at core the important bits seems to be there. 

What I miss is the code coverage change? Where can I see this outputed?

TimoRoth:
It seems like it attributed the coverage to a normal push to master, instead of this PR. The coveralls script itself needs some fixes. I will send them a PR.

TimoRoth:
This seems to work fine now. I'm just not sure why the coverity is as low as it claims to be.
Only thing I don't like about it is the check-spam for every single test-job.

TimoRoth:
Seems like it's working now. And the correct way to do it causes even spam check-spam.

fmaussion:
One last request: can the failing Coveralls tests be ignored in the reports? :slightly_smiling_face:

TimoRoth:
They should not be failing anymore in future builds, unless the coverage of that specific test drops.
It seems silly to have them at all, but I can't see a way to get rid of them without breaking coveralls being able to attribute all builds to the same job.

I'll try once more, but I'm pretty sure the global coverage will drop again.

TimoRoth:
Nope, when I remove the flag for the individual reports, it breaks.
This is actually per coveralls documentation. For a parallel build, this flag has to be set.

TimoRoth:
Hm, thinking about it, it does not seem like they will become green, as they are below the threshold for a successful percentage.
What on earth were they thinking when implementing parallel builds like that?

TimoRoth:
I give up, have tried every possible combination. Their documentation contradicts itself, nothing behaves like I'd expect it to.

TimoRoth:
This looks actually fine to me now. Getting codecov working was much less of a pain than trying to fix coveralls.

fmaussion:
Thanks a lot @TimoRoth - LGTM. Let's hope that codecov is better than coveralls, which was quite annoying

PR #1040, opened 2020-07-03
Add support for "squeezed" flowlines

fmaussion:
This is still a work in progress - in this first task we have the squeezed flowlines with varying grid size. A second task will then interpolate to a regular grid.

cc @drounce 

- [x] Tests added/passed
- [x] Fully documented


fmaussion:
This is all going quite well. Here is how the squeezed flowine inversion compares to the multiple flowline:

Squeezed:

s1

Default:

s2

Their diff:

s3

fmaussion:
Next step is to compute the downstream line in order to extend them

PR #1042, opened 2020-07-20
Switch back to coveralls

TimoRoth:
They have by now figured out how to use Github Actions, and given the overall unreliability of coveralls.io with constant random test-failures, preferring them seems the right choice to me.

Also gives the test image a bump.


fmaussion:
What is the status of this? Is there something I need to do?

BtbN:
From my end it's ready, but I was hoping to have it complete a round of tests before merging it.

TimoRoth:
From my end it's ready, was just hoping for it to complete a round of tests before merging it.

fmaussion:
OK thanks. I'll push to your branch to address the remaining failures. Very annoying again

fmaussion:
@TimoRoth `untested_base:20200720` won't make it because I needed to fix salem for upstream changes in matplotlib in  https://github.com/fmaussion/salem/pull/176

So either we wait for a new untested_base to be available with latest salem or we roll back to an older base version

TimoRoth:
I've just told the Docker-Actions to build a new one. Should be available in an hour or two.

PR #1045, opened 2020-07-22
Inversion now works with trapezoid shapes - streamlines inversion -> forward model workflow

fmaussion:
This is quite a big change in the inversion process: the `mass_conservation_inversion` task now checks if the inverted parabolic beds are "too flat" during the inversion (instead of at the `init_present_time_glacier` task later on). This means that the `mass_conservation_inversion` is much simplified and totally consistent with the inversion tasks - preserving volume and area perfectly. 

I originally wanted to get rid of the `filter_inversion_output` task, but it turns out that it's not possible. Because of the equilibrium assumption, the glacier terminus is too thick and the beds simply don't "look" good. The filtering is much more sensible now however, based on parabolic bedshapes and taking the downstream bedshapes into account.

- [ ] Tests added/passed
- [ ] Fully documented

### DIFF
--- a/benchmarks/track_model_results.py
+++ b/benchmarks/track_model_results.py
@@ -199,6 +199,7 @@ class full_workflow:
         execute_entity_task(tasks.prepare_for_inversion, gdirs)
         # We use the default parameters for this run
         execute_entity_task(tasks.mass_conservation_inversion, gdirs)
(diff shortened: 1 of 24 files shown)


fmaussion:
Merging now and will take care of the graphics test in separate PR

PR #1051, opened 2020-08-05
Try to avoid random IO errors when writing to tar

fmaussion:
@TimoRoth an example of error happening on cluster: 

```
2020-08-05 09:57:38: oggm.workflow: Execute entity task gdir_to_tar on 3537 glaciers
Traceback (most recent call last):
  File "/work/fmaussion/71/oggm_env/bin/oggm_prepro", line 8, in <module>
    sys.exit(main())
  File "/work/fmaussion/71/oggm_env/lib/python3.7/site-packages/oggm/cli/prepro_levels.py", line 470, in main
    run_prepro_levels(**parse_args(sys.argv[1:]))
  File "/work/fmaussion/71/oggm_env/lib/python3.7/site-packages/oggm/cli/prepro_levels.py", line 222, in run_prepro_levels
    utils.base_dir_to_tar(l_base_dir)
  File "/work/fmaussion/71/oggm_env/lib/python3.7/site-packages/oggm/utils/_workflow.py", line 2855, in base_dir_to_tar
    tar.add(dirname, arcname=os.path.basename(dirname))
  File "/usr/local/pyenv/versions/3.7.8/lib/python3.7/tarfile.py", line 1949, in add
    recursive, filter=filter)
  File "/usr/local/pyenv/versions/3.7.8/lib/python3.7/tarfile.py", line 1941, in add
    with bltn_open(name, "rb") as f:
FileNotFoundError: [Errno 2] No such file or directory: '/home/users/fmaussion/run_scripts/run_rgitopo/RGI62/b_010/L0/RGI60-12/RGI60-12.00/RGI60-12.00845.tar.gz'
srun: error: node06: task 0: Exited with exit code 1
srun: Terminating job step 71.0
```

This is truely random but happens quite often, at least once among the 19 RGI regions. It doesn't make sense because the file which is apparently "not there" should be created by the statement, so this has to be some OS or NFS issue not responding fast enough or something.

### DIFF
--- a/oggm/cli/prepro_levels.py
+++ b/oggm/cli/prepro_levels.py
@@ -230,7 +230,7 @@ def run_prepro_levels(rgi_version=None, rgi_reg=None, border=None,
 
     # Which DEM source?
     if dem_source.upper() == 'ALL':
-        # This is the complex one, just do the job an leave
+        # This is the complex one, just do the job and leave
         log.workflow('Running prepro on ALL sources')
         for i, s in enumerate(utils.DEM_SOURCES):
             rs = i == 0
@@ -430,6 +430,8 @@ def parse_args(args):
         if rgi_reg is None:
             raise InvalidParamsError('--rgi-reg is required!')
     rgi_reg = '{:02}'.format(int(rgi_reg))
+    if rgi_reg not in ['{:02}'.format(int(r)) for r in range(1, 20)]:
+        raise InvalidParamsError('--rgi-reg should range from 01 to 19!')
 
     rgi_version = args.rgi_version
 
--- a/oggm/cli/prepro_levels.py
+++ b/oggm/cli/prepro_levels.py
@@ -430,7 +430,8 @@ def parse_args(args):
         if rgi_reg is None:
             raise InvalidParamsError('--rgi-reg is required!')
     rgi_reg = '{:02}'.format(int(rgi_reg))
-    if rgi_reg not in ['{:02}'.format(int(r)) for r in range(1, 20)]:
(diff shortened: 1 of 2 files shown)


TimoRoth:
What do you mean, should be created by the statement?
It's trying to read a file, which is a tar file, and trying to add it to another tar file?
The odd thing is, it seems like it goes into the isdir path in https://github.com/python/cpython/blob/v3.7.8/Lib/tarfile.py#L1949 which makes sense, but then a file it just enumerared from that dir apparently isn't there?

What confuses me a bit is why it's trying to tar up something from /home, while it's working in /work?

fmaussion:
> What do you mean, should be created by the statement?
> It's trying to read a file, which is a tar file, and trying to add it to another tar file?

Yes sorry you are right.

> The odd thing is, it seems like it goes into the isdir path in https://github.com/python/cpython/blob/v3.7.8/Lib/tarfile.py#L1949 which makes sense, but then a file it just enumerared from that dir apparently isn't there?

Yes I did not try to make sense of these errors because they are random. Restarting the run could work for example.

> What confuses me a bit is why it's trying to tar up something from /home, while it's working in /work?

Yes, that's part of the `prepro_gdir` cli workflow: instead of writing the file locally I directly tar them on /home -> that's specific to this workflow though. Should I try differently?

TimoRoth:
No, this should definitely work, just struck me as a bit odd.
I'll try changing some options on the NFS server.

fmaussion:
You are right, it's an OS/NFS issue. It's trying to tar the entire `RGI60-12.00` directory, so `RGI60-12.00845.tar.gz` is not given by the python code, it's provided by the OS as being there but then erroring at writing.

PR #1053, opened 2020-08-06
Avoid caching sample files in current dir on cluster

fmaussion:
When installing a new OGGM on an image the oggm-sample data is not available and downloaded in an ad-hoc cache. This should be cleaner

### DIFF
--- a/oggm/utils/_downloads.py
+++ b/oggm/utils/_downloads.py
@@ -275,9 +275,13 @@ def _cached_download_helper(cache_obj_name, dl_func, reset=False):
         fb_cache_dir = os.path.join(cfg.PATHS['working_dir'], 'cache')
         check_fb_dir = False
     except KeyError:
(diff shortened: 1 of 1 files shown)


TimoRoth:
Wouldn't it be better to explicitly configure cfg.CACHE_DIR to something in /home, so that stuff is only actually downloaded once?

fmaussion:
> Wouldn't it be better to explicitly configure cfg.CACHE_DIR to something in /home, so that stuff is only actually downloaded once?

That's the thing -> cfg.CACHE_DIR *is* it home, but on singularity it still pollutes the directory from which the script is started. Let me check why

fmaussion:
Merging now - I think it's OK as it is but happy for a better solutions. It mimicks the caching mechanism on cluster, and is only necessary because cfg.PATHS[`working_dir`] is not set yet

PR #1054, opened 2020-08-06
Add option to override params.cfg parameters at initialize time

TimoRoth:
This should avoid the weird corner-case where cfg.initialize relies on config variables the script in question might set later on.

Also implements its use in prepro_cli.

### DIFF
--- a/oggm/cfg.py
+++ b/oggm/cfg.py
@@ -406,10 +406,11 @@ def initialize_minimal(file=None, logging_level='INFO', params=None):
     PATHS['climate_file'] = cp['climate_file']
 
     # Ephemeral paths overrides
-    if os.environ.get('OGGM_WORKDIR') is not None:
-        PATHS['working_dir'] = os.environ.get('OGGM_WORKDIR')
-        log.workflow('Overriding initial working dir with OGGM_WORKDIR: ' +
-                     PATHS['working_dir'])
+    env_wd = os.environ.get('OGGM_WORKDIR')
+    if env_wd and not PATHS['working_dir']:
+        PATHS['working_dir'] = env_wd
+        log.workflow("PATHS['working_dir'] set to env variable $OGGM_WORKDIR: "
+                     + env_wd)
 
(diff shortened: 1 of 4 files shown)


fmaussion:
Thanks! This might work for oggm_prepro but then all other scripts should do that as well I guess... All the oggm scripts around are doing the same:

```python
cfg.initialize()
cfg.PARAMS[].... 
```

fmaussion:
Another way would be to initialize the working dir with $OGGM_WORKDIR in the same function

fmaussion:
a bit like `oggm_static_paths()`

TimoRoth:
Then it would always ignore the workdir set in the params, which might not be what the users wants.
The cli script (prepro) already evaluates that variable, and it seems the better place for that to me.

fmaussion:
But what I'm saying is that other OGGM scripts will have the same issue then.

TimoRoth:
Yeah, the scripts will have to do the same if they rely on that kinda behaviour.
But do we really want "magic" like that in oggm itself?

fmaussion:
> it would always ignore the workdir set in the params, 

No, because you can always override it. All current OGGM scripts (as per teh documentation, etc) start like:

```python
cfg.initialize()
cfg.PATHS['working_dir'] = 'my path'
```

at the moment thei overwrite an empty string, I don't think is a big issue if they overwrite $OGGM_WORKDIR

TimoRoth:
The whole situation just seems very bad to me, that cfg.initialize relies on cfg variables itself, which the user commonly sets after it.
I'd rather have the recommendations to change to set more parameters via this new params parameter, seems a lot cleaner and less prone to weird behavior due to differing config variables between init and runtime.

fmaussion:
"Differing config variables between init and runtime" is sometimes what you want - but having another way to change it is good as well.

fmaussion:
Thanks! I'd like to add a tiny thing, I'll push to your branch

Issue #1061, opened 2020-09-24
Installing oggm in pyenv makes pytest weird


fmaussion:
@TimoRoth I have a fantastic one for you.

simple_py3 is a pyenv environment based on python 3.8.5, with all packages installed except oggm (at first).

`dummy` is simply a folder with a dummy test in it. 

Now look at this:

```bash
(simple_py3) mowglie@flappi-top:~/Documents/git/dummy$ time pytest .
============================= test session starts ==============================
platform linux -- Python 3.8.5, pytest-6.0.2, py-1.9.0, pluggy-0.13.1
Matplotlib: 3.3.2
Freetype: 2.6.1
rootdir: /home/mowglie, configfile: pytest.ini
plugins: mpl-0.122
collected 1 item                                                               

test_dummy.py .                                                          [100%]

============================== 1 passed in 0.14s ===============================

real	0m0,378s
user	0m0,653s
sys	0m0,367s
(simple_py3) mowglie@flappi-top:~/Documents/git/dummy$ cd ../oggm-official/
(simple_py3) mowglie@flappi-top:~/Documents/git/oggm-official$ pip install -e .
Obtaining file:///home/mowglie/disk/Dropbox/HomeDocs/git/oggm-official
Requirement already satisfied: numpy in /home/mowglie/.pyenv/versions/3.8.5/envs/simple_py3/lib/python3.8/site-packages (from oggm==1.3.2.dev95+gc05d330) (1.19.2)
Requirement already satisfied: scipy in /home/mowglie/.pyenv/versions/3.8.5/envs/simple_py3/lib/python3.8/site-packages (from oggm==1.3.2.dev95+gc05d330) (1.5.2)
Requirement already satisfied: pandas in /home/mowglie/.pyenv/versions/3.8.5/envs/simple_py3/lib/python3.8/site-packages (from oggm==1.3.2.dev95+gc05d330) (1.1.2)
Requirement already satisfied: matplotlib>=2.0.0 in /home/mowglie/.pyenv/versions/3.8.5/envs/simple_py3/lib/python3.8/site-packages (from oggm==1.3.2.dev95+gc05d330) (3.3.2)
Requirement already satisfied: shapely in /home/mowglie/.pyenv/versions/3.8.5/envs/simple_py3/lib/python3.8/site-packages (from oggm==1.3.2.dev95+gc05d330) (1.7.1)
[...]

PR #1062, opened 2020-09-24
Use local conftest.py instead of installing global pytest plugin

TimoRoth:
This avoids needlessly importing oggm into each and every pytest instance started from the same venv.


fmaussion:
There is already a conftest here: https://github.com/OGGM/oggm/blob/master/oggm/tests/conftest.py 

Better merge them?

TimoRoth:
pytest works its way up the directory tree when looking for conftest plugins, so I just renamed the existing file to save cluttering the already existing one up.
Moving everything in there would work as well though, but make it a fairly large and multi-purposed file.

fmaussion:
I prefer to merge these two files, because I don't thing a conftest.py file should be present in the oggm namespace since it's really only for tests. I'll merge this soon.

closes https://github.com/OGGM/oggm/issues/1061, although the issue with pycharm remains - I have to investigate, but it is something nasty going on either in pycharm or oggm with threads when oggm is imported

PR #1064, opened 2020-09-25
Make the test dir more unique using environment version ids

fmaussion:
It happened to me for the first time that my various envs were not happy sharing pickles (this is a bit worrying). But also, this change will make things much more predictable

### DIFF
--- a/oggm/utils/_workflow.py
+++ b/oggm/utils/_workflow.py
@@ -136,7 +136,7 @@ get_temp_dir = gettempdir
 
 
 def get_sys_info():
-    """Returns system information as a dict"""
+    """Returns system information as a list of tuples"""
 
     blob = []
     try:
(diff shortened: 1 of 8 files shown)


oggm/utils/_workflow.py
```diff
@@ -203,12 +188,42 @@ def show_versions(logger=None):
         except BaseException:
             deps_blob.append((modname, None))
 
+    return deps_blob
+
+
+def get_git_ident():
+    ident_str = '$Id: f2cb553755951450761f5ee64e4a4c433ef61521 $'
```
TimoRoth, line 195: This is not going to work unless you also change .gitattributes accordingly.

fmaussion: Thanks - will do

PR #1068, opened 2020-10-02
Re-Try tar extract on empty result

TimoRoth:
This is a horrible hack and I can't say I like it, but it does work around the issue reliably from what I tested.


fmaussion:
Welcome to the horrible hack world! I like it if it works ;)

PR #1069, opened 2020-10-14
change how the length of a glacier is calculated

### LINKED ISSUE #914: Problem with flowline length definition
the flowline length is calculated by the following code in the `class` `Flowline(Centerline):`
`@property`
`def` `length_m(self):`
-  `pok = np.where(self.thick> 0.)[0]` 
    `  return  len(pok) * self.dx_meter` 

However, self.thick is sometimes extremely small at the end of a glacier in OGGM (e.g. 1e-17 m!  on some cases for Hintereisferner). 
I would  suggest a threshold of at least 0.1m  or 1m thickness for a glacier to exist. This will decrease slightly the length and area compared to the current definition. 
A suggested change in code would be:
  `pok = np.where(self.thick > 0.1)[0]`

matthiasdusch:
Currently the `length_m` property of a FlowlineModel calculates the length of the main flowline based on every glaciated grid point.  If this flowline is sperated (ice free) at some point this will still take into account the "dead" ice further downstream.

Problems with current implementation:
If there is long lasting dead ice downstream while the main glaicer retreats this is not (or less) reflected in the length property.
This happens for me in two cases:
i) if two glaciers are merged and one of them behaves quite differently due to wrong calibration. This is a rather individual problem and I could also calculate the glacier length seperately.
ii) Probably more likely: I have also seen this for strongly retreating glaciers with steep slopes above a flat tongue. Here the ice on the slopes might vanish before the tongue is completely melted. 

Proposed fix:
Calculate the glacier length as the continuous ice along the main flowline. 

Problems/Implications of the fix:
E.g. if the flowline seperates for a short time period this could cause strong fluctuations in the time series.

Still to consider:
What about volume? 

A chance to also close #914
I am happy to take suggestions for minimum thickness of a glacier...

- [ ] Tests added/passed
- [ ] Fully documented

@fmaussion, @lilianschuster

### DIFF
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -53,6 +53,10 @@ Breaking changes
 - the `vascaling` module has been removed from oggm core (:pull:`1065`). It
   is now available via a separate package (`oggm-vas <https://github.com/OGGM/oggm-vas>`_,
   maintained by Moritz Oberrauch).
+- new options to compute the length of a glacier during a run:
+  `PARAMS['min_ice_thick_for_length']` and `PARAMS['glacier_length_method']`
+  (:pull:`1069`). By `Matthias Dusch <https://github.com/matthiasdusch>`_.
+
 
 Enhancements
 ~~~~~~~~~~~~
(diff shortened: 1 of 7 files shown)


fmaussion:
Added the change as an option for now. Note that the default is still the "old" behaviour, for your use case you should set 'consecutive'.

I've also tried to address #914 with a limit of 1m for now.

fmaussion:
TODO: tests for the new "consecutive option"

matthiasdusch:
lgtm

the minimum ice thickness is a tricky one. I think there should be a threshold but maybe this needs a bit of testing. keeping it optional for now seams reasonable.

PR #1075, opened 2020-10-20
make merged glaciers use original DEM source

matthiasdusch:


fmaussion:
Why is this necessary? I think that this should not need a change

matthiasdusch:
the rgi_id of merged glaciers has a '_merged' suffix which can not be found in the new DEM_SOURCE_TABLE. Another solution would be to exclude such suffixes within `default_dem_source`

matthiasdusch:
first idea opened a can of worms including user specified DEM paths like those during testing.

So I propose this simpler fix which just assures that no rgi_id suffixes are considered at `default_dem_source`

fmaussion (approved):
That's better I think!

PR #1099, opened 2020-11-19
Monster lockdown push PR: ERA5, calving, and more

fmaussion:
I did what we should not do: a monster PR. This has grown a bit out of control, but I'll merge soon and hope for the best.

In order not to let this linger further, I'll try to write a proper documentation in a big documentation push over the holidays.

Here is a summary of the main changes:
- `calibrate_inversion_from_consensus_estimate` is renamed to `calibrate_inversion_from_consensus` and works much better. It also allows to use sliding in case the consensus volumes are way too low for oggm.
- the logging level "WORKFLOW" is now logging at a higher level than errors or warnings, i.e. setting to "WORKFLOW" will hide individual glacier error and avoid cluttering of the run output (but also hiding potential very large numbers of errors). Because many of the tutorials have this logginf level in them I have to add a deprecation warning in code
- the `prepro_levels` CLI has a large number of changes. 
   - We now can select normal flowlines, elevation band flowlines, or a mix of both. It's a long story, but for global runs I'm really leaning towards making elevation band flowlines the default.
   - The default workflow now calibrates A from the consensus at the regional level, 
   - and also calibrates the mass-balance at the regional level using regional MB estimates. This is a very coarse and temporary solution
   - there are now 5 instead of 4 prepro levels. The 5th level contains also the "warm-up" historical runs, i.e. GCM runs can be started from 2019 (depending on historical climate) directly
   - there is a new utility function called `utils.extend_past_climate_run` which extends the past historical runs to account for late RGI date, by assuming a fixed glacier geometry
- there was a bug in `_filter_small_slopes`: in this task I cut the flowline for a min slope, which was set to 0. But in the inversion it is set to 1.5°! So now this is correct, but it actually changes the look of the flowlines in some cases, and the reason why I had to update the graphics test
- `elevation_band_flowline` now can also bin other data as well, e.g. the thicknesses from the consensus
- `invert_with_calving` now has a new parameter, `calving_min_mu_star_frac`, which tells OGGM not to reduce mu* below a certain fraction of the original (without calving) mu*
- the flowline model's `run_until_and_store` now writes all variables (including calving ones) regardless if it is a calving glacier or not. This makes the output larger but more predictable (I think: not tooooo large since we are using compression...)
- we now use `cftime` to read the CMIP file's time out of the netcdf
- I checked for many places where we clip the slope to be sure, but I think this should be checked again.
- new `bedtopo.py` module in the shop to project the consensus topography into a glacier directory.

TODOS:
- [x] deprecation warning in logging level
- [x] in another PR: make the Ice caps flowlines start from highest point instead of highest on outline.

### DIFF
--- a/docs/run_examples/_code/mb_crossval.py
+++ b/docs/run_examples/_code/mb_crossval.py
@@ -21,7 +21,7 @@ log = logging.getLogger(__name__)
 rgi_version = '62'
 
 # Initialize OGGM and set up the run parameters
-cfg.initialize(logging_level='WORKFLOW')
+cfg.initialize(logging_level='ERROR')
 
 # Local paths (where to find the OGGM run output)
 dirname = 'OGGM_ref_mb_RGIV{}_OGGM{}'.format(rgi_version, oggm.__version__)
--- a/docs/run_examples/_code/run_errors.py
+++ b/docs/run_examples/_code/run_errors.py
@@ -13,7 +13,7 @@ start = time.time()
 log = logging.getLogger(__name__)
 
 # Initialize OGGM and set up the default run parameters
-cfg.initialize(logging_level='WORKFLOW')
+cfg.initialize(logging_level='ERROR')
 
 # Here we override some of the default parameters
 # How many grid points around the glacier?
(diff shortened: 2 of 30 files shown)


fmaussion:
OK let's let this beast in, and never do that again.

(hope dies last)

PR #1108, opened 2020-12-07
Update test image

TimoRoth:
Switched the test images around.
The default one is now Python 3.8.

Some tests are failing on it, at least on the images own validation build.


TimoRoth:
I have a hard time telling if the failing tests are an issue in oggm, salem, or (py)proj being broken.

fmaussion:
Thanks! I'll have a look at it soon.

If you feel like it we could tackle the question if salem has an issue too in the process of migrating salem to github actions :blush:

TimoRoth:
Will re-create after the big changes are merged.

PR #1120, opened 2020-12-17
Make sure only ever one Manager exists

TimoRoth:
Calling Manager(), specially on the global scope, creates and starts a new Manager-Process every time.
This in theory works, but is not exactly nice and "used as intended" and has for unknown reasons lead to multiprocessing locking up while starting the manager.

I cannot say for sure if this fixes the weird deadlock issue we are seeing, but it's certainly an improvement.

### DIFF
--- a/oggm/cfg.py
+++ b/oggm/cfg.py
@@ -45,8 +45,8 @@ CONFIG_FILE = os.path.join(os.path.expanduser('~'), '.oggm_config')
 CONFIG_MODIFIED = False
 
 # Share state accross processes
-DL_VERIFIED = Manager().dict()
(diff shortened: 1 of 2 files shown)


fmaussion:
Thanks! I don't recal we have tests for this anyway, but I'm just going to assume that this works as expected, i.e. two processes won't verify the same file in the same job? 

Feel free to merge, thanks a lot!!!!

TimoRoth:
The only reason that ever worked before was forking multiprocessing mode, which implicitly has the same manager for all processes.
So in spawn mode, without this change, it plain wouldn't sync the dicts to begin with, since every worker spawns its own two Managers.

PR #1122, opened 2020-12-18
Fix stalling problem on cluster

fmaussion:


fmaussion:
@TimoRoth  that was it

fmaussion:
https://github.com/pydata/xarray/issues/4710

Issue #1107, opened 2020-12-02
utils.get_cmip5_file() don't work anymore


Keeptg:
The [utils.get_cmip5_file()](https://github.com/OGGM/oggm/blob/e20e673a0e50151f5a195ba29e1890d8447fd3e7/oggm/utils/_downloads.py#L2313) function seems not work anymore. I guess it is because the  dataset can't be accessed by the default link.

fmaussion:
Thanks for the report! I forgot that this function existed. 

In my code I use `file_downloader` with the following url: https://cluster.klima.uni-bremen.de/~oggm/cmip5-ng/

fmaussion:
(we should either delete the function or update it)

PR #1124, opened 2020-12-20
Fix cmip5 link

Keeptg:
Closes https://github.com/OGGM/oggm/issues/1107

- [x] Tests added/passed
- [ ] Fully documented


fmaussion:
That's great thanks!

Eventually I think that we should either:
- add more functionality to this (like offering a list of available GCMs / scenarios, etc), add CMIP6 data 
- or remove it entirely.

Let's discuss this eslewhere

PR #1133, opened 2021-01-05
Fix a bug where the calving bucket would melt more than one grid point

fmaussion:
- [x] Tests added/passed
- [ ] Fully documented


fmaussion:
This also adds more diagnostics to the CFL error message.

PR #1135, opened 2021-01-05
Do not error when SLURM_JOB_CPUS_PER_NODE is weird

fmaussion:
@TimoRoth , fyi, @drounce is working on a system where `SLURM_JOB_CPUS_PER_NODE` gets the value `24(x2)` 

I wonder if these shenanigans with SLURM_JOB_CPUS_PER_NODE are good or if we should just rely on `multiprocessing.cpu_count()` always?

### DIFF
--- a/oggm/cfg.py
+++ b/oggm/cfg.py
@@ -471,7 +471,7 @@ def initialize_minimal(file=None, logging_level='INFO', params=None,
             mpp = int(os.environ['SLURM_JOB_CPUS_PER_NODE'])
             log.workflow('Multiprocessing: using slurm allocated '
                          'processors (N={})'.format(mpp))
-        except KeyError, ValueError:
+        except (KeyError, ValueError):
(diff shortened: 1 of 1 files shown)


TimoRoth:
> I wonder if these shenanigans with SLURM_JOB_CPUS_PER_NODE are good or 
> if we should just rely on |multiprocessing.cpu_count()| always?

That'd lead to potentially massive over or under-subscription of shared 
workers, which would hurt performance and worst case affect potential 
other users of the node.
This change seems sensible to prevent non-numerical values.

I never heard of slurm setting it like that, the official documentation 
describes it as integer number.
Is this some third party tool trying to mimic a slurm environment?

drounce:
I don't believe there's anything special.

https://uaf-rcs.gitbook.io/uaf-rcs-hpc-docs/using-batch/batch-scripts

PR #1137, opened 2021-01-15
fix bug: core/gis.py: use non-existent topo_valid_mask key from nc.variables

8fdafs2:
solution: delete the related line

### DIFF
--- a/oggm/core/gis.py
+++ b/oggm/core/gis.py
@@ -758,7 +758,6 @@ def glacier_masks(gdir):
         v[:] = glacier_ext
 
         dem = nc.variables['topo'][:]
-        valid_mask = nc.variables['topo_valid_mask'][:]
 
         # Last sanity check based on the masked dem
         tmp_max = np.max(dem[np.where(glacier_mask == 1)])


TimoRoth:
On 15.01.2021 11:55, 8fdafs2 wrote:
> solution: delete the related line

I'm not sure how that is a solution.
That variable is used in the function, and just removing it is going to 
cause issues further down the line.

fmaussion:
Are there some bots that do this kind of stuff in PRs?

If yes, why would someone use such a bot?

8fdafs2:
> 
> 
> Are there some bots that do this kind of stuff in PRs?
> 
> If yes, why would someone use such a bot?

I have to admit it's a reckless PR, but not a bot here.

8fdafs2:
> 
> On 15.01.2021 11:55, 8fdafs2 wrote: solution: delete the related line
> I'm not sure how that is a solution. That variable is used in the function, and just removing it is going to cause issues further down the line.

I apologize it's reckless PR. I totally ignored the use of 'valid mask' in the following lines.
The reason why I removed this line is that there are some KeyError threw at the removed line as the key 'topo_valid_mask' doesn't exist at that moment and the issue seems to get 'solved' after the removal of that line.

8fdafs2:
This PR is not a solution but I suspect the issue I mentioned above might be a real one.

Issue #1101, opened 2020-11-23
allow to use calendar years in place of hydro years


fmaussion:
if the user sets:

```
hydro_month_nh = 1
hydro_month_sh = 1
```

it means that the hydro year convention does not hold

The fix should be really easy

fmaussion:
closed via https://github.com/OGGM/oggm/pull/1142

PR #1142, opened 2021-01-21
Allows hydro_month=1 and massbalance-sandbox to work with OGGM

lilianschuster:
Discussed together with @fmaussion that the hydro_years should be the year of the last month. So, in case of hydro_month_nh/hydro_month_sh = 1, the last month is 12 (December) and the hydro_year is equal to y0, if hydro_month_nh/hydro_month_sh  > 1 , the last month is in the next year, hence hydro_year is y0+1, which is how it has been so far. This only gives changes if the option hydro_month_nh/hydro_month_sh = 1 is chosen, otherwise no changes occur.

In addition, some minor changes in the code are made in order that the ERA5_daily dataset and different mb_modules can be used with https://github.com/OGGM/massbalance-sandbox. 

If this is merged, it should close the issue https://github.com/OGGM/oggm/issues/1101
If it is merged, this might need to be documented somewhere, but not sure where !

All old tests passed and one new one has been created for testing changes in hydro_month!

- [x] Tests added/passed


oggm/cfg.py
```diff
@@ -253,6 +253,13 @@ def _log_param_change(self, key, value):
 _doc = 'The historical monthly climate timeseries stored in a netCDF file.'
 BASENAMES['climate_historical'] = ('climate_historical.nc', _doc)
 
+# so far, this is only ERA5_daily and does not work with the default OGGM
+# mass balance module
+_doc = 'The historical daily climate timeseries stored in a netCDF file. \
+    (only temperature is really changing on daily basis, precipitation is just\
+     assumed constant for every day)'
+BASENAMES['climate_historical_daily'] = ('climate_historical_daily.nc', _doc)
```
fmaussion, line 261: Can you use parentheses to snitch multi-line strings instead of `\` ? (see examples above)

oggm/tests/test_shop.py
```diff
@@ -402,6 +402,135 @@ def test_vdr(self, class_case_dir):
             # Fake tests, the plots look plausible
             np.testing.assert_allclose(d2.gradient.mean(), -0.0058, atol=.001)
             np.testing.assert_allclose(d2.temp_std.mean(), 3.35, atol=0.1)
+            
+
```
fmaussion, line 406: one or two empty lines too many

oggm/tests/test_shop.py
```diff
@@ -402,6 +402,135 @@ def test_vdr(self, class_case_dir):
...
+         'ERA5':{'start_year':1979,'end_year':2018},
+         'ERA5dr':{'start_year':1979,'end_year':2019},
+         'HISTALP':{'start_year':1850,'end_year':2014},
+          'CERA':{'start_year':1901,'end_year':2010},
+         'ERA5L':{'start_year':1981,'end_year':2018},
+        }
+        
+        gdir = hef_gdir
+        # this workkaround needed to get h,w because no fls there apparently
+        # maybe there is something easier in the testing environment
```
fmaussion, line 425: https://github.com/OGGM/oggm/blob/4dfbfa39e8b22a13ecde936387ce3944ef7cb583/oggm/utils/_workflow.py#L2581

oggm/utils/_workflow.py
```diff
@@ -2321,7 +2321,13 @@ def np_convert(o):
 
     def get_climate_info(self, input_filesuffix=''):
         """Convenience function handling some backwards compat aspects"""
-
+            
+        # I don't know another way right now, 
+        # as I was not able to get the input_filesuffix 
+        # to those functions that use get_climate_info 
+        if cfg.PARAMS['baseline_climate'] == 'ERA5_daily':
+             input_filesuffix = '_daily'
```
fmaussion, line 2329: that's really not clean indeed. I'm not sure to understand yet where you need that, but maybe it's something to delegate to the mass-balance models which are calling that?

oggm/utils/_workflow.py
```diff
@@ -2546,7 +2559,16 @@ def write_monthly_climate_file(self, time, prcp, temp,
...
             v = nc.createVariable('prcp', 'f4', ('time',), zlib=zlib)
             v.units = 'kg m-2'
-            v.long_name = 'total monthly precipitation amount'
+
+            # check if prcp has really monthly format 
+            if len(prcp)==(nc.hydro_yr_1-nc.hydro_yr_0+1)*12:
+                v.long_name = 'total monthly precipitation amount'
+            else:
+                v.long_name = 'total monthly precipitation amount'
+                import warnings
```
fmaussion, line 2568: import at top of module

oggm/utils/_workflow.py
```diff
@@ -2546,7 +2559,16 @@ def write_monthly_climate_file(self, time, prcp, temp,
...
+                warnings.warn("there might be a conflict in the prcp"
+                              "timeseries, please check!")
```
fmaussion, line 2570: Why is this needed?

oggm/utils/_workflow.py
```diff
@@ -2663,7 +2685,7 @@ def get_ref_mb_data(self, y0=None, y1=None):
             out = self._mbdf
         return out.dropna(subset=['ANNUAL_BALANCE'])
 
-    def get_ref_mb_profile(self):
+    def get_ref_mb_profile(self, check = True):
```
fmaussion, line 2688: `check=True`

oggm/workflow.py
```diff
@@ -304,10 +304,14 @@ def init_glacier_regions(rgidf=None, *, reset=False, force=False,
         prepro_border = int(cfg.PARAMS['border'])
 
     if from_prepro_level and prepro_border not in [10, 80, 160, 250]:
+        # does not work for HUSS flowlines ...
```
fmaussion, line 307: This we need to talk offline in more detail

fmaussion (changes requested):
Thanks! I've added a few comments and will have a detailed look on the next round.

oggm/utils/_workflow.py
fmaussion: Or, if that's too many changes in code, at least you can do:

```python
        if not input_filesuffix and (cfg.PARAMS['baseline_climate'] == 'ERA5_daily'):
             input_filesuffix = '_daily'
```

lilianschuster: I will use your proposition. The problem is that I don't understand how `get_climate_info()` is used in OGGM when an input_filesuffix is used (e.g. in get_ref_mb_data, climate.py, massbalance.py). How does it make sure that the right climate is used? If I have now also a daily climate file and want to use the same methods, (get_ref_mb_data...), it had happened that the wrong climate file is used for get_climate_info(), therefore I needed to change this in OGGM, but I guess I misunderstand here something.

oggm/utils/_workflow.py
lilianschuster: ok, maybe this won't happen anymore, I just thought that if somehow someone uses daily instead of monthly data as input, they would get a warning at least, but I won't do this anymore as I have now a separate function, so  I will remove it

oggm/tests/test_shop.py
lilianschuster: changed, long story, I did not use get_inversion_flowline_hw, because there was no inversion_flowline when using the test environment with class_case_dir as in e.g. test_all_at-once, so I changed several things, among others I use instead the hef_gdir, ... still a bit confused about all these test things, but it will come ...

fmaussion (approved):

fmaussion:
Thanks a lot @lilianschuster ! Awesome first contribution :smiley_cat:

Issue #1145, opened 2021-01-23
Remove ELA from current model output files


fmaussion:
We added the ELA to the output files on popular demand (@anoukvlug  ;-)), and it was useful and will continue to be, but on insight it should not be part of the model output files.

ELA is a diagnostic variable that can be computed on post processing after the run (it's not even dependent on glacier geometry). I think that from now on it should be part of a new "SMB diagnostics" file, which design needs to be defined soon when we work on the other hydrological diagnostics for e.g.  https://github.com/OGGM/oggm/issues/1114

PR #1152, opened 2021-01-25
Remove ELA from current model output files

fmaussion:
Closes https://github.com/OGGM/oggm/issues/1145


fmaussion:
closes https://github.com/OGGM/oggm/issues/1153

PR #1157, opened 2021-01-27
Error: Default T_Melt is -1°C and not 0°C

sarah-hanus:
The default value of T_Melt is -1 if you initialize the parameters and print cfg.PARAMS['temp_melt']. This is also the value stated in the OGGM paper.


fmaussion:
Thanks a lot!

PR #1166, opened 2021-01-30
Try trapezoid lambdas=2

fmaussion:


fmaussion:
OK so this is rather a minor change, but in practice it could lead to dirty problems if people are using old directories with the new defaults. But yeah, v1.4 will be a fresh start and we will be careful communicating this.

PR #1178, opened 2021-02-04
allow prcp_fac to change in PastMassbalance instance

lilianschuster:
This allows to change the precipitation scaling factor after instantiation of the PastMassBalance class
to be faster when optimising this factor. Some tests are added. 

- [x] Tests added/passed
- [ ] Fully documented

### DIFF
--- a/oggm/core/massbalance.py
+++ b/oggm/core/massbalance.py
@@ -416,7 +416,7 @@ class PastMassBalance(MassBalanceModel):
         # just to check that no invalid prcp_factors are used
         if new_prcp_fac <= 0:
             raise InvalidParamsError('prcp_fac has to be above zero!')
-        self.prcp = self.prcp * new_prcp_fac / self._prcp_fac
+        self.prcp *= new_prcp_fac / self._prcp_fac
         # update old prcp_fac in order that it can be updated
         # again ...
         self._prcp_fac = new_prcp_fac
--- a/oggm/core/massbalance.py
+++ b/oggm/core/massbalance.py
@@ -13,7 +13,7 @@ from oggm.cfg import SEC_IN_YEAR, SEC_IN_MONTH
 from oggm.utils import (SuperclassMeta, lazy_property, floatyear_to_date,
                         date_to_floatyear, monthly_timeseries, ncDataset,
                         tolist, clip_min, clip_max, clip_array)
-from oggm.exceptions import InvalidWorkflowError
+from oggm.exceptions import InvalidWorkflowError, InvalidParamsError
 from oggm import entity_task
 
 # Module logger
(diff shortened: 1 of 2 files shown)


oggm/core/massbalance.py
```diff
@@ -385,6 +396,31 @@ def __init__(self, gdir, mu_star=None, bias=None,
...
+        if self.check_calib_params:
+            msg = ('You want to change the precipitation scaling'
+                   'factor which was used for calibration. Set '
+                   '`check_calib_params=False` to ignore this '
+                   'warning.')
+            raise InvalidWorkflowError(msg)
+        # just to check that no invalid prcp_factors are used
+        if new_prcp_fac <= 0:
+            raise InvalidParamsError('prcp_fac has to be above zero!')
+        self.prcp = self.prcp * new_prcp_fac / self._prcp_fac
```
fmaussion, line 419:
```suggestion
        self.prcp *= new_prcp_fac / self._prcp_fac
```

fmaussion: For performance purposes

oggm/tests/test_models.py
```diff
@@ -350,6 +351,49 @@ def test_past_mb_model(self, hef_gdir):
...
 
+    def test_prcp_fac_update(self, hef_gdir):
+
+        gdir = hef_gdir
+        init_present_time_glacier(gdir)
+
+        mb_mod = massbalance.PastMassBalance(gdir, bias=0,
+                                             check_calib_params=False)
+        # save old precipitation time series
+        prcp_old = mb_mod.prcp
```
fmaussion, line 362:
```suggestion
        prcp_old = mb_mod.prcp.copy()
```
Because of the optimisation above

oggm/tests/test_models.py
```diff
@@ -350,6 +351,49 @@ def test_past_mb_model(self, hef_gdir):
...
+
+        # check if an error is raised if check_calib_params
+        # is true
+        mb_mod = massbalance.PastMassBalance(gdir, bias=0,
+                                             check_calib_params=True)
+        with pytest.raises(InvalidWorkflowError):
+            mb_mod.prcp_fac = 10
+
+        # normally this should also be implemented for mu_star ...
+        # with pytest.raises(InvalidWorkflowError):
```
fmaussion, line 393: Yes, I arguably thought that the chances that someone changes mu* happens by mistake are low, but it's also OK to be careful. Leave this bit here so that we remember

fmaussion (changes requested):
Thanks a lot! Minor suggestions from my side

lilianschuster:
I accepted all og @fmaussion requested changes, but Github tells me that there are still changes requested!  What do I have to do?

fmaussion:
GitHub is not always correct ;-)

Thanks Lily! Great addition

PR #1187, opened 2021-02-10
Cloud doc edits

ehultee:
- [Y] Fully documented
- [N] Entry in `whats-new.rst` 

Made small changes in phrasing to the newly split `cloud.rst` and `hub.rst` pages.  Also tried to add a table comparing the two options so that new users can easily choose what fits their needs.  The formatting of the table follows RST `list-table`, but it looks very bad for me--I'm editing in the browser due to unstable computer, so maybe that's why.  Happy to edit or remove it if it looks bad for you too.

These are very small doc changes - not sure a `whats-new.rst` entry is needed.


fmaussion:
Thanks  a lot! :raised_hands: 

This is much appreciated

PR #1203, opened 2021-02-22
Update test image

TimoRoth:
New image has a Python version with the latest security issue fixed.


fmaussion:
Thanks! One day it would be good to take care of the deprecation warnings where possible - Im quite sure some of them are in dependencies, not in OGGM itself

PR #1204, opened 2021-02-22
Numpy deprecation fixes

TimoRoth:
This is pretty much just a simple find&replace run for all the known deprecation warnings.
Some of the warnings are triggered by code in dependencies, and there's not much we can do about that other than wait for those dependencies to fix them.


TimoRoth:
This seems to have calmed down the spam quite a bit.
The majority of remaining warnings are caused by salem.

Might be worth reviewing of any of the np.int->int this replaces should be np.int32 or np.int64 instead.

fmaussion (approved):
Thanks a lot, that's very useful.

I think that I would have preferred to cast everything to `np.int64` and `np.float64` for the sake of being explicit, but that's the default behavior to cast `int` to `np.int64` and `float` to `np.float64` anyways, so we're safe

PR #1208, opened 2021-03-01
Add warning for monthly timesteps

sarah-hanus:
Add warning if `store_monthly_step == True` but `mb_elev_feedback = 'annual'` to make sure users don't expect output to reflect monthly processes if mass balance is only calculated annually. 

Changes are made on `flowline_model_run` because it is the function all different run options call and to which the tow parameters are passed.

### DIFF
--- a/oggm/core/flowline.py
+++ b/oggm/core/flowline.py
@@ -2128,6 +2128,12 @@ def flowline_model_run(gdir, output_filesuffix=None, mb_model=None,
     kwargs : dict
         kwargs to pass to the FluxBasedModel instance
      """
+    if 'mb_elev_feedback' in kwargs:
+        mb_elev_feedback = kwargs.get("mb_elev_feedback")
+    else:
+        mb_elev_feedback = 'annual'
+    if ((store_monthly_step == True) and (mb_elev_feedback == 'annual')):
+        warnings.warn("Mass balance is computed yearly. If you want output to reflect monthly processes set mb_elev_feedback = 'monthly'.")
 
     if cfg.PARAMS['use_inversion_params_for_run']:
         diag = gdir.get_diagnostics()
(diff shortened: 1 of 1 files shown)


oggm/core/flowline.py
```diff
@@ -2128,6 +2128,12 @@ def flowline_model_run(gdir, output_filesuffix=None, mb_model=None,
     kwargs : dict
         kwargs to pass to the FluxBasedModel instance
      """
+    if 'mb_elev_feedback' in kwargs:
+        mb_elev_feedback = kwargs.get("mb_elev_feedback")
+    else:
+        mb_elev_feedback = 'annual'
```
fmaussion, line 2134: The above can be simplified to:

```python
mb_elev_feedback = kwargs.get('mb_elev_feedback', 'annual')
```

oggm/core/flowline.py
```diff
@@ -2128,6 +2128,12 @@ def flowline_model_run(gdir, output_filesuffix=None, mb_model=None,
...
+    if ((store_monthly_step == True) and (mb_elev_feedback == 'annual')):
```
fmaussion, line 2135:
```suggestion
    if store_monthly_step and (mb_elev_feedback == 'annual'):
```

oggm/core/flowline.py
```diff
@@ -2128,6 +2128,12 @@ def flowline_model_run(gdir, output_filesuffix=None, mb_model=None,
...
+        warnings.warn("Mass balance is computed yearly. If you want output to reflect monthly processes set mb_elev_feedback = 'monthly'.")
```
fmaussion, line 2136: we try to keep the lines < 80 chars (max 100 chars if 80 is really ugly). You can split strings with:

```python
    warnings.warn("Mass balance is computed yearly. If you want output to "
                  "reflect monthly processes set mb_elev_feedback = 'monthly'")
```

fmaussion (commented):
Thanks a lot! Some minor formatting comments and this is good to go!

sarah-hanus:
Thanks Fabien for the feedback :) Helps me learning!

fmaussion:
Thanks for the contribution!

PR #1211, opened 2021-03-04
rename prcp_bias to prcp_fac and allow it to change in tasks

lilianschuster:
I have tried to rename everything that is called prcp_bias in OGGM  to prcp_fac as it is a multiplicative factor, not a bias and I added some documentation. 
 In addition:
- when calling `self.temp` in a PastMassbalance instance, the bias corrected temperature time series is shown and not the uncorrected as before
-  the precipitation_factor has been added as a keyword argument for tasks.run_random_climate/run_constant_climate...  
- in order to change precipitation factor or temperature bias one has to set the `check_calib_params` to False, otherwise an error occurs, if this gets finally merged, a change has to be done in the water resources notebook of oggm-edu that I will commit then (but if this does not make sense for the temperature bias I can remove it again)  

The tests were updated. All but one tests pass: there is still an issue in the UncertainMassbalance test, somehow it changes the basis_model but I did not find out how, I first thought these were just rounding errors but they are too large. I just don't understand what is happening there 

- [x] Tests added/passed

### DIFF
--- a/oggm/tests/test_models.py
+++ b/oggm/tests/test_models.py
@@ -310,8 +310,6 @@ class TestMassBalanceModels:
                                    atol=1e-2)
 
         mb_mod = massbalance.PastMassBalance(gdir)
-        # set that to False in order to allow temp_bias to change!
-        mb_mod.check_calib_params = False
         for yr in mbdf.index.values:
             my_mb_on_h = mb_mod.get_annual_mb(h, yr) * SEC_IN_YEAR * rho
             mbdf.loc[yr, 'MY_MB'] = np.average(my_mb_on_h, weights=w)
@@ -358,8 +356,7 @@ class TestMassBalanceModels:
         gdir = hef_gdir
         init_present_time_glacier(gdir)
 
-        mb_mod = massbalance.PastMassBalance(gdir, bias=0,
-                                             check_calib_params=False)
+        mb_mod = massbalance.PastMassBalance(gdir, bias=0)
         # save old precipitation/temperature time series
         prcp_old = mb_mod.prcp.copy()
         temp_old = mb_mod.temp.copy()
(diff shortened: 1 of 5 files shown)


oggm/tests/test_models.py
```diff
@@ -735,52 +827,68 @@ def test_uncertain_mb(self, hef_gdir):
...
+        # problem: UncertainMassBalance does something to ref_mod,
+        # it slightly changes the precipitation and temperature time series
+        # maybe there is a problem because of rounding issues???
+        # because I always update the time/prcp time series immediately...
+        # this might be the reason that the tests fail, but actually the tests
+        # fail more with larger _sigma, (see below)
+        # so there must be something in ref_mb that changes ref_mod stronger
+        # with a larger _sigma, the temp_bias and prcp_fac stay however the same
+        assert_allclose(temp_1, temp_2, rtol=1e-5)
+        assert_allclose(temp_1, temp_3, rtol=1e-5)
```
lilianschuster, line 863: I am totally confused what is happening here, @fmaussion  , do you have an idea what the problem is, it can't be only because of rounding issues?!

oggm/tests/test_models.py
```diff
@@ -388,6 +407,67 @@ def test_prcp_fac_update(self, hef_gdir):
...
         with pytest.raises(InvalidWorkflowError):
             mb_mod.prcp_fac = 10
+        with pytest.raises(InvalidWorkflowError):
+            mb_mod.temp_bias = 1
+
+        # normally this should also be implemented for mu_star ...
+        # with pytest.raises(InvalidWorkflowError):
+        #    mb_mod.mu_star = 100
+
+    def test_prcp_fac_temp_bias_update_constant_mb(self, hef_gdir):
```
lilianschuster, line 417: I just added this to check if it also works for ConstantMassBalance because I thought this could be the error for the UncertainMassBalance but it works normally, this can be also removed again

oggm/tests/test_models.py
fmaussion: yes rounding issues very likely. Whether it's acceptable or not depends on the test and the change in accuracy I guess, will have a look

fmaussion:
OK this looks very good to me, thanks a lot @lilianschuster . A few high level comments and I'll have a closer look at the failing tests after that:

- thanks for the warning in case people try to access `prcp_bias`. Does it also raise when set, i.e. when people do `mb.prcp_bias = value`? If not it might need a `@setter` as well.
- can you put this deprecation warning at the parent class level to avoid code duplication? 
- On second thoughts (sorry about that) I'm against the params check in the setter. I'll explain why:
    - you don't warned to get warned about something you do "willingly", i.e. you know that you are messing around with something when setting a property
    - to be able to set the property, you now need to set "check_calib" to false at class instantiation. But at instantiation, the check actually makes sense (we can still check that the default MB Model is well instantiated, it's not our problem if the things are changed afterwards) 
- can you make a quick assesment of what changed that causes to the floating rounding error? Is it datatype conversions (float32 data being used somewhere?) or something else? If you don't find it I'll have a closer look.

lilianschuster:
As suggested, I changed the following:
- prcp_bias is now at MassBalance level, however here I am unsure about which warning to show up, I removed the DeprecationWarning because when trying it out it was not visible for me, and when trying to set the value an error is raised now
- I removed all 'check_calib_params` (for temp_bias as well as prcp_fac)
- I am still unsure about the uncertaintyModel problem, what has changed is mainly that the former prcp_bias and temp_bias were only applied when doing `get_monthly_mb` or `get_annual_mb`. But now they are already applied when changing temp_bias or prcp_fac. That means that when this is done a lot of times there and back, the climate time series won't be the same because of rounding. This effect might be stronger for temperature because of the addition. Still, I think there has to be another problem than that one ...

fmaussion:
Thanks @lilianschuster !

Issue #1217, opened 2021-03-09
a method to store (external) mass balance parameters


lilianschuster:
as discussed with @fmaussion, it would be great to have a way to store the mass balance parameters of an external model (in my case e.g. melt_f and prcp_fac) inside of OGGM such that they are available inside of the output of model_diagnostics and model_run when running e.g. run_from_climate_data().

PR #1219, opened 2021-03-09
Add more options to run_until_and_store

fmaussion:
cc @lilianschuster - this stores scalar attrs of mb models without asking for permission. What do you think?

Closes https://github.com/OGGM/oggm/issues/1217

- [x] Tests added/passed
- [x] Fully documented


lilianschuster:
Thanks @fmaussion.  I will have a check when you have merged the pull request to OGGM or should I check it already now? If I understand that right, the mass balance attributes from an external module can be added with that to the output which would be things like melt_f and pf. So, I think it does what I want ;-)

Issue #1218, opened 2021-03-09
inconsistency in hydrodate <-> calendaryear definition for hydro_month = 1


lilianschuster:
I encountered a new inconsistency about the hydrodate <-> calendaryear conversion:
as far as I understood from the discussions together with @fmaussion , we define now that if hydro_month =1, then calendar year corresponds to hydro year. Therefore, I commited some changes in #1142 but apparently I missed the hydrodate, calendaryear conversion functions:
at the moment they output :
`hydrodate_to_calendardate(2020, 1, start_month = 1)` : -> 2019, 1

`calendardate_to_hydrodate(2020, 1, start_month = 1)` : -> 2021, 1

But to be consistent, I believe the output for both should be **2020, 1**!

Please confirm if I understood this right, then I can prepare a pull request where the functions are changed and check if the tests work, and do a new search for hidden functions that still use the old convention

fmaussion:
Yes I agree! Nice catch thanks

PR #1220, opened 2021-03-09
hydro year corresponds to calendar year if hydro month=1

lilianschuster:
hydrodate_to_calendardate  and calendardate_to_hydrodate are adapted such that hydro_year is equal to calendar_year. In addition, in order that the "time" coordinate is right, I  adapted run_from_climate_data. Tests have passed and I added one for hydromonth = 1. 

Closes https://github.com/OGGM/oggm/issues/1218

- [x] Tests added/passed

### DIFF
--- a/oggm/core/flowline.py
+++ b/oggm/core/flowline.py
@@ -2433,10 +2433,11 @@ def run_from_climate_data(gdir, ys=None, ye=None, min_ys=None, max_ys=None,
         except AttributeError:
             ys = gdir.rgi_date
         # The RGI timestamp is in calendar date - we convert to hydro date,
-        # i.e. 2003 becomes 2004 (so that we don't count the MB year 2003
-        # in the simulation)
+        # i.e. 2003 becomes 2004 if hydro_month is not 1 (January)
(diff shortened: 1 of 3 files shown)


oggm/core/flowline.py
```diff
@@ -2433,10 +2433,11 @@ def run_from_climate_data(gdir, ys=None, ye=None, min_ys=None, max_ys=None,
         except AttributeError:
             ys = gdir.rgi_date
         # The RGI timestamp is in calendar date - we convert to hydro date,
-        # i.e. 2003 becomes 2004 (so that we don't count the MB year 2003
-        # in the simulation)
+        # i.e. 2003 becomes 2004 if hydro_month is not 1 (January)
```
fmaussion, line 2436: This is a bit arguable - in the alps for example the image is from Aug 2003 -> is it better to start from Jan 2003 or Jan 2004? Although it's not super important, I'd argue for 2004...

lilianschuster: ok, I understand. Then ys+=1 makes  sense independent of the hydro date conversion and I will add your comment just to make that clear

Issue #1199, opened 2021-02-18
Store liquid precipitation in OGGM


sarah-hanus:
Until now, OGGM just "throws away" liquid precipitation. For using OGGM for hydrological purposes, we need that liquid precipitation is stored, in order to use it together with volume change as runoff input to the hydrological model. 
glacier runoff = volume change + liquid precipitation
Ideas how to implement it:

-  `get_monthly_climate`: store prcpliquid for each elevation band and timestep instead of only storing prcpsol

Simple implementation: 
- transfer the prcpliquid stored per elevation band to prcpliquid per glacier using the areas of each elevation band of a fixed glacier geometry (e.g. RGI geometries). This implementation would mean that rain output is always fixed to a constant glacier area and does not account for retreating of glaciers.

More complicated implementation:
- incorporate calculation of prcpliquid per glacier in the `FluxBasedModel` and calculate total liquid precipitation at each timestep based on glacier area at each elevation band and timestep (somehow similar to how `total_mass` is calculated in `FluxBasedModel` now). This implementation would allow the precipitation output to change with changing glacier area (e.g. retreating glaciers), so that only rainfall on the actual glacier area is computed.

Any other ideas, suggestions?

sarah-hanus:
It would be good to store liquid precipitation as a separate variable from glacier melt. So to have in the output of an OGGM run besides `volume` also a variable `liquid_prcp`. It would be good to separate it, so that when using it for hydrological applications we are still able to say something about how much glacier melt contributes to runoff (so not taking rain on glaciers as glacier melt).

fmaussion:
Agreed! Will try to work on this by the end of the week and iterate with you. 

(I'd prefer to have a look myself here first, because it has consequences on the model structure - which module is in charge of storing what - which are not trivial)

Holmgren825:
This will be relevant for my thesis (#1114). Still early in the process, but commenting to stay updated.

fmaussion:
See https://github.com/OGGM/oggm/pull/1224 

We should discuss this together (also how to implement the TODOs) either at the next OGGM meeting or in a separate call.

sarah-hanus:
Great Fabi! I will have a look at it. Larissa told me that there are other people interested as well in using OGGM output as runoff input into other hydrological models. So I think it would be valuable to at least shortly discuss your progess during the next meeting and then maybe plan a separate call if needed.

PR #1224, opened 2021-03-15
First attempt at adding hydro output

fmaussion:
This is a first attempt at https://github.com/OGGM/oggm/issues/1199 

It took some time because its not so trivial - but I'm quite happy with the result.

This is currently slow because it does not happen during the run but after it - there is a list of TODOs and improvements for the near future.

- [x] Tests added/passed
- [ ] Fully documented


fmaussion:
@TimoRoth any idea what could cause the failing test in 3.6? Maybe we should remove 3.6 tests altogether?

PR #1232, opened 2021-03-24
Implement monthly time steps in run_with_hydro

sarah-hanus:
This is a attempt to improve #1224 by making it possible to store hydrological output monthly. 
This is a bit more tricky than expected.

It does not work yet because it has some issues:
1)updating existing model dianostics in line 2791 does not work if model is only run at yearly timesteps: so `store_monthly_step= True` should be used when monthly hydro output is wanted

2) FileModel does not work correctly if `store_monthly_step= True` (line 2616), it still gives only monthly output, so `fmod.months` is always 1, therefore also `fmod.run_until` does not work on monthly timescales. This is however necessary to compute `model_area` and `model_area` at monthly timescales to correct melt on glacier with residual mass balance.

I think it makes sense to be ale to use `run_with_hydro` also with `'store_monthly_step= True`  and `mb_elev_feedback='monthly'`. I am not sure why FileModel does not output monthly timesteps, although the netcdf file that is used as input in FileModel is in monthly timesteps.

Any suggestions are welcome

### DIFF
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -25,7 +25,7 @@ Enhancements
 ~~~~~~~~~~~~
 
 - Mass-balance models now properly refer to ``prcp_fac`` (was incorrectly named
-  ``prcp+bias``) (:pull:`1211`).
+  ``prcp_bias``) (:pull:`1211`).
   Additionally, the ``run_*`` tasks in ``oggm.core.flowline`` can now also adjust
   the precipitation factor for sensitivity experiments.
   By `Lilian Schuster <https://github.com/lilianschuster>`_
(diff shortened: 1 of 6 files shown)


fmaussion:
Thanks @sarah-hanus ! Will have a look soon.

> updating existing model dianostics in line 2791 does not work if model is only run at yearly timesteps: so store_monthly_step= True should be used when monthly hydro output is wanted

yes sorry about that, my mistake. 

> mb_elev_feedback='monthly'

I'm really not sure about that. I think it is not advisable to run the dynamical model this way

>  I am not sure why FileModel does not output monthly timesteps, although the netcdf file that is used as input in FileModel is in monthly timesteps.

I'm unsure either, will have a look.

fmaussion:
Still a TODO: add support for constant climate (currently wrongly returning average annual climate)

fmaussion:
Test commitment run on Hintereisferner:

image


image

sarah-hanus:
Thanks Fabien! I could reproduce the results of your test run! I also ran all glaciers of the Rhine basin (796 glaciers) for 200 years on the JupyterHub. It is not super fast (took like 17 minutes to run) but it works fine :) It's also nice to compare the input data (rainfall vs snowfall over the years). I attach a figure
precipitation_from_glacier_60_80
precipitation_from_glacier_00_20

fmaussion:
That's great, thanks @sarah-hanus !

PR #1234, opened 2021-04-01
small hydro_month correction for extend_past_climate_run

lilianschuster:
calendar year = hydro year for extend_past_climate_run() if hydro_month = 1

- [x ] Tests passed


fmaussion:
thanks!

PR #1239, opened 2021-04-22
Add checksum file lifetime

TimoRoth:
Simply checks and updates the files modification time and updates it when checked for validity.


oggm/utils/_downloads.py
```diff
@@ -246,7 +253,7 @@ def verify_file():
 
         logger.info('Done downloading.')
 
-        verify_file()
+        verify_file(True)
```
fmaussion, line 256:
```suggestion
        verify_file(force=True)
```

More explicit.

fmaussion:
Thanks! Neat.

PR #1243, opened 2021-05-02
Update URL testing for earthdata.nasa.gov credentials

oberrauch:
Found this by chance while updating my credentials, the URL must have changed.


fmaussion:
Nice catch! I think that the other urls will also need an update, e.g. here then: 

https://github.com/OGGM/oggm/blob/7cfc96c1c0c4c3951cb761c81b28f9a0caea4187/oggm/utils/_downloads.py#L1095

- also on cluster then unfortunately. Will have a closer look later.

PR #1245, opened 2021-05-10
Reduce progress spam

TimoRoth:
It should only print one progress-line every 15 seconds with this, while still staying fully interactive in actual terminal sessions.


fmaussion:
Nice, thanks!

PR #1248, opened 2021-05-12
new `workflow.match_geodetic_mb_for_selection` function to match the MB bias for any selection of glaciers

pat-schmitt:
adapted function match_regional_geodatic_mb, things to discusse:
- should hydro month be set to 1 for OGGM SMB, because the data starts in January
- should only the geodetic data be used where also a OGGM SMB is available (I think so, but I added a comment which can be deleted afterwards)
- what is the right way (loggging and/or error) if no geodetic data is available for the current glacier selections
- until now the current options are still available, but I think with the new file (with decade 1 and 2 and 20 years) this could also be deleted (maybe only the current 'hugonnet' option and keep the 'zemp' option)

- [x] Tests added/passed
- [x] Fully documented

### DIFF
--- a/oggm/workflow.py
+++ b/oggm/workflow.py
@@ -843,7 +843,9 @@ def match_geodetic_mb_for_selection(gdirs, period='2000-01-01_2020-01-01',
     """Shift the mass-balance residual to match geodetic mb observations.
 
     This method finds the "best mass-balance residual" to match all glaciers in
-    gdirs with geodetic mass-balance measurements from Hugonnet 2021.
+    gdirs with available OGGM mass balance and available geodetic mass-balance
+    measurements from Hugonnet 2021. It is possible to provide your own
+    geodetic mass-balance measurements by defining the 'file_path'.
 
     Parameters
     ----------
@@ -854,7 +856,14 @@ def match_geodetic_mb_for_selection(gdirs, period='2000-01-01_2020-01-01',
        '2000-01-01_2010-01-01',
        '2010-01-01_2020-01-01'.
     file_path: str
-       local file path, depricated if file uploaded online
+       local file path to csv file containing geodetic measurements, file must
+       contain the columns:
+           - 'rgiid': is the RGIId as in the RGI 6.0
+           - 'period': time intervall of the measurements in the format shown
+             above
+           - 'dmdtda': the specific-mass change rate in meters water-equivalent
+             per year,
+           - 'area': is the glacier area (same as in RGI 6.0) in meters square
     """
 
     # Get the mass-balance OGGM would give out of the box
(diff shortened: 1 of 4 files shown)


fmaussion:
> should hydro month be set to 1 for OGGM SMB, because the data starts in January

No, let's just ignore that at this level - on decadal scales the few months won't make a big difference

> should only the geodetic data be used where also a OGGM SMB is available (I think so, but I added a comment which can be deleted afterwards)

Yes: compute the correction on the largest subset of glaciers with data on both sides. Then, apply the correction to the residual for all glaciers (also those which are not in Hugonnet)

> what is the right way (loggging and/or error) if no geodetic data is available for the current glacier selections

Raise an `InvalidWorkflowError` if there are no glaciers at all.

> until now the current options are still available, but I think with the new file (with decade 1 and 2 and 20 years) this could also be deleted (maybe only the current 'hugonnet' option and keep the 'zemp' option)

Yes I'll let you know when the file is online

oggm/workflow.py
```diff
@@ -745,8 +745,9 @@ def to_minimize(x):
 
 
 @global_task(log)
-def match_regional_geodetic_mb(gdirs, rgi_reg=None, dataset='hugonnet',
-                               period='2000-01-01_2020-01-01'):
+def match_regional_geodetic_mb(gdirs, rgi_reg=None, dataset='hugonnet_reg',
```
fmaussion, line 748: Please make a new function for that and keep the old one!

pat-schmitt:
ok, now I think only the base_url must be included once the file is online and the local file_path can be deleted

pat-schmitt:
I am just realized that I currently not using only the set of glaciers where both (OGGM SMB and geodetic mb) are available, instead i first exclude all glacier with no OGGM SMB and use all the remaining for the calculation of the total OGGM specific mb. In the next step I get the geodetic mb for all OGGM SMB glaciers and use _all_ _available_ data to calculate the geodetic specific mb (I print a message for how many OGGM SMB glaciers a geodetic mb is available, and the relative unused glacier area).
I could change this in the next days to only use glaciers where both are available...

fmaussion:
OK! Thanks a lot Patrick - I'm not sure to understand but yes please update the code so that the largest subset of glaciers with both OGGM and Ref MB data is used, at your earliest convenience (not urgent). 

Otherwise, I just uploaded the data to the oggm servers. In the code please use something like

```python
from oggm import utils
import pandas as pd

df = pd.read_hdf(utils.file_downloader('https://cluster.klima.uni-bremen.de/~oggm/geodetic_ref_mb/hugonnet_2021_ds_rgi60_pergla_rates_10_20_worldwide.hdf'))
```

pat-schmitt:
Now the largest subset is used for the calculation of the MB shift. 
And I added a possibility to use custom geodetic measurements by providing a local `file_path`. But this maybe could be erased if not needed...

pat-schmitt:
And I included the data download from the URL

fmaussion:
Thanks @pat-schmitt !

PR #1251, opened 2021-05-26
Add `cook_rgidf()` to pretend non-RGI with RGI format

Keeptg:
I add a function `cook_rgidf()` in `utils._funcs` to make it easier of using a non-RGI glacier inventory in OGGM.
Will add a tutorial to show how it can be used in future!

- [x] Tests added/passed
- [x] Fully documented

### DIFF
--- a/oggm/tests/test_utils.py
+++ b/oggm/tests/test_utils.py
@@ -331,7 +331,7 @@ class TestFuncs(unittest.TestCase):
         cfg.PATHS['working_dir'] = utils.mkdir(working_dir, reset=True)
         path = utils.get_demo_file('cgi2.shp')
         cgidf = gpd.read_file(path)
-        rgidf = utils.cook_rgidf(cgidf,
+        rgidf = utils.cook_rgidf(cgidf, region='13',
                                  save_special_columns={'Glc_Long': 'CenLon',
                                                        'Glc_Lati': 'CenLat'})
         rgidf['Area'] = cgidf.Glc_Area * 1e-6
--- a/oggm/tests/test_utils.py
+++ b/oggm/tests/test_utils.py
@@ -323,6 +323,22 @@ class TestFuncs(unittest.TestCase):
         t2 = timeit.timeit('utils.clip_min(a, 15)', number=n, setup=s2)
         assert t2 < t1
 
+    def test_cook_rgidf(self):
+        from oggm import workflow
+        cfg.initialize()
+        cfg.PARAMS['use_intersects'] = False
+        cfg.PATHS['working_dir'] = utils.get_temp_dir(dirname='test_cook_rgidf')
+        path = utils.get_demo_file('cgi2.shp')
+        cgidf = gpd.read_file(path)
+        rgidf = utils.cook_rgidf(cgidf, save_special_columns={'Glc_Long': 'CenLon', 'Glc_Lati': 'CenLat'},
+                                 assign_col_values={'Area': cgidf.Glc_Area*1e-6})
+        gdirs = workflow.init_glacier_directories(rgidf)
+        workflow.gis_prepro_tasks(gdirs)
+        workflow.download_ref_tstars()
+        workflow.climate_tasks(gdirs)
+        workflow.inversion_tasks(gdirs)
+        df = utils.compile_glacier_statistics(gdirs, inversion_only=True)
+        assert np.any(np.isnan(df.inv_vol_km3))
 
 class TestInitialize(unittest.TestCase):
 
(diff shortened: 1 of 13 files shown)


Keeptg:
Not sure what's the reason the test program failed. Any suggestions wil be appriciated!

fmaussion:
The tests fail because the test requires to download input data which is not in the test data, a situation we try to avoid. I think we don't need to test that everything works in your tests, let me have a look at your code later today.

oggm/tests/test_utils.py
```diff
@@ -323,6 +323,26 @@ def test_clips_performance(self):
...
+        cfg.initialize()
+        cfg.PARAMS['use_intersects'] = False
+        cfg.PATHS['working_dir'] = utils.mkdir(utils.get_temp_dir(dirname='test_cook_rgidf'),
+                                               reset=True)
+        path = utils.get_demo_file('cgi2.shp')
+        cgidf = gpd.read_file(path)
+        rgidf = utils.cook_rgidf(cgidf, save_special_columns={'Glc_Long': 'CenLon',
+                                                              'Glc_Lati': 'CenLat'},
+                                 assign_col_values={'Area': cgidf.Glc_Area*1e-6})
+        gdirs = workflow.init_glacier_directories(rgidf)
```
fmaussion, line 337: Yes so I think you need to stop the test here, and ensure that OGGM could still create the glacier directories properly (that's the main test anyway)

oggm/utils/_funcs.py
```diff
@@ -780,3 +781,96 @@ def shape_factor_adhikari(widths, heights, is_rectangular):
...
+               assign_col_values=None):
+    """ Cook the user's glacier inventory with the RGI format
+
+    Parameters
+    ----------
+    gi_gdf : :py:geopandas.GeoDataFrame
+        the GeoDataFrame of the user's glacier inventory
+    region : str
+        Glacier RGI region code, only usable when set `confirm_region=False`.
+        The default is '13'.
```
fmaussion, line 797: There should be no default here - it should fail if the region is not provided and cannot be inferred automatically

oggm/utils/_funcs.py
```diff
@@ -780,3 +781,96 @@ def shape_factor_adhikari(widths, heights, is_rectangular):
...
+    version : str
+        Glacier inventory version code. The default is '60'.
```
fmaussion, line 799: Explain that this is necessary to generate the RGI Ids

oggm/utils/_funcs.py
```diff
@@ -780,3 +781,96 @@ def shape_factor_adhikari(widths, heights, is_rectangular):
...
+            'Area': -9999, 'Zmin': -9999, 'Zmax': -9999, 'Zmed': -9999, 'Slope': -9999,
+            'Aspect': -9999, 'Lmax': -9999, 'Status': 0, 'Connect': 0, 'Form': 0,
+            'TermType': 0, 'Surging': 0, 'Linkages': 1, 'check_geom': None, 'Name': ''}
+
+    # Construct the output GeoDataFrame
+    cooked_rgidf = gpd.GeoDataFrame(data=data, geometry=geom, crs='epsg:4326')
+
+    # If there are specifical column in the original glacier inventory we want to keep
+    if save_special_columns is not None:
+        for col in save_special_columns.keys():
```
fmaussion, line 868: loop like:
`for key, val in dict.items()`

oggm/utils/_funcs.py
```diff
@@ -780,3 +781,96 @@ def shape_factor_adhikari(widths, heights, is_rectangular):
...
+        Glacier RGI region code, only usable when set `confirm_region=False`.
+        The default is '13'.
+    version : str
+        Glacier inventory version code. The default is '60'.
+    ids : list of integer, each element should be a str of length 5
+        Assign special id number to each glacier. The default is None, the id code
+            following the glacier order.
+    id_suffix : str or None
+         Add a suffix to the glacier id. The default is None, no suffix
+    save_special_columns : dict or None
```
fmaussion, line 805: could this be only one argument (`assign_col_values`) where the right column name in the RGI format is already used? The users can change the name themselves if they want to, I'm not sure this dict here is very useful.

fmaussion (changes requested):
Thanks! This is looking good. Some minor suggestions to the names, otherwise this looks good!

oggm/utils/_funcs.py
Keeptg: OK! `assign_col_values` argument was deleted!

oggm/utils/_funcs.py
Keeptg: Thanks! Didn't know this before.

Keeptg:
There are still two error reported.
Don't know how to deal with them......

fmaussion (approved):

fmaussion:
Thanks @Keeptg ! sorry this took so long, excellent work!

Issue #1256, opened 2021-06-02
Add mean ice velocity to glacier statistics


Keeptg:
I think maybe add the mean ice velocity and the avilabel ice velocity glacier area (in percent) to `glacier_statistics()` wil be useful.
Will glad to submit a PR!

fmaussion:
Do you mean of the ITS_LIVE velocities? I think this is going a bit too far for OGGM itself in terms of functionality.

However, what I think would be good is to allow users to define an arbitrary function to be run by `compile_glacier_statistics` that would allow the user to add any variable they want to the output! This would be quite elegant...

Keeptg:
> Do you mean of the ITS_LIVE velocities? I think this is going a bit too far for OGGM itself in terms of functionality.

Yes, I means ITS_LIVE velocities. What I thought at the beginning is statistics the mean ice velocity if the `obs_icevel_x` and `obs_icevel_y` variables is writen in `gridded_data.nc` files.

> what I think would be good is to allow users to define an arbitrary function

Do you means add a key argument like `statis_func` to pass in custom-defined function in `compile_glacier_statistics`?

fmaussion:
> Do you means add a key argument like statis_func to pass in custom-defined function in compile_glacier_statistics?

Yes, maybe rather `apply_func`, which accepts a gdir and dict as input and simply fills the dict with any data (only scalar data though)

PR #1259, opened 2021-06-04
Add apply_func argument in utils.compile_glacier_statistics()

Keeptg:
Closes https://github.com/OGGM/oggm/issues/1256

- [x] Tests added/passed
- [x] Fully documented

### DIFF
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -16,6 +16,9 @@ Enhancements
 - Added ``cook_rgidf()`` function in ``oggm.utils`` to simplify the use
   of a non-RGI glacier inventory in OGGM (:pull:`1251`).
   By `Li Fei <https://github.com/Keeptg>`_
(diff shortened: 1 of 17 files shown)


Keeptg:
Again, some failed test that I don't understand :-(

fmaussion:
Thanks @Keeptg ! 

The error comes from the fact that you were using the function for multiprocessing use, and you can't pickle a function which is in the main namespace like that. 

Anyways, I've simplified your code a bit for what I had in mind with the function. I did not find the kwarg useful, because the user will always apply only one function of his / her choice. I think that in the future I'd like to add more options to this global task, but for now this will get the job done.

Keeptg:
> The error comes from the fact that you were using the function for multiprocessing use, and you can't pickle a function which is in the main namespace like that.

Thanks for the detailed explain!

> I did not find the kwarg useful, because the user will always apply only one function of his / her choice. 

Yes, you're right! I think I should think more carefully before choosing an argument!

PR #1270, opened 2021-07-16
Added run_with_hydro_daily function

sarah-hanus:
`run_with_hydro_daily` function is added which is similar to `run_with_hydro` function but is able to output fluxes on daily timescale. For this a mass balance model which has a function `get_daily_mb `is necessary which is available in TIModel of massbalance sandbox (https://github.com/OGGM/massbalance-sandbox) and not in OGGM core.

compile_run_output had to be adjusted to compile the output of the run_with_hydro_daily function in the format (years,days).

This commit relates to issue #1199


fmaussion:
Thank! I think that for now this should live outside of the OGGM namespace, as long as @lilianschuster 's TIModel isn't in OGGM proper... I'll be keen to take this into OGGM in one go with the new MB models later this year!

PR #1278, opened 2021-08-13
Update test images

TimoRoth:
I'd expect those tests to still fail.
Opening this to have something tracking that issue.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -24,9 +24,9 @@ jobs:
           - workflow-multi
           - graphics-mpl
         container:
-          - oggm/untested_base:20210221
-          - oggm/untested_base:py36
+          - oggm/untested_base:20210808
           - oggm/untested_base:py37
+          - oggm/untested_base:py39
         include:
           - container: python:3.8-slim
             test-env: minimal


fmaussion:
@TimoRoth can you rebase and check again? Tests are green on master since https://github.com/OGGM/oggm/pull/1279

TimoRoth:
You can just hit Re-Run on the checks, it should always run against current base branch with PR merged into it.

TimoRoth:
I think the error changed, did it? Don't see it complaining about lon/lat anymore at least.

fmaussion:
> You can just hit Re-Run on the checks, it should always run against current base branch with PR merged into it.

That's what I thought as well but the fixes I merged are still occurring.

fmaussion:
For example, `KeyError: 'no index found for coordinate longitude'` happens because a `ValueError` became a `KeyError` in xarray, and is fixed in https://github.com/OGGM/oggm/pull/1279/files

TimoRoth:
Makes me wonder if the PR Action needs fixed. It should be tested the merged PR.

fmaussion:
> Makes me wonder if the PR Action needs fixed. It should be tested the merged PR.

Yes, but this is only when clicking on the button "re-run all tests" - repush worked fine.

TimoRoth:
On 16.08.2021 10:16, Fabien Maussion wrote:
>     Makes me wonder if the PR Action needs fixed. It should be tested
>     the merged PR.
> 
> Yes, but this is only when clicking on the button "re-run all tests" - 
> repush worked fine.

Well, it was a rebase on the latest changes. So it seems like the tests 
run right on the PR branch, and not on the merged base.

PR #1281, opened 2021-08-16
Fix isdir TypeError when passing it a boolean

TimoRoth:
This has recently started exploding on the latest Python versions, because os.path.isdir now throws a TypeError if it does not get passed a string/path-like object.

### DIFF
--- a/oggm/workflow.py
+++ b/oggm/workflow.py
@@ -385,6 +385,15 @@ def init_glacier_regions(rgidf=None, *, reset=False, force=False,
     return gdirs
 
 
+def _isdir(path):
+    """os.path.isdir, returning False instead of an error on non-string/path-like objects
+    """
+    try:
+        return os.path.isdir(path)
+    except TypeError:
+        return False
+
+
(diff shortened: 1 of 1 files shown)


oggm/workflow.py
```diff
@@ -385,6 +385,13 @@ def init_glacier_regions(rgidf=None, *, reset=False, force=False,
     return gdirs
 
 
+def _isdir_wrapper(path):
```
fmaussion, line 388:
```suggestion
def _isdir(path):
"""os.path.isdir, returning False instead of an error on non-string/path-like objects
"""
```

oggm/workflow.py
```diff
@@ -508,7 +515,7 @@ def init_glacier_directories(rgidf=None, *, reset=False, force=False,
                     # List of str
                     pass
 
-            if os.path.isdir(from_tar):
+            if _isdir_wrapper(from_tar):
```
fmaussion, line 518:
```suggestion
            if _isdir(from_tar):
```

PR #1285, opened 2021-09-05
Move cfg.DATA dict to be shared accross processes

fmaussion:
This will be needed for later use (e.g. @lilianschuster 's MB models). I'm a bit wary of this change, let's hope this won't have too many impacts.

### DIFF
--- a/oggm/cfg.py
+++ b/oggm/cfg.py
@@ -46,6 +46,8 @@ CONFIG_MODIFIED = False
 # Share state accross processes
 DL_VERIFIED = dict()
 DEM_SOURCE_TABLE = dict()
+DATA = dict()
+LRUHANDLERS = dict()
 
 # Machine epsilon
 FLOAT_EPS = np.finfo(float).eps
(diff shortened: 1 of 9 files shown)


oggm/cfg.py
```diff
@@ -857,9 +857,19 @@ def set_manager(manager):
...
+        DATA = new_dict
+
+        new_dict = manager.dict()
+        new_dict.update(LRUHANDLERS)
+        LRUHANDLERS = new_dict
     else:
         DL_VERIFIED = dict(DL_VERIFIED)
         DEM_SOURCE_TABLE = dict(DEM_SOURCE_TABLE)
+        DATA = dict(DATA)
+        LRUHANDLERS = dict(LRUHANDLERS)
```
fmaussion, line 872: @TimoRoth what was the purpose of this re-cast to a dict? They all are dict per definition....

fmaussion: Ah I think I know.

fmaussion:
Yeah OK so I learned something about the mp manager (which I kind of knew already). Objects *stored* in the dict of course can't be modified by the processes, they are meant to be read-only.

Anyways, the LRU handler dict is one of the very useless things I wrote. I think that currently, each process creates its own (which is not bad, but a lot of boilerplate code is useless).

oggm/cfg.py
TimoRoth: To get rid of the manager, and only keep the data.

PR #1297, opened 2021-09-22
Extend execute_entity_task capabilities

TimoRoth:
This adds the ability to pass a list of tasks to execute_entity_task, which will then execute all tasks in that list in sequence on each gdir, potentially more efficiently distributing the work, avoiding CPU downtime.

Also documents some abilities the function had for a long time about per-gdir kwargs.

### DIFF
--- a/oggm/tests/test_utils.py
+++ b/oggm/tests/test_utils.py
@@ -499,10 +499,8 @@ class TestStartFromTar(unittest.TestCase):
 
         # Go - initialize working directories
         gdirs = workflow.init_glacier_directories(self.rgidf)
-        workflow.execute_entity_task(tasks.define_glacier_region, gdirs)
-
-        # End - compress all
-        workflow.execute_entity_task(utils.gdir_to_tar, gdirs)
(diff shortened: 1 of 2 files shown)


TimoRoth:
Some performance comparisons:

Old method:
slurm-9698.out:real     18m34.619s
slurm-9753.out:real     18m56.377s
slurm-9754.out:real     18m23.709s
slurm-9755.out:real     18m35.420s
slurm-9756.out:real     18m51.713s
slurm-9757.out:real     18m26.689s

New method:
slurm-9699.out:real     18m18.392s
slurm-9759.out:real     18m13.198s
slurm-9760.out:real     18m17.620s
slurm-9761.out:real     18m35.503s
slurm-9762.out:real     18m17.189s
slurm-9763.out:real     18m26.404s

So there is a small performance benefit for my cobbled together test case. And more importantly: Not a performance regression.

fmaussion:
Cheers! I'll try to think of this when updating the notebooks in future OGGM versions

PR #1308, opened 2021-10-08
New flowline diagnostics outputs and new signature for `run_until_and_store()` 

fmaussion:
This PR will change the signature of the `run_until_and_store()` method in a non backwards compatible way. 

- the new default is to simply output the full glacier diagnostics only (volume, length, etc). Previously, it would also output what we call "model geometry files"
- "model geometry" files are a good idea, they are kind of "restart" files. Indeed, their main purpose is to be able to restart from any modeled state with the `FileModel`. Other things (such as retrieving diagnostics) is slow and tedious and is now better done with the new diagnostic file. With this change, they are now not written per default, i.e. users will need to set this explicitly if they need them e.g. for spinup runs (its only a global option change).
- I added a new type of optional output (`store_fl_diagnostics`) which provide diagnostics along the flowlines (instead of aggregated at the glacier level). These files can then be used to retrieve elevation, widths, velocity, etc. for all grid points along the flowlines.

Some work is needed to document this.

### DIFF
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -13,11 +13,16 @@ Breaking changes
 - In the process of adding new output diagnostic files (:pull:`1308`), the
   signature and return output of ``FlowlineModel.run_until_and_store``
   changed. We hope that this change won't affect too many of our users but
-  if it does, it should be relatively straightforward to update your code.
+  if it does, it should be relatively straightforward to update your code:
+  users now control the number of outputs with the ``fl_diag_path`` and
+  ``geom_path`` kwargs. Most users will probably have used the ``run_*``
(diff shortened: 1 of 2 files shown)


nchampollion:
Hi Fabien, just a quick question. What do you mean by "to store model geometry" ? You mean to store each year the widths at each flowline points retrieved using thickness and bed shape

fmaussion:
> What do you mean by "to store model geometry" ?

Yes, the model geometry files that can be used to retrieve full glacier geometry using the `FileModel`. Almost nobody uses it. In fact I would like to change the full signature of the `run_until_and_store` method to return more useful things, but I'm afraid it will break too many things and I have to consider it a bit more...

PR #1332, opened 2021-11-03
stop for loop if mustar candidate found in `mu_star_calibration_from_geodetic_mb`

lilianschuster:
Inside of `mu_star_calibration_from_geodetic_mb`, there is a loop to get a `mu_star` and `step` (ref_hgt_change) combination. I changed the function in order that the loop actually stops when one suitable combination is found (i.e., the one where we can find a `mu_star` with the smallest ref_hgt_change possible. Because for this application, we only take the first combination, so stopping the loop just saves hopefully some time!  

- [x] Tests added/passed


fmaussion:
Thanks @lilianschuster !

PR #1339, opened 2021-11-15
Switch Docker-Image to ghcr.io

TimoRoth:
Since Docker Hub is putting more and more limits on their free tier, moving to somewhere else has become almost inevitable.
All new base Images will be on GitHubs registry, starting today.

This PR switches over the main OGGM image, as well as the tests to the new Pyston-Based main image, all also on GitHubs registry.

Also refactored and modernized the CI Workflow a bit while at it.


fmaussion (approved):
Thanks Timo!

PR #1342, opened 2021-11-24
added dynamic_spinup function

pat-schmitt:
- [x] Tests added/passed
- [ ] Fully documented

General idea: 
- conduct a spinup from 1980 to the RGI date and match area OR volume at RGI date (e.g. 2003)
- Timeline of spinup: START from 'RGI date' `flowlines` -> run model with `ConstantMassBalance` (defined between 1980 and 2003) with a temperature bias `t_bias` for 23 years (2003 - 1980) -> resulting `flowlines` are defined as glacier state in year 1980 -> run model with `PastMassBalance` until 2003 -> compare resulting area OR volume with RGI date area OR volume -> change `t_bias` and start with next iteration at START
- all model settings are the default ones of prepro level 4

Things to discuss:
- general idea
- Spinup tries to match area OR volume at RGI date. For volume: the function is intended to be used after the volumes of the glaciers are calibrated to the consensus estimate on a regional scale (e.g. starting from glaciers at prepro level 4). Maybe the difference between the resulting spinuped glaciers and the consensus estimate should be checked again after the spinup on a regional scale.
- should the resulting temperature bias and mismatch be stored somewhere (maybe in `model_diagnostics`)
- are the included tests sufficient

### DIFF
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -73,6 +73,7 @@ Enhancements
   historical to projections outputs (if needed).
   By `Fabien Maussion <https://github.com/fmaussion>`_
 - The package "descartes" is no longer a required dependency for oggm.
+- Added a new entity task ``run_dynamic_spinup``. This task dynamically spinup the glacier to match the area or volume at the RGI date. To do so the glacier is simulated from the recent past (default 1980) to the RGI date. The unknown glacier geometry at the start of the simulation is iteratively changed with a short constant climate run with a varying temperature bias (:pull:`1342`). By `Patrick Schmitt <https://github.com/pat-schmitt/>`_
 
 Bug fixes
 ~~~~~~~~~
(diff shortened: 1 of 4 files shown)


fmaussion:
Excellent work @pat-schmitt , this is really really nice.

- your points about volume are well taken, but it does not matter much I think whether this has been calibrated (at least not for your code). It's a general documentation issue, we should be clear about that indeed.
- temperature bias and mismatch (and other relevant parameters) can be stored e.g. with `gdir.add_to_diagnostics`, this way they will automatically also be added to the glacier_statistics mega-tables 
- the tests looks sufficient to me, maybe you can have a look at the [test coverage](https://coveralls.io/repos/75824/builds) if everything you want is covered.

Can you add an entry to what's new? The next step now will be to add this as option to the preprocessing cli and test it "large scale" -> let's talk about it tomorrow!

fmaussion:
Thanks a lot @pat-schmitt ! Sorry I forgot about it.

PR #1348, opened 2021-12-04
Fix ERA5L-HMA support

Keeptg:
- [x] Tests added/passed
- [ ] Fully documented


fmaussion:
Thanks @Keeptg ! I think its fine here to only change `oggm/core/climate.py ` and do not test for it.

The tests do not allow downloads from such large data, so it would imply creating test files, etc. - which is too tedious I think.

Keeptg:
@fmaussion Aha, then it will be quite easy. Will update the PR soon!

Keeptg:
@fmaussion Aha, then it will be quite easy. Will update the PR soon!

fmaussion:
Thanks!

PR #1351, opened 2021-12-14
corrected mb_climate_on_height for hydro month = 1 and qc as diagnostics

lilianschuster:
when using `match_geod_pergla`, so far, the MB mean from 1999 to end of 2019 was calibrated to match the geodetic estimates. Normally it should be 2000-2019, but `mb_climate_on_height` falsely used the year before y0 even if hydro_month is equal to 1. This resulted in a not perfectly calibrated mb model and was found out during the development of this notebook:  https://nbviewer.org/urls/cluster.klima.uni-bremen.de/~lschuster/error_analysis/working_glacier_gdirs_comparison.ipynb#id-dmdt-dmdtda-qc3. I proposed a correction for that. All old tests of OGGM still pass with this correction.  If I am right and the corrections have to be done, we need to recalibrate the preprocessed gdirs above level 2 that use `match_geod_pergla` ! 

I also wrote a test that checks if the 20-year modelled specific mass-balance is now equal to the reference geodetic calibration data. Now it perfectly matches, however, and this is also tested inside of OGGM:
- For the test, we need the newest climate data (till end of 2019), however inside of https://cluster.klima.uni-bremen.de/~oggm/test_climate/, for both ERA5 and CRU, the climate goes only till 2018 or 2014, 
     - Therefore, I use  `process_dummy_cru_file` to create a climate in the right time range and to test then if the calibration has worked
     - I will create new pre-processed directories without the small error, for that I also added the climate_qc_months to the diagnostics if historical_qc_months is called!

### DIFF
--- a/oggm/core/climate.py
+++ b/oggm/core/climate.py
@@ -409,6 +409,8 @@ def historical_climate_qc(gdir):
     default_grad = cfg.PARAMS['temp_default_gradient']
     g_minmax = cfg.PARAMS['temp_local_gradient_bounds']
     qc_months = cfg.PARAMS['climate_qc_months']
+    # add that historical climate qc was done and add amount of climate_qc_months
+    gdir.add_to_diagnostics('climate_qc_months', qc_months)
     if qc_months == 0:
         return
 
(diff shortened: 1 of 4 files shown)


fmaussion:
Fantastic! Thanks @lilianschuster

PR #1361, opened 2022-01-13
Dynamic spinup function

pat-schmitt:
Here is my updated version of the dynamic spinup function.
The main changes are:

- safeguards included which prevent taking too large steps between the minimisation iterations (also changed from a polyfit to a splinfit = a linear interpolation)
- errors that can occur during minimisation are handled explicitly now (e.g. glacier out of domain, or glacier ice free after spinup but still too large at the end) 
- include `ignore_error` option, if an error occurs the glacier is stored without a dynamic spinup (so they are available for the following tasks)
- included the dynamic spinup function in `run_prepro_levels`

Things to do:
- [ ] the final test of current implementation (currently in queue on cluster, but no big changes between the current version and the last successfully tested version)
- [ ] define additional glaciers for more advanced testing, of the handling of all error cases that can occur during the dynamic spinup 
- I am currently adding quite some variables to the glacier diagnostics, maybe we can discuss if they are necessary for operational runs

- [ ] Fully documented


pat-schmitt:
I have now added some glaciers to test the error handling of `run_dynamic_spinup`.
But for this, I added a prepro URL to the `secure_url_retrieve` of `conftest.py`. So as already discussed these glaciers should be combined in a smaller tar file in order to save computing time during the tests.

oggm/tests/test_utils.py
```diff
@@ -1205,6 +1210,24 @@ def test_elev_bands_run(self):
...
+                          'dynamic_spinup_run_output_11.nc')
+
+        import matplotlib.pyplot as plt
+        with xr.open_dataset(hf) as dsh:
+            dsh = dsh.load()
+        with xr.open_dataset(ef) as dse:
+            dse = dse.load()
+        with xr.open_dataset(sf) as dss:
+            dss = dss.load()
+        assert len(dss.time) > 1
```
fmaussion, line 1228: @pat-schmitt I rebased your branch and added a commit. You'll need to force pull, but I think I didn't brake anything.

Can you please have a look at this test and what I'm doing wrong? the output only has one year, and I didn't see any error message pointing me to the right direction here

pat-schmitt:
Now I have added an option `store_model_evolution` to save the whole model evolution of the spinup and adapted my tests to test if it is working (e.g. delete all files if an error occurs; the model evolution is saved for more than one year if no error occurs; ...).

fmaussion:
Thanks a lot @pat-schmitt !

Issue #1269, opened 2021-07-01
How to update the "rgi62_dem_frac.h5" file when using a non-RGI glacier inventory?


Keeptg:
Since version 1.4, OGGM automatically choosen the DEM source according `rgi62_dem_frac.h5` file. This avoid the using of bad DEM in OGGM as much as possible. 

While the default `rgi62_dem_frac.h5` file only support RGI glacier inventory. For the using of custom glacier inventory (modified using `utils.cook_rgidf()`), the `rgi62_dem_frac.h5` file might return wrong information (the mismatch between modified glacier inventory and the real RGI) or errors (if the glacier number of the custom glacier inventory are more than that in RGI for a RGI region).

For my case, I'm running the prepro_levels script (modified from [`oggm.cli.prepro_levels.py`](https://github.com/OGGM/oggm/blob/master/oggm/cli/prepro_levels.py)) based on the updated GAMDAM, and I got `KeyError` for all of the glaciers with RGI id behind `RGI60-13.54431` (the last RGI id in the real RGI60 for the region 13).

Instead of using preprocessed `rgi62_dem_frac.h5`, I'm wondering if it's possible that OGGM can do this part of job in the real time? Or how can I update `rgi62_dem_frac.h5` to make sure OGGM work well under the driven of updated GAMDAM?

fmaussion:
Thanks for asking. I think that when I implemented this I thought that custom inventories should specify their DEM sources explicitly as per the tutorial: https://oggm.org/tutorials/stable/notebooks/dem_sources.html#choose-a-dem-source-e-g-srtm

fmaussion:
Other ideas welcome!

Keeptg:
> 
> 
> Other ideas welcome!

How did you generate the `rgi62_dem_frac.h5` file? Do you think it is possible to add an argument that could make OGGM calculate the fraction and decide which dem should be used during execute `define_glacier_region()` function? It will be great if the `rgi62_dem_frac.h5` file or like this can be generated during executing prepro_levels procedure.

fmaussion:
`rgi62_dem_frac.h5` was generated based on RGI-TOPO https://rgitools.readthedocs.io/en/latest/dems.html 

You need to download and process all DEMs to generate this file, which is overkill for this purpose in my opinion. Taking your use case as an example, all glaciers in HMA fit within the NASA dataset, so that it's much faster to just set it manually!

Keeptg:
[...]

PR #1364, opened 2022-01-20
Add COPDEM30

### LINKED ISSUE #1362: COPDEM90 and COPDEM30
For RGI7 I think we really need to support COPDEM30 (I think that the idea will be to use it for all glaciers with a spatial resolution < 90 m). At the same time, we need to rename COPDEM to COPDEM90.

I just had a look today at the download instructions, etc, and needless to say it's a bit of a maze for me.

@matthiasdusch I would completely understand if you can't or don't want to do it, but would you be interested in having a look at that?

related: https://github.com/OGGM/oggm/issues/1074

matthiasdusch:
Closes  #1362 

- [x] Tests added/passed
- [x] Fully documented
-----------------------------------------

Work in progress. I hope to get a lookup shapefile from Copernicus in the next days. Otherwise this has to be made more or less manually.

Open to discussion: Copernicus offers different releases of both COP30 and 90. Some info can be found here under "Releases": https://spacedata.copernicus.eu/web/cscda/dataset-details?articleId=394198
I guess it makes sense to use the latest version, but it could cause some inconsistencies for people.

note to myself: #961 was the previous PR

### DIFF
--- a/oggm/utils/_downloads.py
+++ b/oggm/utils/_downloads.py
@@ -69,7 +69,7 @@ logger = logging.getLogger('.'.join(__name__.split('.')[:-1]))
 # The given commit will be downloaded from github and used as source for
 # all sample data
 SAMPLE_DATA_GH_REPO = 'OGGM/oggm-sample-data'
-SAMPLE_DATA_COMMIT = 'cf480d9790f3559cc52764899d7f44eef4cd418a'
+SAMPLE_DATA_COMMIT = 'df250ee69405c2108b6782781d6aa6c22aa3919d'
 
 GDIR_L1L2_URL = ('https://cluster.klima.uni-bremen.de/~oggm/gdirs/oggm_v1.4/'
                  'L1-L2_files/centerlines/')
--- a/oggm/utils/_downloads.py
+++ b/oggm/utils/_downloads.py
@@ -86,8 +86,8 @@ CHECKSUM_LIFETIME = 24 * 60 * 60
 WEB_N_PIX = 256
 WEB_EARTH_RADUIS = 6378137.
 
-DEM_SOURCES = ['GIMP', 'ARCTICDEM', 'RAMP', 'TANDEM', 'AW3D30', 'MAPZEN',
-               'DEM3', 'ASTER', 'SRTM', 'REMA', 'ALASKA', 'COPDEM', 'NASADEM']
+DEM_SOURCES = ['GIMP', 'ARCTICDEM', 'RAMP', 'TANDEM', 'AW3D30', 'MAPZEN', 'DEM3',
+               'ASTER', 'SRTM', 'REMA', 'ALASKA', 'COPDEM30', 'COPDEM90', 'NASADEM']
 DEM_SOURCES_PER_GLACIER = None
 
 _RGI_METADATA = dict()
@@ -1151,12 +1151,12 @@ def _download_topo_file_from_cluster_unlocked(fname):
     return outpath
 
 
-def _download_copdem_file(cppfile, tilename):
+def _download_copdem_file(cppfile, tilename, version):
     with get_lock():
-        return _download_copdem_file_unlocked(cppfile, tilename)
+        return _download_copdem_file_unlocked(cppfile, tilename, version)
 
 
-def _download_copdem_file_unlocked(cppfile, tilename):
+def _download_copdem_file_unlocked(cppfile, tilename, version):
     """Checks if Copernicus DEM file is in the directory, if not download it.
 
     cppfile : name of the tarfile to download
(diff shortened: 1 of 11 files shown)


fmaussion:
Thanks so much!!!

> I guess it makes sense to use the latest version, but it could cause some inconsistencies for people.

Yes, we don't really care about backwards compatibility here (for both the 90 and 30 m products). Cheers!

oggm/utils/_downloads.py
```diff
@@ -2159,7 +2159,7 @@ def _in_grid(grid_json, lon, lat):
         return True
     elif source == 'SRTM':
         return np.max(np.abs(lat_ex)) < 60
-    elif source == 'COPDEM':
+    elif (source == 'COPDEM30') or (source == 'COPDEM90'):
```
fmaussion, line 2162:
```suggestion
    elif source in ['COPDEM30', 'COPDEM90']:
```

matthiasdusch:
needs https://github.com/OGGM/oggm-sample-data/pull/12 first and than the hash updated

Seems to pull the correct files. No differences on larger glaciers. Visible differences on smaller glaciers.

HEF:
RGI60-11 00897_dem_topo_color

RGI60-16.02207 (Peru):
RGI60-16 02207_dem_topo_color

RGI60-16.01640 (Mt Kenya):
RGI60-16 01640_dem_topo_color

RGI60-15.00060 (South Asia East):
RGI60-15 00060_dem_topo_color

RGI60-01.08063 (Alaska):
RGI60-09 00552_dem_slope

fmaussion:
Looking good, thanks!!!!!

You should be able to change the commit hash now and there seems to be a few pep8 to fix

oggm/utils/_downloads.py
```diff
@@ -2190,6 +2202,9 @@ def default_dem_source(rgi_id):
     sel = cfg.DEM_SOURCE_TABLE[rgi_reg].loc[rgi_id]
     for s in ['NASADEM', 'COPDEM', 'GIMP', 'REMA', 'TANDEM', 'MAPZEN']:
         if sel.loc[s] > 0.75:
+            # this can go as soon as 'rgi62_dem_frac.h5' is updated:
```
matthiasdusch, line 2205: @fmaussion what do you think about this? Not sure when/how/where this 'rgi62_dem_frac.h5' gets updated. Or if this might break somewhere else.

fmaussion: Yeah I think this is irrelevant for this PR - I really need to rethink this logic (related: https://github.com/OGGM/oggm/issues/1269)

oggm/utils/_downloads.py
```diff
@@ -1528,29 +1528,41 @@ def alaska_dem_zone(lon_ex, lat_ex):
...
+    if version == '90':
+        asec = '30'
+    elif version == '30':
+        asec = '10'
+    else:
+        raise InvalidDEMError('COPDEM Version not valid.')
+
     # path to the lookup shapefiles
-    gdf = gpd.read_file(get_demo_file('RGI60_COPDEM_lookup.shp'))
+    df = pd.read_csv(get_demo_file('copdem{}_2021_1.csv'.format(version)))
```
fmaussion, line 1544: Note for a later PR: this may need to be read into a permanent container to not have to read the same file over again each time a glacier wants to know the tiles

oggm/utils/_downloads.py
```diff
@@ -1528,29 +1528,41 @@ def alaska_dem_zone(lon_ex, lat_ex):
...
-        flist.append((cpp, eop))
-
+    for lat in lats:
+        # north or south?
+        ns = 'S' if lat < 0 else 'N'
+        for lon in lons:
+            # east or west?
+            ew = 'W' if lon < 0 else 'E'
+            Lat = '{}{:02.0f}'.format(ns, abs(lat))
+            Lon = '{}{:03.0f}'.format(ew, abs(lon))
```
fmaussion, line 1558: `lon_str`, `lat_str` would be a better name here (sorry)

matthiasdusch:
sure, will do. I am just wondering at the moment why these tests take so long or probably timeout? I can't really see at which point they are stuck... can you?

oggm/utils/_downloads.py
matthiasdusch: fair point. if you have any similar case which is working, I could do it

fmaussion: `cfg.DATA` is a multiprocessing aware dictionary used for example here:

https://github.com/OGGM/oggm/blob/919d9044fea7c54320d533e67607b2f5d1b7c51a/oggm/utils/_downloads.py#L1316-L1358

fmaussion:
Thanks a lot! This is awesome. I have hired a student to work on applying this for RGI7, work will start end of February we'll keep you in touch.

PR #1366, opened 2022-01-26
get ref mb profile with constant step size

lilianschuster:
- related to: https://github.com/OGGM/oggm-shop-notebooks/pull/5 and to https://github.com/OGGM/oggm-sample-data/pull/11
- allows to get the interpolated constant step size MB profiles (i.e. altitude difference for all profile levels is dh=50m) via `gdir.get_ref_mb_profile(constant_dh=True)`
- simple test added, but this will only work with the next pull request (with the updated oggm-sample-data download commit number)

- [x] Tests added/passed

### DIFF
--- a/oggm/utils/_workflow.py
+++ b/oggm/utils/_workflow.py
@@ -3252,7 +3252,7 @@ class GlacierDirectory(object):
             # list of years
             self._mbprofdf = pd.read_csv(reff, index_col=0)
 
-        elif self._mbprofdf_cte_dh is None and constant_dh is True:
+        if self._mbprofdf_cte_dh is None and constant_dh:
             flink, mbdatadir = get_wgms_files()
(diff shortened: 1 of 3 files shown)


fmaussion:
Thanks! Yes open the PR with the commi hash change first, then we ce re-run the tests here

oggm/utils/_workflow.py
```diff
@@ -3247,16 +3252,40 @@ def get_ref_mb_profile(self, input_filesuffix=''):
             # list of years
             self._mbprofdf = pd.read_csv(reff, index_col=0)
 
+        elif self._mbprofdf_cte_dh is None and constant_dh is True:
```
fmaussion, line 3255:
```suggestion
        if self._mbprofdf_cte_dh is None and constant_dh:
```

fmaussion:
Thanks!

PR #1369, opened 2022-02-07
align the keywords being used for compile_ela and compile_fixed_geome…

anoukvlug:
This PR is supposed to align the `compile_ela` and `compile_fixed_geometry_mass_balance` functions. I thought it might be nice if they look more alike and the same keywords as being optional in `compile_ela` can also be used by  the `compile_fixed_geometry_mass_balance`.

- [ ] Tests added/passed


fmaussion:
Thanks!

PR #1376, opened 2022-02-17
Add hydro test to spinup

fmaussion:
cc @pat-schmitt FYI

### DIFF
--- a/oggm/tests/test_models.py
+++ b/oggm/tests/test_models.py
@@ -4044,6 +4044,86 @@ class TestHydro:
         assert_allclose(odf['dom_area'], odf['dom_area'].iloc[0], rtol=1e-4)
 
     @pytest.mark.slow
+    def test_hydro_dynamical_spinup(self, hef_gdir, inversion_params):
+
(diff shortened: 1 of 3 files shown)


oggm/tests/test_models.py
```diff
@@ -4043,6 +4043,86 @@ def test_hydro_ref_area(self, hef_gdir, inversion_params):
...
+
+        assert odf_hist.index[0] == 1971
+        assert odf_hist.index[-1] == 1990
+        assert odf_run.index[0] == 1991
+
+        odf = pd.concat([odf_hist, odf_run])
+
+        # Domain area is constant and equal to the first year
+        odf['dom_area'] = odf['on_area'] + odf['off_area']
+        # assert_allclose(odf['dom_area'], odf['dom_area'].iloc[0], rtol=1e-4)
```
pat-schmitt, line 4090: @fmaussion I think this line should be uncommented

fmaussion: No it should be deleted - the glacier grows a bit during the simulation so this does not work

PR #1377, opened 2022-02-17
added ys as a option to dynamic spinup

pat-schmitt:
Adding the option for setting the spinup period length OR the start year of the dynamic spinup

- [x] Tests added/passed
- [ ] Fully documented


fmaussion:
Can you also consider changing the default tolerance to 1%? I realized 10% is probably too big.

fmaussion:
Thanks @pat-schmitt !

PR #1383, opened 2022-02-24
Add option to plot velocity in graphics

minallah:
Plot velocity after inversion (u_surface, u_integrated) - similar to plot_inversion for thickness.


fmaussion:
Thanks for the contribution! I haven't looked into the details, but would it also be possible to add a keyword argument to `plot_inversion` as to which variable to plot (thickness or velocity, thickness being the default) and which cmap to use? This would avoid code duplication if at all possible.

minallah:
I have updated the plot_inversion to intake velocity and change the color bar. Tested and it works fine.

fmaussion:
Thanks @minallah ! Great first contribution :tada:

PR #1384, opened 2022-02-24
added level 4 as potential start_level in run_prepro_levels

pat-schmitt:
Added level 4 as potential start_level in run_prepro_levels

- [ ] Tests added/passed

### DIFF
--- a/oggm/cli/prepro_levels.py
+++ b/oggm/cli/prepro_levels.py
@@ -22,6 +22,8 @@ from oggm.core import gis
 from oggm.exceptions import InvalidParamsError, InvalidDEMError
 
 # Module logger
+from oggm.utils import get_prepro_base_url, file_downloader
+
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


fmaussion:
is this ready to merge? I cant check in detail but I trust this works - you can see by yourself if start from 4 actually works

pat-schmitt:
Yes, if the last (currently running) test passes it could be merged. But you can also wait until the cluster runs finished (I am installing oggm from my fork), so I will see if it actually works.

pat-schmitt:
Ok, I have overlooked that also for L5 some files are copied from L3. But the same files were already copied to L4, so a workaround to be able to start from L4 would be to get excess to the `L4/summary/` folder from the prepro URL and then copy the needed files from their into `L5/summary/`. (The same could also be used when we want to integrate the option to start from L3).

I am not sure what is the best way to get excess to `L4/summary/`, but there are already some downloaders available. I thought maybe I can use `file_downloader` (together with `get_prepro_base_url`) to download the `/summary/` folder  to a local path and use this path directly as `sum_dir_L4` (which replace than `sum_dir_L3` at https://github.com/OGGM/oggm/blob/master/oggm/cli/prepro_levels.py#L646 if `start_level=4`). @fmaussion what do you think about this?

pat-schmitt:
ok I played around a little bit, and it seems to work, but instead of downloading the whole directory I will download only the files of the current rgi-region with the `file_downloader`. So I will implement this. And I plan to also implement the possibility to start from L3 with the same logic, while I am at it

pat-schmitt:
I think this is ready to merge.

PR #1402, opened 2022-03-23
Update of GitHub security

nchampollion:
Maybe this has been already discussed and changed before

- [ ] Tests added/passed
- [ ] Fully documented


fmaussion:
Thanks @nchampollion ! If that's ok with you I'd prefer not to bloat the already very long oggm docs with stuff related to github only and that people might be able to google themselves. Would you agree?

nchampollion:
Yeah, for me that's not a problem. I wonder if that's useful for students or if googleling themselves is enough. Up to you :-)

PR #1403, opened 2022-03-27
Water terminating

jmalles:
My code for inverting and running water-/marine-terminating glaciers.

### DIFF
--- a/oggm/core/climate.py
+++ b/oggm/core/climate.py
@@ -1306,24 +1306,28 @@ def mu_star_calibration(gdir, min_mu_star=None, max_mu_star=None):
 @entity_task(log, writes=['inversion_flowlines'],
              fallback=_fallback_mu_star_calibration)
 def mu_star_calibration_from_geodetic_mb(gdir,
+                                         fa_data_path=None,
                                          ref_mb=None,
                                          ref_period='',
                                          step_height_for_corr=25,
                                          max_height_change_for_corr=3000,
                                          ignore_hydro_months=False,
                                          min_mu_star=None,
-                                         max_mu_star=None):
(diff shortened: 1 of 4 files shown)


fmaussion:
OK so I had a quick look at all the changes - in my view this looks great, but of course there is still some work before merge ;-) 

I think that something that would be very useful as a first step is to make sure that all the tests unrelated to calving pass. The calving tests will need to be adapted for sure, also including inversion. 

Do you think you can manage to do that @jmalles ? If you need help, it's worth discussing if this could become a team effort with @anoukvlug as this is quite an important improvement to OGGM.

fmaussion:
@jmalles is this the version of the code which was used for the paper? Is it worth that I and @pat-schmitt  have a look at it over the christmas break or did you want to change things?

jmalles:
Yes, I think this should be the version used for the paper and I was not planning to change things at the moment. Though I think the geodetic calibration function with using Will's data needs some more work, as it is kind of hardcoded right now... Other than that, I hope it is in a shape that is worth having a look at.

fmaussion:
@jmalles I changed my mind with respect to what I said on Slack: please keep this PR unchanged and open two new, fresh PRs. Lets start with one only adding the parameterisation (not the calibration). It should be possible to add the parameterization and test it without the rest. Please remove bits of code which have become obsolete (e.g. the diagnostic variables you did not use) because we can retriev them later from this PR. Please add some inline comments where it might be necessary. 

If we work together on smaller bits we can get there!

fmaussion:
Leaving this open for documentation for a while - but fixing this will require a full set of new code to deal with the specifics of v1.6

PR #1408, opened 2022-04-04
Numeric instability filter

pat-schmitt:
This is the first attempt at implementing a filter that checks for numeric instabilities and filters them out. This is still work in progress, but for discussing the design and implementation I open it in its really early phase.

The principle Idea: An instability is defined as a zigzag pattern in the ice velocity. If this zigzag pattern is visible over a minimum number of grid points it is classified as an instability. To filter this out the surface slope is smoothed in two steps. First a running mean is used which considers some grid points before and after the instability (for a smooth transition). Afterwards, these smoothed slopes are used to recalculate the surface heights and a mass conserving step is conducted. This step compares the mass before and after the smoothing and redistributes the difference along all instability grid points (using a gaussian shape to ensure a smooth transition between good grid points and the instability grid points).

With the current implementation, you can choose to:
- do no smoothing `False` (as it was until now)
- smooth `'yearly'`
- smooth `'each_timestep'`

Possible other options could become in the future (maybe useful for global runs):
- smooth all instabilities if `max(u)` is contained in instability
- smooth only instability which contains `max(u)`

The changes in the source code (which can be discussed):
- small changes to utils `signchange` function to be able to use it not only with time series but also with numpy arrays, and also included the treatment (and added a test) if the given array contains only zeros
- included a function in utils that find zigzag patterns in a given array
- included some parameters in params.cfg for instability smoothing options, some of them can maybe be deleted after testing, but for testing, I find it easier to have easy excess to them
- added a dummy method `smooth_instabilities_in_flowlines()` to the `FlowlineModel` class, which should be implemented in the child classes
- in the `FlowlineModel` method `run_until` I added the options to smooth instabilities yearly or at each time step (is calling `smooth_instabilities_in_flowlines()` method then)
- in the `FluxBasedModel` I implemented the `smooth_instabilities_in_flowlines()` method
- For testing purposes, I added in `flowline_model_run` and `dynamic spinup` a diagnostic variable that indicates if smoothing was used or not

What to discuss further (from my point of view):
- How should the smoothing be tested on runs? Currently only tested on one glacier, so I think there will be other issues popping up when testing it on a larger scale.
- How should the smoothing function should be tested with pytest?
- Is this methodologically acceptable?

Must Dos:
- [x] Tests added for `detect_instabilities`
- [ ] Tests added for smoothing function
- [ ] Fully documented

### DIFF
--- a/oggm/cfg.py
+++ b/oggm/cfg.py
@@ -569,6 +569,19 @@ def initialize_minimal(file=None, logging_level='INFO', params=None,
     PARAMS[k] = cp[k]
     k = 'glacier_length_method'
     PARAMS[k] = cp[k]
+    k = 'use_instability_smoothing'
+    if cp[k] == 'False':
+        PARAMS[k] = cp.as_bool[k]
+    else:
+        PARAMS[k] = cp[k]
+    k = 'instability_min_length'
+    PARAMS[k] = cp.as_int[k]
+    k = 'instability_smoothing_window'
+    PARAMS[k] = cp.as_int[k]
+    k = 'instability_smoothing_steps'
+    PARAMS[k] = cp.as_int[k]
+    k = 'instability_smoothing_mass_redis_fct'
+    PARAMS[k] = cp[k]
 
     # Others
     PARAMS['tidewater_type'] = cp.as_int('tidewater_type')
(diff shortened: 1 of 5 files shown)


pat-schmitt:
I am closing this PR here, as after some discussions it looks like this will only be relevant for COMBINE directly and not for OGGM itself.

fmaussion:
Please open an issue briefly writing down some of the ideas developed here for future reference. (just a couple sentences)

PR #1419, opened 2022-05-17
allow to use gswp3_w5e5 in OGGM, use glacier-individual prcp.fac

lilianschuster:
This is a work-in-process pull-request!
It is a first try:
- to incorporate gswp3_w5e5 (ISIMIP3a) data into OGGM
- to use glacier-individual prcp. fac that decrease with larger winter precipitation, the actual fit comes from a calibration of reference glaciers where mustar and prcp. fac were calibrated to match both, the geodetic bias and the winter MB (this was done with the MBsandbox
- you can find a first workflow with open questions here: [https://github.com/OGGM/massbalance-sandbox/blob/master/docs/geodetic_mb_calibration_to_volume_changes.ipynb](https://github.com/OGGM/massbalance-sandbox/blob/master/docs/geodetic_mb_calibration_to_volume_changes.ipynb) 
- the main question is: how do we deal with the glacier-individually changing prcp. fac? I have proposed one option, but maybe it should be done differently or always be checked directly inside of `PastMassBalance` ? 

- [x] Tests added/passed

### DIFF
--- a/oggm/core/climate.py
+++ b/oggm/core/climate.py
@@ -538,7 +538,7 @@ def mb_climate_on_height(gdir, heights, *, time_range=None, year_range=None):
         # Have we decided on a factor yet?
         try:
             df = gdir.read_json('local_mustar')
-        except:
+        except FileNotFoundError:
             df = {}
         prcp_fac = df.get('prcp_fac_from_winter_prcp')
         if prcp_fac is None:
@@ -1330,6 +1330,7 @@ def mu_star_calibration(gdir, min_mu_star=None, max_mu_star=None):
 
 
 def decide_winter_precip_factor(gdir):
+    """Utility function to decide on a precip factor based on winter precip."""
 
     # We have to decide on a precip factor
     if 'W5E5' not in cfg.PARAMS['baseline_climate']:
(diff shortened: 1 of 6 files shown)


lilianschuster:
I added `process_isimip_data` and tests. 
- For that, there are some test isimip gcm files inside of https://cluster.klima.uni-bremen.de/~oggm/test_climate/cmip6/isimip3b
- the actual (only monthly) ISIMIP3b flattened files are now in: https://cluster.klima.uni-bremen.de/~oggm/cmip6/isimip3b/

The workflow to do projections with the data is in the updated notebook:
-> here: https://github.com/OGGM/massbalance-sandbox/blob/master/docs/geodetic_mb_calibration_to_volume_changes.ipynb 

What is still unclear:
- how do we deal with the glacier-individually changing prcp. fac? I have proposed one option, but maybe it should be done differently or always be checked directly inside of PastMassBalance ?

lilianschuster:
(I don't know from where the error in the one test comes from: when I did pytest on my computer before uploading to github, all tests passed)

fmaussion:
@lilianschuster can you please review? I made some substantial changes and I may have overseen something.

In particular in `def climate_statistics` I seem to have removed code: please add the computation of winter prcp allways (not on keyword) similar to the other variables (e.g. also for the same period).

Thanks!

Issue #1422, opened 2022-05-26
the function "mu_star_calibration_from_geodetic_mb" write to json error


aierfulz:
Dear OGGM development team:
       The line 1495 of climate.py ` gdir.add_to_diagnostics('ref_hgt_calib_diff', sel_steps[0] - start)` need to modify as ` gdir.add_to_diagnostics('ref_hgt_calib_diff', int(sel_steps[0] - start))
` to avoid write to json error.

fmaussion:
Thanks for the report! you can try to make the change yourself if you want! 

Edit the file online at: https://github.com/OGGM/oggm/blob/master/oggm/core/climate.py

fmaussion:
I see no open pull-request? You would need to send us a request change.

aierfulz:
Dear Maussion: I had revised it again. this time may be right.

PR #1423, opened 2022-05-30
Prevent write to json error on some platforms

aierfulz:
Closes https://github.com/OGGM/oggm/issues/1422


fmaussion:
Thanks a lot!

aierfulz:
> 
That's a very very small contribution. OGGM is a sate-of-art glacier model at present.    I hope it will be better in the future with the efforts of many excellent scholars.

PR #1385, opened 2022-03-02
Add option to ingest user-provided mass balance in run* tasks

minallah:
Made minor changes to following tasks:
run_dynamic_spinup
run_from_climate_data
run_constant_climate
run_random_climate

PR #1425, opened 2022-06-07
Adding dynamic mu star calibration

pat-schmitt:
With this PR I add a new function to dynamically calibrate mu star to a geodetic mass-balance. Moreover, all dynamic calibration functions (`run_dynamic_spinup` to match area or volume at the RGI date, `run_dynamic_mu_star_calibration` with different options on how to do this) are now combined in the new module `oggm.core.dynamic_spinup`.

The different options for the dynamic_mu_star_calibration at the moment are:
- `area/dmdtda/atonce/inversion` (first analysis starting from gdirs v1.6/ERA5/qc0: full success 80% and improvements 8% of the global RGI area): At each iteration, the whole inversion chain is conducted followed by one dynamical run from 1980 to 2020. This option creates a consistent past model run with:
    - A-parameter is calibrated glacier individually to match the 'calibrated volume' ('calibrated volume' is the volume of the glacier after calibrating a whole region to the consensus volume (with a regional A-parameter), this calibration must be done before starting with the dynamic mu star calibration)
    - The area at the RGI date of the model matches the observation (inside a user provided precision)
    - mu star is calibrated that the modelled geodetic mass-balance matches the observation (inside the provided observation error)
- `area/before/dmdtda` (first analysis starting from gdirs v1.6/ERA5/qc0: full success 87% and improvements 12% of the global RGI area): The second option is to set the RGI date for all glaciers to 2000 and conduct a dynamic spinup from 1980 to 2000 to match the RGI area (inside a user provided precision). After this conduct a dynamic mu star calibration from 2000 to 2020 to match the geodetic mass-balance. This option creates an inconsistency during the past model run, as mu_star changes after the year 2000 and forces the RGI date to 2000.
- `area/dmdtda/atonce` (first analysis starting from gdirs v1.6/ERA5/qc0: full success 79% and improvements 9% of the global RGI area): A potential third option (now implemented) could be to try to make an iteration without an inversion but still with one consistent past model run. It is the same as `area/dmdtda/atonce/inversion` without doing an inversion at each iteration.

Main changes in the code:
- included the first two above options into `run_prepro_levels`
- adapted `run_dynamic_spinup` to also be able to run until a user provided year (which is larger than the RGI year). Therefore I introduced two new variables in `run_until_and_store`, namely `volume_m3_min_h` and `area_m2_min_h`, and I also included `dynamic_spinup_min_ice_thick` in `params.cfg`. All this is needed to use a minimum ice thickness when comparing the modelled values to the observations, and as I want to conduct one model run (for option 1 above) I need to save these values during the run as there is currently no option to stop a model run in between to read out some data and continue afterwards. (with the minimum ice thickness very small changes due to climate variability are excluded when comparing the model to the observation)
- added an option in `calibrate_inversion_from_consensus` to provide a user-specific volume to calibrate for (needed for the first option); I also added a test for this in `test_calibrate_inversion_from_consensus`
- including the `run_dynamic_mu_star_calibration` function together with 4 functions which define how one iteration should look like. (4 functions because for each function there is a fallback function defined which tells the algorithm what to do if the main iteration function fails and `ignore_errors=True`; this could be useful to define a plan b to use as many glaciers as possible for regional/global studies)
- included tests for the new `run_dynamic_spinup` properties (like running until a user provided `ye`)
- included tests for `run_dynamic_mu_star_calibration`:
    - for this to work I had to include a new 'secure_url'
    - some basic tests and hydro tests for both options mentioned above (just copied from hydro tests for `run_dynamic_spinup`)
    - I also 'included' a test in `test_elev_bands_and_spinup_run`, but for this to work the example climate data must be extended (currently only goes until 2005, but here we need it up to 2020), therefore I included some lines of code as comments, which can be uncommented once the example data goes up to 2020

- [x] Tests added/passed locally (beside one graphics test `test_coxe()`; and `test_prepro.py::TestClimate::test_fake_ref_mb_glacier` is failing if running all tests using `pytest --pyargs oggm --run-slow --mpl`, but is running successful if only running `TestClimate` in pycharm)
- [ ] Fully documented
- [ ] ~~in-depth analysis of prepro gdirs (three different options `area/dmdtda/atonce/inversion`, `area/dmdtda/atonce` or `area/before/dmdtda`; and the same options for `volume` instead of `area`)~~
- [ ] Introduce dynamic mu star calibration in 'Dynamic spinup for past simulations' tutorial
- [x] add changes of other PRs to `run_dynamic_spinup` before merging (e.g. https://github.com/OGGM/oggm/pull/1385)


pat-schmitt:
The remaining pep8 issue can not be solved as the function name is too long. Maybe I will change this long name (which includes all details) to an abbreviation.

pat-schmitt:
I changed now the tasks of the perpro levels 4 and 5 (up to level 3 everything is unchanged).
In level 4 now the dynamic calibration runs and the historical runs (old default, maybe they can also be deleted) are added. This is done because the dynamic calibration runs also conduct inversions, and therefore need all files. Further, we can start from an already existing level 3 prepro level and add the dynamic calibration runs by doing so.
Level 5 now just creates the minigdirs -> no inversion is possible at this level, but gdir size is considerably reduced (same as old level 4)

pat-schmitt:
I am currently running a last test on the cluster. Once this has finished (I will add a comment) I think the PR is ready for review/merge!

fmaussion:
The remaining failures are unrelated - merge?

pat-schmitt:
No, I am currently adding an additional strategy for the dynamic mu calibration if the rgi_year is smaller than the provided climate data (e.g. dynamic mu calibration completely fails for Region 2 for ERA5).

pat-schmitt:
From my side, it is ready to merge now. I added the changes in whats-new and will open an issue for the documentation and the tutorials.

fmaussion:
:partying_face:  awesome work!

Issue #1424, opened 2022-05-31
Should we change a line in climate.py?


bowenbelongstonature:
I think this [line](https://github.com/OGGM/oggm/blob/master/oggm/core/climate.py#L1613) just send the start year and the end year to `mb_years` for subsequent calculations.

fmaussion:
@lilianschuster can you have a look?

lilianschuster:
Thanks Bowen for looking into the code in these details!

In OGGM/massbalance-sandbox, I always predefine mb_years (as an array) when using `apparent_mb_from_any_mb` (see: 
https://github.com/OGGM/massbalance-sandbox/blob/b2208c1028fb0d47b520f47cfd8f3f98a0ed8e13/MBsandbox/help_func.py#L2241 ). 

So, I am also confused from where the [y0,y1-1] comes from. I did some "research": The mentioned line was included in this pull request (https://github.com/OGGM/oggm/pull/1286), also see https://github.com/OGGM/oggm/commit/9e433f4c3044e27c09b18949bd8c7b44f67c12f7 line 1567  . According to the documentation, `mb_years` should be an array, and not just two years. **So after debugging and looking into my own code, I totally agree with @bowenbelongstonature , that  [line](https://github.com/OGGM/oggm/blob/master/oggm/core/climate.py#L1613) should be changed to  `np.arange(y0, y1,1)`** 

In the tests, also just two values are applied on this line https://github.com/OGGM/oggm/blob/1fca0cbd3564c33ff82ab3c4c41307055e7c7b85/oggm/tests/test_models.py#L4359 . 
Either I don't understand the concept here, or it should also there be an array? 

In any case, if we change this, it has consequences for some preprocessed gdirs! I don't know, however if they are important. 

What do you think, @fmaussion ?

fmaussion:
Yes this is a bug, thanks for catching this. I think @drounce might be affected in his own code, depending how he calls `apparent_mb_from_any_mb`.

I think we should:
- change the line as suggested
- raise an error when the number of years given to `apparent_mb_from_any_mb` is smaller than 10.

Can someone take care of this?

bowenbelongstonature:
I've made a PR.

Will some preprocessed  data in Shop be affected by this? @fmaussion

bowenbelongstonature:
[...]

PR #1426, opened 2022-06-07
Change default `mb_years` in #L1613 of climate.py

bowenbelongstonature:
Make sure `mb_years` get an array and raise an error when the count/length of years given to `apparent_mb_from_any_mb` is smaller than 10.

- [ ] Tests added/passed
- [ ] Fully documented

### DIFF
--- a/oggm/tests/test_prepro.py
+++ b/oggm/tests/test_prepro.py
@@ -1999,7 +1999,7 @@ class TestClimate(unittest.TestCase):
         np.testing.assert_allclose(mb_new.bias, 0)
 
         # Check that inversion works
-        climate.apparent_mb_from_any_mb(gdir, mb_years=np.arange(1953, 2003, 1))
+        climate.apparent_mb_from_any_mb(gdir, mb_years=[1953, 2003])
 
         # Artificially make some arms even lower to have multiple branches
         fls = gdir.read_pickle('inversion_flowlines')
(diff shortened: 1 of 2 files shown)


bowenbelongstonature:
#1424

bowenbelongstonature:
This [line](https://github.com/OGGM/oggm/blob/master/oggm/tests/test_models.py#L4359) in test_models.py is also change accordingly.

bowenbelongstonature:
3 checks failed since the changed code get `mb_years` shorter than 10. Solving the issue beyonds my authority, I think.

fmaussion:
> 3 checks failed since the changed code get mb_years shorter than 10. Solving the issue beyonds my authority, I think.

I think it is ;-) - If you don't mind, we would welcome a fix for these 3 lines to call `apparent_mb_from_any_mb` as intended (with a list of years)

bowenbelongstonature:
Of course not.

oggm/tests/test_prepro.py
```diff
@@ -1950,7 +1950,7 @@ def test_geodetic_mb_calibration_single_fl(self):
         assert 5 < mb_new.mu_star < 500
 
         # Check that inversion works
-        climate.apparent_mb_from_any_mb(gdir, mb_years=np.(1953, 2003, 1))
+        climate.apparent_mb_from_any_mb(gdir, mb_years=np(1953, 2003, 1))
```
fmaussion, line 1953: you want this to be `np.arange`

bowenbelongstonature: > you want this to be `np.arange`

I've changed them.

But 2 `mb_years` in `test_find_calving_any_mb` get just 1 year (which is 2000), I have no idea about this.
Maybe it's related to this line `mb = massbalance.LinearMassBalance(ela_h=2000)`? What should we do for this?

fmaussion:
Thanks @bowenbelongstonature !

PR #1435, opened 2022-07-07
allow to use gswp3_w5e5 in OGGM, use glacier-individual prcp.fac

lilianschuster:
Sorry for opening a new pull-request ...

As suggested by @fmaussion, the `..._uncorrected_winter_daily_mean_prcp` is always computed in climate_diagnostics. Once for the time period 1979-2019 (if possible) + for the time period that is chosen. Apart from that, I checked the code and it should be working. 

--- 

I just copy the old pull-request text here : 

This is a work-in-process pull-request!
It is a first try:
-  to incorporate gswp3_w5e5 (ISIMIP3a) data into OGGM
- to use glacier-individual prcp. fac that decrease with larger winter precipitation, the actual fit comes from a calibration of reference glaciers where mustar and prcp. fac were calibrated to match both, the geodetic bias and the winter MB (this was done with the MBsandbox
- you can find a first workflow with open questions here: https://github.com/OGGM/massbalance-sandbox/blob/master/docs/geodetic_mb_calibration_to_volume_changes.ipynb
- the main question is: how do we deal with the glacier-individually changing prcp. fac? I have proposed one option, but maybe it should be done differently or always be checked directly inside of PastMassBalance ?

- [x] Tests added/passed


fmaussion:
Let's get this in! Thanks @lilianschuster!

This will probably need a tutorial soon

Issue #1401, opened 2022-03-21
Millan et al. 2022 in OGGM-shop


bearecinos:
cc. @nchampollion @JordiBolibar @fmaussion 

Hello ... 

So I started playing with this data but like I explained before I did not manage to do the automatic download of the data and the paring of each glacier to a tif from Millan et. al 2022.

How to process the data in the same style as the itslive data can be found here (probably needs improvement):
https://gist.github.com/bearecinos/26a828f8b0e1d0b4bff19bd5192deb54

The relevant function is in here:
https://github.com/bearecinos/k_calibration_greenland_jog/blob/k_new_oggm/k_tools/utils_thick.py#L215-L282

next week I am on leave .. but I will try to come up with a table like @fmaussion suggested. 
I personally think having this in the oggm shop will be very useful! Happy to help with coding this too.

JordiBolibar:
Hi @bearecinos and @fmaussion! What is the current situation of this? With ODINN we've hit a point where having this dataset available is pretty much necessary. 

I've seen it is included for the next release. Is there an approximate timeline for this? I'm not sure if I could be of any help with my limited knowledge of OGGM, but I'd be willing to help/test things if you need me to. 

Thanks again in advance, really looking forward to this! 😄

fmaussion:
There is actually very little that is OGGM specific in this. 95% of the code is already there in itslive (https://github.com/OGGM/oggm/blob/master/oggm/shop/its_live.py) and can be reused, the only thing left to do is to have a lookup table between the several Millan files and the glaciers ID (i.e. we need to know which file to fetch for which glacier).

I can take care of the rest but I would be glad if someone could work on this lookup file (either between RGI IDs or RGI regions and the respective file to fetch)

nchampollion:
I made a lookup table for RGI regions and Millan's velocity and thickness data ([https://github.com/nchampollion/Millan_data](https://github.com/nchampollion/Millan_data)).

I have few questions:
	- which format do you prefer, csv ?
	- there is usally few geotiff of Millan's data for one region, is that an issue ?
	- for regions 13, 14 and 15, geotiff Millan's data are combined, is that an issue ?
	
Finally, where do you want I upload the data ?

JordiBolibar:
Wow, that was fast! Thanks so much for working on this @nchampollion. I can volunteer to test the code with the look-up table and see if everything works smoothly, and report on this. I can also help with documentation if that can be helpful.

fmaussion:
[...]

PR #1443, opened 2022-07-19
Add Millan thickness and velocity to the shop

fmaussion:
Closes https://github.com/OGGM/oggm/issues/1401

- [x] Tests added/passed
- [ ] Fully documented


bearecinos:
Thanks @fmaussion this is awesome

fmaussion:
Yes in the end it was quite some work, partly because of some issues in the orginal dataset but also because this needs to be done with care... The kind of tasks I'd happily hand over to the community one day :sweat_smile:

JordiBolibar:
Thanks so much @fmaussion!

bearecinos:
BTW ... to use this I just want to point out that you need a new version of shapely > 1.7.1! update your environment might be best.

PR #1446, opened 2022-07-20
small bug fix to use glacier prcp_scaling_factor in dynamic mu star calibration

pat-schmitt:
Just found a small bug when using the glacier individual prcp_scaling_factor together with the dynamic mu star calibration. I will test the creation of the new prepro gdirs from this PR and will add a comment ones they passed with no errors.


fmaussion:
Thanks!

PR #1450, opened 2022-08-23
add water_level to FileModel

### LINKED ISSUE #1449: do_calving=True for non-tidewater glaciers
When reading the modeloutput of an non-tidewater glacier, the attribute `do_calving` becomes `True`. 

grafik

This seems to happen here: https://github.com/OGGM/oggm/blob/5aa684bcdcf2c8c74ad3943369d677bbb3d44999/oggm/core/flowline.py#L2126, as the dataset contains` calving_m3.values`, even if they are zero. 

This should be changed, as it causes an error in e.g. graphics.plot_modeloutput_section(model=fmod):
grafik

juliaeis:
This commit add the water_level attribute to the FileModel. This is a first step to close #1449, but did not solve the problem directly. It only solves the error from `graphics.plot_modeloutput_section`. 

Closes https://github.com/OGGM/oggm/issues/1449

### DIFF
--- a/oggm/core/flowline.py
+++ b/oggm/core/flowline.py
@@ -2123,6 +2123,7 @@ class FileModel(object):
         try:
             with xr.open_dataset(path) as ds:
                 self._calving_m3_since_y0 = ds.calving_m3.values
+                self.water_level=ds.water_level
                 self.do_calving = True
         except AttributeError:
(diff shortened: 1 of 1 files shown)


fmaussion:
Thanks!

PR #1454, opened 2022-08-30
Fix various doc and source comment typos

luzpaz:
Found via `codespell -q 3 -L alos,asign,nd,ot,ro`


anoukvlug:
Thank you @luzpaz for correcting the typos.

PR #1462, opened 2022-09-06
Try to fix fake mb test

fmaussion:


fmaussion:
@TimoRoth this fixes the test - still wonder if there isn't a deeper bug with preprocessing and shared objects accross processes somewhere but right now I'm tempted to ignore

PR #1463, opened 2022-09-06
Better isolate tests from one another

TimoRoth:
Resets multiprocessing before every test and ensures per-class and per-test cfg changes don't leak into other tests.


fmaussion (approved):
Thanks! You can merge at wish

Issue #1474, opened 2022-09-12
Change ice density in dynamic mu star calibration


pat-schmitt:
It just came to my mind that I used the oggm ice density (900 kg m-3) when I calculate the modelled geodetic mass balance in the dynamic mu star calibration, but actually, I should use the same ice density as was used when the geodetic mass balance was created (e.g. 850 kg m-3). With this in the end we match actually the volume change instead of the mass change.

My plan is to include a new kwarg `geodetic_mb_ice_density` into the dynamic mu star calibration, so if someone is using a geodetic mass balance which was derived using a different ice density assumption this could be changed.

fmaussion:
This has been a decision made at a meeting involving @drounce and others
two years ago. To discuss at the workshop

pat-schmitt:
Yes, let's discuss this at the workshop!

pat-schmitt:
We should stay with the default OGGM density of 900 kg m-3 for the conversion of volume change to mass change because OGGM only simulates ice and no snow/firn.

PR #1475, opened 2022-09-12
adds kwarg for ice density when calculating modelled geodetic mass balance

pat-schmitt:
Here I add a new kwarg to the dynamic mu star calibration to provide `geodetic_mb_ice_density`. Previously it used an ice density of 900 kg m-3 when calculating the modelled geodetic mass balance out of the modelled volume change. But to actually match the volume change instead of the mass change we should use the same ice density conversion as was used when deriving the observed geodetic mass balance out of the observed volume change (e.g. 850 kg m-3 for Hugonnet).

Closes https://github.com/OGGM/oggm/issues/1474

- [ ] Tests added/passed
- [ ] Fully documented


pat-schmitt:
I close this PR because after some discussions I understand now that we should use the OGGM default density because OGGM only simulates ice and no snow.

Issue #1448, opened 2022-08-15
Re-distribute modeled (historic) flowlines


phiscu:
Dear OGGM-Team,
I'm trying to reconstruct the state of glaciation in a catchment with (today) 38 glaciers in 1980 as starting point for a hydrological model run. Therefore, I need the distribution of ice mass in elevation bands.
Using your great tutorials I managed to derive time series of volume and area change easily and I see that NCDF containing the changing glacier geometry are generated. But it seems `distribute_thickness_per_altitude` only works on `gdirs` not on oggm-model output. Is there a way to apply the same "redistribution trick" on the modeled historic state, e.g. using the model_geometry NCDFs?
Thank you very much! I appreciate your work!

Best,
Phillip

anoukvlug:
Dear @phiscu,

Currently there is no way to do that with OGGM yet. However I'm working on a PR to include that. It should be included in the coming ~weeks/month. If you need it earlier, just me send message.

Cheers,
Anouk

phiscu:
Thanks for you instant reply Anouk! Honestly, I'd love to use it right now as I currently have no other idea how to solve this. But I don't want to bother you with my problem. Whenever you are ready, I am happy to apply it.

Best,
Phillip

anoukvlug:
Hello Phillip (@phiscu), 
I'm sorry for my slow response. It think I have what you need. However it still needs some changes to make it suited to be merged into the OGGM code base. If you like to use it now, I think it would be easiest to have a short video call in which I can explain how you can use the code, and what the limitations are of redistributing the data from the flowline after a simulation. If you would like that, please contact me directly or send an e-mail to info@oggm.org.

fmaussion:
@anoukvlug , why not opening a PR right away? We don't have to merge yet, and PRs are the best to iterate on ideas. Maybe this will speed things up and its good if people can join in on the conversation on how to do it.

anoukvlug:
I didn't do that yet, because I put everything in a notebook. I can copy the code to a better suited location and make a PR :)

fmaussion:
Ah OK - but then just open a PR with a notebook in it! It doesn't matter at all.

PR #1490, opened 2022-10-04
Extrapolate from flowline

anoukvlug:
This PR contains a notebook that shows an option for extrapolating the glacier covered area, simulated along the flowline again to a 2-D grid. In addition it re-distributes the ice thickness. The notebook is far from perfect. Some things to be addressed are the time it takes to run the notebook, which is extremely long, so suggestions regarding that are especially welcome. Also the resolution is still quite course, making it finer would be an option once the run time has decreased. 

Generally speaking I'm looking for constructive feedback and discussing the different options. (This notebook isn't meant to be merged.) 

This issue addresses https://github.com/OGGM/oggm/issues/1448


fmaussion:
Maybe some collective knowledge will help ;-) 

cc @phiscu @lilianschuster @pat-schmitt @LvdLaan

lilianschuster:
write down if you have a look at it (it would be bad if there are two people working on it at the same time). I won't look at it this week but could try next week.

fmaussion:
Yes - the other day I also had an even simpler idea for this, lets discuss all three of us tomorrow or another day maybe

pat-schmitt:
To add to my comment above, if we want to vectorize some functions for real, we could have a look at Numba https://numba.pydata.org/ ;)

fmaussion:
yes, although this should really just be the last step, done after the possible optimisations are done

fmaussion:
Two papers which might explain how the GloGEM model does the extrapolation (at least according to a text I just read):

- doi:10.5194/esurf-10-723-2022
- Compagno, L., Huss, M., Zekollari, H., Miles, E. S., & Farinotti, D. (2022). Future growth and decline of high mountain Asia's ice-dammed lakes and associated risk. Communications Earth & Environment, 3(1), 1-9

Can someone skim them through and check how this is done there?

pat-schmitt:
I understand that they extrapolate the ice thickness change of individual elevation bands to the corresponding 2D pixels. The idea is that each _oggm-flowline_ grid point has 2D pixels assigned. With this, the elevation change of the _oggm-flowline_ could be extrapolated on the belonging 2D pixels. 

What I think is needed for this:

- Initial thickness map of the 2D glacier, with glacier bed
- convert the 2D glacier (outline & DEM) into elevation bands and save for each pixel to which elevation band it belongs. This is already done in elevation_band_flowline and just needs to be stored in the gdir (tricky could be the conversion with fixed_dx_elevation_band_flowlines). Also, save the _initial-oggm-flowline_ for later.
- sort the pixels inside of each elevation band, e.g. order them by initial thickness and distance from the outline. This defines an order in which pixels inside the elevation band should melt first (this could be useful to conserve area and volume, see below)

All these steps only need to be done once in a postprocessing step. In the end, we have for each 2D pixel the initial thickness, the glacier bed, the belonging grid point of the _oggm-flowline_, and the ranking inside the elevation band.

The actual extrapolation could look like this:

1. extrapolate the thickness change (compared to the _initial-oggm-flowline_) of all _oggm-flowline_ grid points onto the 2D map
2. check for area conservation in each elevation band (e.g. add or remove grid cells following the internal elevation bands pixel ordering)
3. check for volume conservation in each elevation band (e.g. add or remove delta volume evenly over all grid cells without creating or deleting any grid cell)

I think the actual extrapolation could be quite fast in the end and could be done for any timestep without the need of computing preceding glacier states.

Note: this method only works if the resulting glacier is smaller than the initial glacier. If we also want to be able to advance we need to think of connecting ice-free 2D pixels to the _oggm-flowline_ in some way...

fmaussion:
I think I also understand what Anouk meant that one cannot preserve area and volume easily - do they mention this?

pat-schmitt:
No, they basically only mention the extrapolation in one sentence, with not much detail:

- '_For computational reasons, the model is discretized into 10m elevation bands but results on area and thickness changes in individual bands are extrap- olated to the same 10×10m grid_'
- '_Combining the surface DEM with the modelled rate of ice thickness change, we model the glacier evolution in three dimensions_'

pat-schmitt:
I think with the idea I had over the weekend it is possible to preserve area and volume per elevation band (first add/delete pixels to match the area, second add/remove delta volume over remaining pixels without deleting any pixel)

anoukvlug:
Thanks for all the ideas and feedback! I will now close this PR, as I just created a fresh one.

PR #1491, opened 2022-10-05
adding trapezoidal downstream line

pat-schmitt:
In this PR I add an option to define a trapezoidal downstream line instead of a parabol one. The shape can be selected with `cfg.PARAMS['downstream_line_shape']` with options `'parabola'` or `'trapezoidal'` before calling `init_present_time_glacier(gdir)`. The trapezoidal shape is defined in a way that the valley cross section area is the same between the parabola and the trapezoidal when looking at the terrain height (which is defined by the DEM).

- [x] Tests added/passed
- [ ] Fully documented


pat-schmitt:
Can I merge? The error in the test is unrelated to my changes (I only added an entry in whats-new in the last commit).

The error has something to do with the retriggering of the download from the prepro_base_url (https://github.com/OGGM/oggm/blob/master/oggm/tests/test_utils.py#L807).

PR #1496, opened 2022-10-12
add option to plot flowline velocity on map

pat-schmitt:
In this PR I added an option to `graphics.plot_modeloutput_map` to be able to plot the flowline velocities onto a 2D map. This is not a real distributed 2D velocity field! It is just a nice representation of the flowline velocities we get from the model (similar to what is done for the thickness in (e) and (f) in this figure https://docs.oggm.org/en/latest/_images/ex_workflow.png).

Should I add a test figure for this new option, or is the already implemented test for the thickness plot sufficient? If I should add a test for the velocity this could probably be done by returning one plot with two subplots for the thickness and the velocity in the test https://github.com/OGGM/oggm/blob/master/oggm/tests/test_graphics.py#L437.

- [ ] Tests added/passed
- [ ] Fully documented


fmaussion:
Thanks! Not need for a test, no

fmaussion (approved):

PR #1500, opened 2022-10-26
Init glacier directories for rgi7

afisc:
@fmaussion 
adapted the `init_glacier_directories()` code so that it's able to handle the new **RGI7** format as well.

### Notes:
-currently `cfg.PARAMS['use_intersects'] = False` has to be set when working with **RGI7** entities, as no intersects files are available.
-Glacier Type, Terminus Type and Status (from RGI6) are currently all set to the flag '0' when working with **RGI7** entities.


fmaussion:
Thanks @afisc - what we are missing are tests here but we can do that later.

Issue #1360, opened 2022-01-11
Depending on the 'prepro-border' the number of trapezoid grid points changes for some elevation-band glaciers


pat-schmitt:
For some glaciers the number of trapezoid grid points of the `model_flowlines` change depending on the `prepro-border`. For the elevation band flowlines it is expected (as far as I know/believe) that all grid points with ice are trapezoid. However for the example below when the `prepro_border` is set to 80 there are 29 trapezoid grid points, but 34 grid points with ice (when a `prepro_border` of 160 is used there are the expected 34 trapezoid grid points).

Here is the example:

```python
from oggm import cfg, workflow, utils
cfg.initialize()
cfg.PATHS['working_dir'] = utils.gettempdir(dirname='OGGM-GettingStarted', reset=True)

from_prepro_level = 3
base_url = 'https://cluster.klima.uni-bremen.de/~oggm/gdirs/oggm_v1.4/L3-L5_files/CRU/elev_bands/qc3/pcp2.5/no_match/'


def print_trap_grid_points(prepro_border):
    gdirs= workflow.init_glacier_directories(['RGI60-16.02444'], # Artesonraju
                                         from_prepro_level=from_prepro_level,
                                         prepro_base_url=base_url,
                                         prepro_border=prepro_border)
    fl = gdirs[0].read_pickle('model_flowlines')[0]
    print(f'prepro_border {prepro_border}: \n'
          f'Trapezoid grid points: {sum(fl.is_trapezoid)}\n'
          f'Grid points with ice: {sum(fl.thick>0.)}\n\n')


prepro_border = 80
print_trap_grid_points(prepro_border)  

prepro_border = 160
print_trap_grid_points(prepro_border) 
```

PR #1502, opened 2022-11-07
adapt inversion flux and filter_inversion_output

pat-schmitt:
Here I add an adapted version of `filter_inversion_output`.

The old implementation always smooths with a parabolic bed shape and overwrites the original cross-section shape. Further no smoothing was conducted if the '_smoothed-parabolic bed shape factor_' was smaller than the minimum (`cfg.PARAMS['mixed_min_shape']`).

Now the adapted version preserves the cross-section shape. Additionally, if the '_smoothed-parabolic bed shape factor_' is smaller than the defined minimum the cross-section shape is changed to a trapezoid (similar to what is done during the inversion with mixed bed shapes).

I opened this PR for discussion and to see which tests need to be adapted.

Further, the old implementation of this function probably causes the problem described in https://github.com/OGGM/oggm/issues/1360, as the difference in the grid points is 5 grid points, which is the number of grid points which are smoothed and hence forced to a parabolic shape with the old function. And probably depending on the prepro-border the bed shapes of the downstream line are different (which partly decides if the inversion output is smoothed or not). But this needs further investigation.

- [x] Tests added/passed
- [ ] Fully documented


fmaussion:
@pat-schmitt sorry for changing my mind over and over about this topic, but looking at the cross sections without filter I just think it still makes more sense to use it in the "default" workflow, until COMBINE becomes the default. 

Would you be willing to continue this PR by adding the filter again in the HEF test case and try it on a few other glaciers with other bedshapes to see that it more or less works as expected? 

I think that the new default should be to filter the inversion output and use dynamical spinup where appropriate

fmaussion:
For reference here is the section without filter:

image


And with filter:

image

Of course the filter was designed with Hintereisferner in mind, but maybe we can find something which "works" more or less, and by "working" i mean just not having crazy bed deepenings at the tongue...

pat-schmitt:
Here are five examples of the new filter I propose (not committed right now). The last example just shows that the filter is doing nothing if not necessary.

What do you think? Are there other glaciers I should test? Should I also include more test graphics or is HEF enough?

For Centerlines:

centerlines_filter

For Elevation bands:

elev_bands_filter

fmaussion:
This looks great! Let's go for it.

pat-schmitt:
@fmaussion we discussed that I should try to delete the artificial increase of the ice flux at the last grid point (https://github.com/OGGM/oggm/blob/master/oggm/core/inversion.py#L116), now I was wondering if I also should delete the same thing for the tributaries just below (https://github.com/OGGM/oggm/blob/master/oggm/core/inversion.py#L120) or only for the main flowline.

fmaussion:
@pat-schmitt yeah delete there also  - better not inventing stuff

fmaussion:
@pat-schmitt I trust your judgment here. Either put back the flux at 1/10, or maybe (probably better) find a way to make it look better using your idea of prolongating the immediate downstream slope? 

As a general rule I think the prolongation of the downstream slope to avoid deepenings or bumps is very clever.

pat-schmitt:
I tested my new idea of the implementation of the inversion flux now (see two examples below). If the terminus is very thin (first row) there is still a small artefact present. However, it looks better at a 'thicker' terminus (second row, right subplots), without the need for an artificial flux.

With this, I will now start fixing the tests (the idealized tests passed right away).

compare_inversions

fmaussion:
:+1: perfect - any artifact at the tongue will have to be addressed with filtering anyway. Thanks!

pat-schmitt:
I managed to fix all tests besides one (and the graphics).

Something with the correction of negative fluxes is not working out in the `mu_star_calibration` (when an individual mu* is calibrated for 'negative-flux flowlines'). In the code (https://github.com/OGGM/oggm/blob/master/oggm/core/centerlines.py#L286), it says 'we can cope with one negative flux grid point', but I do not understand how this is handled. I understand the recursive calibration (https://github.com/OGGM/oggm/blob/master/oggm/core/climate.py#L1121) starting from the top with fixing the individual mus*, but why is one negative flux grid point ok? Has this one negative flux grid point zero thickness after the inversion?

Besides this: for calving flowlines, I decided to use the old implementation so that the thickness of the last grid point is consistent with the calving parametrization.

fmaussion:
> but why is one negative flux grid point ok? Has this one negative flux grid point zero thickness after the inversion?

I think it was arbitrary - sometimes the last grid point was negative for numerical reasons, and it will be clipped to zero and I found this not too bad. I think you can take another decision here.

> Besides this: for calving flowlines, I decided to use the old implementation so that the thickness of the last grid point is consistent with the calving parametrization.

OK

fmaussion (approved):
I skimmed through the code  - this looks good I think?

PR #1506, opened 2022-11-17
Updated define_glacier_region to handle a list of DEM sources

d-otto:
Still a bit new to PR's, apologies if I missed something.

The docs for `tasks.define_glacier_region` say `source` accepts a string or list of strings, but it was not implemented. Passing a list to `source` now tries listed DEM sources until a valid one is found. 

For example, I can now prefer Alaska DEM where it is available :)


fmaussion:
Thanks! Thats great

PR #1507, opened 2022-11-22
adding semi-implicit scheme

pat-schmitt:
In this PR I added the semi-implicit numeric scheme developed by @dngoldberg.

I also added the scheme to the idealised test cases. Further, I compared the FluxBasedModel and the SemiImplicitModel on a mixed flowline with different lambdas along the flowline (including a mixture of rectangular and trapezoidal).

Some comments and things to discuss:

- in `test_min_slope` you can nicely see the instability differences in the slope plot
- in `test_cliff` the volume difference to the MUSCLSuperBeeModel is only the half for the SemiImplicitModel compared to the FluxBasedModel, but it looks like their are some instabilities you can see in the surface_h plot for the SemiImplicitModel, maybe something to have a closer look
-  I have not implemented a MassConservationChecker (as I check all the time against the FluxBasedModel, which is checked), would this be a good idea?
- ~~I have not included a test for the sliding parameter~~
- ~~I have not included tests on real glacier geometry (could also include sliding tests there), what would be the most important real geometry tests?~~
- the used cfl number right now is 0.6 (I came up with this by testing on an unstable glacier for COMBINE), but maybe also would need more testing/justification
- ~~currently, some diagnostics are calculated each timestep, which are actually not needed. They could be calculated somewhere else for increasing the performance~~

ToDo:
- [x] idealised Tests added/passed
- [x] real geometry Tests added/passed
- [ ] Fully documented

### DIFF
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -62,6 +62,8 @@ Enhancements
 - Added option to plot flowline velocities in ``graphics.plot_modeloutput_map()``
   (:pull:`1496`)
   By `Patrick Schmitt <https://github.com/pat-schmitt>`_
+- Added SemiImplicitModel for a single trapezoid or rectangular flowline developed by `Dan Goldberg <https://github.com/dngoldberg>`_ (:pull:`1507`)
+  By `Patrick Schmitt <https://github.com/pat-schmitt>`_
 
 Bug fixes
 ~~~~~~~~~
--- a/oggm/core/flowline.py
+++ b/oggm/core/flowline.py
@@ -16,6 +16,7 @@ import warnings
 import numpy as np
 import shapely.geometry as shpg
 import xarray as xr
+from scipy.linalg import solve_banded
 
 # Optional libs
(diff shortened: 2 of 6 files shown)


pat-schmitt:
Now the diagnostics are calculated using @property decorators to avoid unnecessary calculations during the dynamic run. This is needed because some diagnostics are not computed during the dynamic run (as is the case for FluxBasedModel).

pat-schmitt:
I now added tests using HEF.

For this, I adapted the test glacier directory (`init_hef()`) to be able to select which flowline type you want to use for the test. If you want to use the elevation bands in your tests you can now use `hef_elev_gdir`.

Probably some of my included tests need to be adapted when `filter_inversion_output` is not used anymore (see https://github.com/OGGM/oggm/pull/1502)

fmaussion (approved):
Fantastic! Please just check camelcase parts of code and this is good to go - can't wait to use it!!!!

PR #1510, opened 2022-12-11
Fix issue with TANDEM DEM and minor RGI7 issue

fmaussion:
Thanks @afisc for the report.


fmaussion:
OK so this solves some of the errors we had in downloading certain DEMs for RGI7. 

I'm still in contact with Copernicus DEM support about some missing COPDEM30 files, but this should be ready soon. I'll keep you updated @afisc

PR #1514, opened 2022-12-14
Small dynamic spinup adaption

pat-schmitt:
I realized that after `run_dynamic_mu_star_calibration` the resulting model run is not always starting from the given start year. The reason for this is that the dynamic spinup tries out shorter spinup periods (= later start years) if the glacier is becoming ice-free or growing out of the boundary (e.g. the idea is if the spinup period is shorter the glacier has less time to grow out of the boundary). But for analysis, it would be nice to have a time series of all glaciers which start at the same defined year.

Therefore I introduce here a new option to deal with this in `run_dynamic_spinup`, which is the new default in `run_dynamic_mu_star_calibration`: If the period must be shortened we extend the time series with a shorter fixed-geometry-spinup.

- [x] Tests added/passed
- [ ] Fully documented (in Tutorial)

### DIFF
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -62,6 +62,10 @@ Enhancements
 - Added option to plot flowline velocities in ``graphics.plot_modeloutput_map()``
   (:pull:`1496`)
   By `Patrick Schmitt <https://github.com/pat-schmitt>`_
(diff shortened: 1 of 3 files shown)


fmaussion (approved):
Thanks!

PR #1523, opened 2023-01-17
added implicit model with trapezoidal downstream line to run_prepro_levels

pat-schmitt:
In this PR I add the implicit model together with the trapezoidal downstream line as an option for `run_prepro_levels`. With this, it is now possible to compare the FluxedBasedModel with the SemiImplicitModel on a large scale.

I added tests, which are currently failing because of `filter_inversion_output` but should pass once https://github.com/OGGM/oggm/pull/1502 is merged.

- [x] Tests added/passed
- [ ] Fully documented

### DIFF
--- a/oggm/cli/prepro_levels.py
+++ b/oggm/cli/prepro_levels.py
@@ -80,6 +80,7 @@ def run_prepro_levels(rgi_version=None, rgi_reg=None, border=None,
                       match_regional_geodetic_mb=False,
                       match_geodetic_mb_per_glacier=False,
                       evolution_model='fl_sia',
+                      downstream_line_shape='parabola',
                       centerlines_only=False, override_params=None,
                       add_consensus=False, start_level=None,
                       start_base_url=None, max_level=5, ref_tstars_base_url='',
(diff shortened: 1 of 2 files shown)


pat-schmitt:
Should I add this to whats-new? I think no other changes of `run_prepro_levels` are added there.

fmaussion:
No need I think! Thanks!

pat-schmitt:
It seems something is broken not related to this PR...

fmaussion:
@TimoRoth this looks like bad builds errors - can you have a look?

TimoRoth:
The same errors are popping up on all other builds as well.
Unfortunately, I have no idea where they originate from, or what it's even trying to tell me with `TypeError: No matching signature found`

Given it originates from scipy, my only interpretation is "scipy broke something major", if intentional and we have to adapt _something_, or unintentional... I don't know.

fmaussion:
I fixed a signature problem in scipy in salem a few weeks ago. Pinning scipy to a lower version should help

TimoRoth:
What scipy version broke it, and how do we know if the issue got fixed?
I'd like to avoid permanently pinning stuff to old versions, since in the long run that will make maintaining the images near impossible.

Did they change some behaviour? If so, we'll probably have to adapt to it, right?

fmaussion:
Yeah I don't know, in salem I had to force a certain datatype to be use, i.e. scipy stopped accepting float32 for a specific function: https://github.com/fmaussion/salem/pull/228/files 

I really think its a problem in scipy or how scipy is built, because I see no reason why we would have to change the datatype. googling "no signature found" results in quite some matches, but somehow not exactly what we have. I don't understand....

pat-schmitt:
Ok. Yes, it should be fine as all tests passed previously and I just resolved a conflict in a docstring in the last commit.

PR #1527, opened 2023-01-27
New MB model calibration architecture

fmaussion:
This opens the door to add @lilianschuster 's MB sandbox tools.

Currently only with Monthly model

I added a flexible MB calibration scheme, including the infamous three-step calibration


fmaussion:
This has become much larger than anticipated, but it feels good. The decommissioning of mu* is now!

fmaussion:
Some todos:
- [ ] `override_missing` kwarg in the new calibration could accomodate with a dict or a dataframe of values to allow for arbitrary datasets
- [ ] docstrings of the most important functions: calibration, MB models, run_
- [ ] thinking about the naming of functions and parameters really hard as we wont change them back afterwards

Issue #1540, opened 2023-03-08
volume_m3_min_h  & area_m2_min_h always stored in latest OGGM 


lilianschuster:
After the latest OGGM developments, the variables `volume_m3_min_h` and `area_m2_min_h` are always stored, even if using `cfg.PARAMS['store_diagnostic_variables'] =  ['volume', 'volume_bsl', 'area', 'length']` .  As the variables are always created if 'volume' or 'area' is in the wished stored diagnostic variables (see e.g. https://github.com/OGGM/oggm/blob/master/oggm/core/flowline.py#L1109) , we can not easily remove them. 
A solution could be to:
- only store them if they are explicitly inside `cfg.PARAMS['store_diagnostic_variables']` (to save memory)
- note that at the moment, `volume_m3_min_h` and `area_m2_min_h`  are not inside of the default  `cfg.PARAMS['store_diagnostic_variables']`. When does the dynamic spinup need the two variables from the future runs?

@pat-schmitt , I think this is related to the dynamic spinup, can you have a look at it?

pat-schmitt:
Yes, this variables are only needed for the dynamic spinup and only for the past period. I will adapt this, so that they are only stored for dynamic spinup runs or if explicitly added.

PR #1541, opened 2023-03-08
rearanged use of area_min_h

pat-schmitt:
Here I adapt the usage of the minimum ice thickness for the dynamic spinup. Now the filter is only applied to area (not volume) and the variable `area_m2_min_h` is only saved during the dynamic spinup run or if a user adds it explicitly to `cfg.PARAMS['store_diagnostic_variables']`.

Thanks, @lilianschuster for bringing this up.

Closes https://github.com/OGGM/oggm/issues/1540

- [ ] Tests added/passed
- [ ] Fully documented


fmaussion:
thanks! I'm not convinced yet this is the best solution but I also don't have a much better idea right now

PR #1547, opened 2023-03-15
corrected glacier gridpoints on flattened files

lilianschuster:
This PR uses the now updated flattened GSWP3-W5E5 and ISIMIP3b files (version 2023.2). In the previous version, i.e., at the moment still in here https://cluster.klima.uni-bremen.de/~oggm/test_climate/gswp3-w5e5/flattened/monthly/ 
and https://cluster.klima.uni-bremen.de/~oggm/cmip6/isimip3b/flat/monthly/, there are around 11900 glaciers that did not select the nearest gridpoint (due to an issue in the glacier gridpoint selection of these files). 
- The reason why this only got visible now is that I only tested it for Hintereisferner, where it worked. So it can be important to test on more glaciers. In that case, a simple test on all glaciers would have been necessary.  

- I added a test where I checked it for three glaciers. This test only works with the new (test) flattened files. 
- I also checked for every glacier, if the right gridpoints are now selected. The tests need to download quite some data and take time, so they are just here: https://nbviewer.org/urls/cluster.klima.uni-bremen.de/~lschuster/example_ipynb/flatten_glacier_gridpoint_tests.ipynb?flush_cache=True

The scripts to create these flattened glacier gridpoint files are 
- https://cluster.klima.uni-bremen.de/~lschuster/isimip3b_flat/_scripts/
- https://cluster.klima.uni-bremen.de/~lschuster/isimip3a/_script/

Thanks, @fmaussion, for reporting!

- [x] Tests added/passed

When we are sure that everyone updated to v1.6.1, we should remove the old files with the missing gridpoints.

### DIFF
--- a/oggm/tests/conftest.py
+++ b/oggm/tests/conftest.py
@@ -174,7 +174,6 @@ def secure_url_retrieve(url, *args, **kwargs):
                         'no_match/RGI62/b_040/{}/RGI60-15/RGI60-15.13.tar')
 
     assert ('github' in url or
-            'cluster.klima.uni-bremen.de/~oggm/ref_mb_params' in url or
             'cluster.klima.uni-bremen.de/~oggm/test_gdirs/' in url or
             'cluster.klima.uni-bremen.de/~oggm/demo_gdirs/' in url or
             'cluster.klima.uni-bremen.de/~oggm/test_climate/' in url or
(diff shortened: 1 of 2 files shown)


fmaussion:
Thanks @lilianschuster !

Issue #1555, opened 2023-03-28
ModuleNotFoundError: No module named 'shapely.io'


anoukvlug:
`ModuleNotFoundError: No module named 'shapely.io'` is an error that @lilianschuster found and a few of us have experienced in the mean time. Though the error message can be rather confusing, the easy solution is to update the shapely version you have installed to 2.0.

This issue of having the wrong version of shapely installed is not caught by the test suite at the moment. Therefor I think it might be good if we add/update a test to capture this.

anoukvlug:
Thank you for pointing that out! Do you mean including something like this? 

```
with _open(fp, 'rb') as f:
    try:
        out = pickle.load(f)
    except ModuleNotFoundError:
        import shapely
        if float(shapely.__version__[0:3]) <= 2.0:
            raise ModuleNotFoundError("You need to update shapely to version 2.0 or higher for this to work.")
```

PR #1556, opened 2023-03-29
Improve error message

anoukvlug:
This PR addresses issue #1555, by improving the error message that is being raised for the following error  `ModuleNotFoundError: No module named 'shapely.io'`.


TimoRoth:
Is the new notebook part of this PR intentionally?
Also, the text of the error message can change depending on the python version. And maybe even system language?
Much better to check the name property of the exception instead.

anoukvlug:
Yes, that does make sense. However I don't think I fully understand you suggestion. What do you mean with name property? 

Thanks for noticing the notebook, that was not supposed to be included in this PR.

anoukvlug:
Thanks for the explanation @TimoRoth! Is the latest commit what you had in mind?

TimoRoth:
To re-raise an exception, just call `raise` without any arguments. Otherwise you might obscure where the error actually came from.

One more issue I could see is that by creating a new ModuleNotFoundError without the name/path argument set, you might be breaking conventions.

What you could do instead:

```python
                if err.name == "shapely.io":
                    err.msg = "You need shapely version 2.0 or higher for this to work."
                raise
```

anoukvlug:
Thanks for the coding lesson. That looks indeed better :)

fmaussion:
Thanks!

PR #1562, opened 2023-04-06
adapted dynamic spinup doc

pat-schmitt:
Here I adapted the documentation page of the dynamic spinup to v1.6 and I added a flowchart of the default workflow.


fmaussion:
Thanks!

PR #1563, opened 2023-04-06
add more flexibility to compile_run_output

pat-schmitt:
In this PR I added more flexibility to `compile_run_output`. Now it is possible to compile outputs with different data variables. This is mainly needed for the new minimum thickness strategy during the dynamic melt_f calibration. But maybe could be useful in the future.

I have adapted for 2d and 3d variables (for monthly or daily hydro output) and added a test for both cases.

Closes https://github.com/OGGM/oggm/issues/1561

- [x] Tests added/passed
- [ ] Fully documented

### DIFF
--- a/oggm/core/dynamic_spinup.py
+++ b/oggm/core/dynamic_spinup.py
@@ -41,7 +41,7 @@ def run_dynamic_spinup(gdir, init_model_filesuffix=None, init_model_yr=None,
                        store_model_geometry=True, store_fl_diagnostics=None,
                        store_model_evolution=True, ignore_errors=False,
                        return_t_bias_best=False, ye=None,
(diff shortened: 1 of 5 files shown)


pat-schmitt:
With these changes, I also could delete the kwarg `make_compatible` in `run_until_and_store`.
I introduced it to compile runs with different spinup strategies together. As fixed_geometry_spinup is the fallback for dynamic spinup and has the additional data variable `is_fixed_geometry_spinup`. With `make_compatible` I added `is_fixed_geometry_spinup` as a dummy data variable to all runs, even if no fixed geometry spinup is performed. This is not necessary anymore, with the larger flexibility of `compile_run_output`.

fmaussion:
OK merging now to test - please add the remaing TODOs in subsequent PR

PR #1564, opened 2023-04-08
faster process_cmip_data: allow flattened cmip and selected time series

lilianschuster:
Allows to use flattened CMIP data in `process_cmip_data`. The option is at the moment only used for the flattened overshoot and stabilisation options of gfdl-esm4 that go until 2500. This method is probably faster than always processing the unflattened files.


lilianschuster:
allows to only process a selected time period of the GCMs for `process_cmip_data` and `process_isimip_data`. 

Due to the rolling window for the std. bias correction of the temperature, the selected time period needs to be larger by +/-  window_years_length/2 to create the exact same time series as without `y0` and `y1`.  That means, if bias corrections is done for e.g. from 2000-2019, need to set `y0=1990`

- [x] tests added

PR #1572, opened 2023-04-26
added second argument to check for the rgi7 naming convention when zi…

afisc:
@fmaussion 
creating rgitopo with rgi7 should now result in the same data structure.


fmaussion:
Thanks!

PR #1576, opened 2023-05-03
Extrapolate to grid

anoukvlug:
This PR includes a notebook that can be used to extrapolate the simulated results along the elevation band based flowline to the 2D grid. It is basically a follow-up on PR #1490, which contains a version on the code that was to slow. This new version has been created together with @pat-schmitt. 

The current format of submitting it as a notebook, has been chosen so it can more easily be checked and discussed. I would propose to add the different functions in the notebook to the OGGM code base, and re-write the notebook to be a tutorial.

Comments are welcome 😄


fmaussion:
Thanks! Will have a look a report back

PR #1578, opened 2023-05-10
Update test image

TimoRoth:
Something, in some dependency, changed again, breaking practically all tests.
Luckily they mostly all fail with the same error, in the same place.
So investigating what changed should be fairly simple.

Opening here for investigation.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -24,7 +24,7 @@ jobs:
           - workflow-multi
           - graphics-mpl
         container:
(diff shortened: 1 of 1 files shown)


pat-schmitt:
Probably the recent update of geopandas to version 0.13 is causing this error https://geopandas.org/en/stable/docs/changelog.html

This line has changed: https://github.com/geopandas/geopandas/blame/main/geopandas/geodataframe.py#L447

Should I investigate further?

fmaussion:
If you can fix it would be great!!!

PR #1579, opened 2023-05-10
Fix geopandas

pat-schmitt:
This is the first attempt to fix a bug with the updated Geopandas version.

- [x] Tests added/passed

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -24,7 +24,7 @@ jobs:
           - workflow-multi
           - graphics-mpl
         container:
-          - ghcr.io/oggm/untested_base:20230509
+          - ghcr.io/oggm/untested_base:20230323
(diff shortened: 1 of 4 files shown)


TimoRoth:
I think it'd be better to test this with the old image still in place, to make sure fixing the issue does not break old geopandas versions.

pat-schmitt:
Ok, I have undone your changes of the test image

fmaussion:
Thanks!! @TimoRoth can you check on the new images?

TimoRoth:
For some reason I can't, any attempt to push leaves me with a server 
error, and going to my repository also presents me with a similar error.

Gonna have to wait until Github fixes itself.

pat-schmitt:
I just saw I have forgotten one fix. Will open another PR.
One test failed in master after merging, but not here at the last commit. Are the new images automatically in place when testing master after a merge or why are there these differences?

PR #1583, opened 2023-05-17
update the link to the leclerq dataset in the get_ref_length_data fun…

anoukvlug:
The  `.get_ref_length_data` function is not working for RGI6. Therefore I would suggest updating it. This PR is just a quick fix. If it would be better to still be compatible with RGI5, just let me know, than I will update the PR. 

@juliaeis, thanks for helping me finding this function and solving the issue!


fmaussion:
Thanks @anoukvlug !

Issue #1582, opened 2023-05-16
TODOs for 1.6.1


fmaussion:
- [x] Bug with water level : https://github.com/OGGM/oggm/pull/1589
- [ ] ~~Change default proj to UTM~~ : not going to happen in 1.6
- [x] Lower the tolerance for matching dhdt in the spinup: https://github.com/OGGM/oggm/pull/1588
- [x] Negative melt : https://github.com/OGGM/oggm/pull/1584 
- [x] Redo quite a few prepro dirs 
- [x] Redo global projs ..
- [x] ... and release them "officially"

PR #1584, opened 2023-05-17
move negative melt_on_glacier to snowfall

lilianschuster:
In hydrological projections, negative melt_on_glacier values occurred in the past. Although the absolute values were mostly relatively small, negative melt_on_glacier (and consequently negative glacier runoff) does not makes sense. 

With this pull request, no negative `melt_on_glacier` values should exist anymore. The previously negative values are now clipped to zero, and the negative values are added as absolute values to the `snowfall_on_glacier` (for mass-conservation issues). 

- [ ] There is one line in a test that still fails: https://github.com/OGGM/oggm/blob/6952200852e43ee69ba43563fdb4a8de9a3ea165/oggm/tests/test_models.py#L4311. This is because the total precipitation changes even in a constant climate (because of adding the negative values as absolute values to the `snowfall_on_glacier` variable). I am not sure if this needs to stay constant. @fmaussion, what do you think?  
 
- [ ] I could have a quick check if this makes a difference in the run_with_hydro output (by doing projections for one basin). And test again if the `melt_on_glacier` is now everywhere non-negative. 

This PR is part of the ToDo-List of  https://github.com/OGGM/oggm/issues/1582 

- [x] Tests added/passed (except for one line)

### DIFF
--- a/oggm/tests/test_models.py
+++ b/oggm/tests/test_models.py
@@ -4308,6 +4308,9 @@ class TestHydro:
                            odf['liq_prcp_on_glacier'] +
                            odf['snowfall_off_glacier'] +
                            odf['snowfall_on_glacier'])
+        # this test fails because snowfall_on_glacier changes, as the
+        # formerly negative melt_on_glacier was added to snowfall_on_glacier
+        # is this a strict test??? (the remaining part of the test does not fail)
         assert_allclose(odf['tot_prcp'], odf['tot_prcp'].iloc[0])
 
         # So is domain area
@@ -4337,9 +4340,15 @@ class TestHydro:
         # At the very first timesep there is no glacier so the
         # melt_on_glacier var is negative - this is a numerical artifact
         # from the residual
-        assert_allclose(odf['melt_on_glacier'].iloc[0],
-                        - odf['residual_mb'].iloc[0])
-
+        # assert_allclose(odf['melt_on_glacier'].iloc[0],
+        #                - odf['residual_mb'].iloc[0])
+        # we changed, that, if negative, the absolute value should go to snowfall_on_glacier
+        # let's check that (this happens also at other times than the first step,
+        # but only in the first step it always happens)
+        assert_allclose(odf['snowfall_on_glacier'].iloc[0],
+                        odf['residual_mb'].iloc[0])
+        # check if melt on glacier is always above or equal zero
+        assert np.all(odf['melt_on_glacier'] >= 0)
         # Now with zero ref area
         tasks.run_with_hydro(gdir, run_task=tasks.run_constant_climate,
                              store_monthly_hydro=store_monthly_hydro,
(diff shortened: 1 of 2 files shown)


oggm/tests/test_models.py
```diff
@@ -4558,6 +4581,8 @@ def test_hydro_out_past_climate(self, hef_gdir, inversion_params, store_monthly_
             # Runoff peak should follow a temperature curve
             assert_allclose(odf_ma['melt_on_glacier'].idxmax(), 8)
 
+        # check if melt on glacier is always above or equal zero
+        assert np.all(odf['melt_on_glacier'] >= 0)
```
fmaussion, line 4585: line is missing here

oggm/tests/test_models.py
```diff
@@ -4308,6 +4308,9 @@ def test_hydro_out_from_no_glacier(self, hef_gdir, inversion_params, store_month
                            odf['liq_prcp_on_glacier'] +
                            odf['snowfall_off_glacier'] +
                            odf['snowfall_on_glacier'])
+        # this test fails because snowfall_on_glacier changes, as the
+        # formerly negative melt_on_glacier was added to snowfall_on_glacier
+        # is this a strict test??? (the remaining part of the test does not fail)
```
fmaussion, line 4313: Yes it's ok to remove the first time step - maybe just check that tot_prcp is constant after the first year

oggm/tests/test_models.py
```diff
@@ -4874,7 +4902,9 @@ def test_hydro_out_random(self, hef_gdir, inversion_params, store_monthly_hydro)
         frac = odf['residual_mb'] / odf['melt_on_glacier']
         assert_allclose(frac, 0, atol=0.044)  # annual can be large (prob)
 
-    @pytest.mark.slow
+        # check if melt on glacier is always above or equal zero
+        assert np.all(odf['melt_on_glacier'] >= 0)
```
fmaussion, line 4906: Line missing here as well

fmaussion (approved):
Thanks lily - I didn't check the maths, but its probably worth checking on one region or two if it changes the results

oggm/tests/test_models.py
lilianschuster: ok, I only check now that tot_prcp is constant after the first year. This works, but I don't know if it always works. If `melt_on_glacier` is negative after the first year, then the tot_prcp wouldn't be constant anymore. But I guess in this idealised case, due to the constant climate, the `melt_on_glacier` also does not get negative?

lilianschuster:
I checked whether the results changed between the versions. 
- as expected, `melt_on_glacier` is now always non-negative (checked in RGI reg 11, 06, 15) 
- the new `melt_on_glacier` is thus always equal or larger than the old one, and the new `snowfall_on_glacier` is also equal or larger than the old one
- (same for the monthly values)

For single glaciers and years, the runoff can be visibly different. Over the sum of an entire RGI region, the differences are almost not visible (although the relative differences can be substantial due to sometimes small absolute numbers, specifically when comparing the mean monthly cycle of  `melt_on_glacier` and `snowfall_on_glacier`). 

(checks are in https://nbviewer.org/urls/cluster.klima.uni-bremen.de/~lschuster/runs_oggm_v16/compare_negative_melt_on_glacier_snowfall_on_glacier.ipynb )

PR #1595, opened 2023-06-12
adding dhdt flowline diagnostics

pat-schmitt:
Here is the first version for adding new flowline diagnostics dhdt, climatic mb and flux-divergence.

I decided to calculate dhdt after each year directly along the flowlines and also read out the forcing mass balance (in m yr-1). The flux-divergence (in m yr-1) is calculated using the formula `dhdt = mb + flux_divergence`. I decided to use this equation to make the interpretation of terms more straightforward (e.g. positive calculated flux-divergence means the in-flux is larger then the out-flux), as the original equation includes a minus sign (see e.g. https://docs.oggm.org/en/latest/ice-dynamics.html#ice-flow).

The advantage of this method is that it works for all bed shapes and all numeric solvers. Regarding testing, I am not sure if my very simple test is enough...

Special case retreating glaciers:
However, with this method, we can not calculate the flux-divergence exactly if a grid point at the terminus is getting ice-free. At such points, the 'very' negative climatic mass balance is not compensated by the flux-divergence and we only can calculate a value which tells us the flux-divergence at this point was smaller. To avoid unreasonable spikes in the calculated flux-divergence at such points I decided to smooth these values with a factor of 0.1.

Here are two plots of the resulting terms for a retreating and an advancing glacier, forced by a random mb:

Retreating:

retreating

Advancing:

advancing

(Sidenote: In year 1 you can observe an initial shock if starting from a surface not dynamically initialised).

- [x] Tests added/passed
- [ ] Fully documented

### DIFF
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -15,6 +15,15 @@ Bug fixes
   special data variables (e.g. ``is_fixed_geometry_spinup``) (:pull:`1563`).
   By `Patrick Schmitt <https://github.com/pat-schmitt>`_
 
+Enhancements
+~~~~~~~~~~~~
+
+- Added three new flowline diagnostic variables: thickness change in one year
+  (``dhdt``), forcing climatic mass balance (``climatic_mb``) and flux divergence
+  (``flux_divergence``). All variables are in units meter of ice per year
+  (:pull:`1595`).
+  By `Patrick Schmitt <https://github.com/pat-schmitt>`_
+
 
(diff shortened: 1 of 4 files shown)


pat-schmitt:
At the moment the new diagnostic variables are added to the flowline diagnostics by default. Should I change this?

pat-schmitt:
I opened an issue to add some documentation for this to the tutorials. I think it is ready to merge.

PR #1596, opened 2023-06-13
fixed shapely > v0.2.0 in multipolygon_to_polygon

pat-schmitt:
I found a small bug when using shapely versions > 0.2.0. Their `np.array(MultiPolygon)` is not supported anymore. I updated it in `multipolygon_to_polygon`. Not sure if there are other locations which need updates.

- [ ] Tests added/passed

### DIFF
--- a/oggm/utils/_funcs.py
+++ b/oggm/utils/_funcs.py
@@ -538,7 +538,13 @@ def multipolygon_to_polygon(geometry, gdir=None):
     rid = gdir.rgi_id + ': ' if gdir is not None else ''
 
     if 'Multi' in geometry.geom_type:
-        parts = np.array(geometry)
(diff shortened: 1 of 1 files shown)


pat-schmitt:
This is only important if you want to use your own inventory, where some glaciers have a Multipolygon-outline.

Should I add a dummy glacier with a Multipolygon-outline to the sample data (https://github.com/OGGM/oggm-sample-data/tree/master/test-files/cgi2), to test for this in the future (https://github.com/OGGM/oggm/blob/master/oggm/tests/test_utils.py#L352)?

pat-schmitt:
As we discussed during the oggm meeting no additional tests are needed. So I think it is ready to merge.

PR #1600, opened 2023-06-26
added more flexiblity to dynamic spinup

pat-schmitt:
Here I added more flexibility to the dynamic spinup function.

The old implementation only allowed a dynamic spinup matching the area or the volume of the provided initial flowline at the rgi date. This is still the default behaviour.

Additionally, with this PR, it is now possible to provide a freely changeable target year with the target value. This is probably not useful for global oggm runs but can be handy for local studies where more data is available (e.g. a second oultine).

- [x] Tests added/passed
- [ ] Fully documented (will open an issue to add this change to the tutorials)

### DIFF
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -31,14 +31,14 @@ Enhancements
   functionality is currently in the sandbox and needs to be documented.
   By `Anouk Vlug <https://github.com/anoukvlug>`_ and
   `Fabien Maussion <https://github.com/fmaussion>`_
-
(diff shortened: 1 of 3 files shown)


fmaussion:
Great! It's a bit a detail but I think I would prefer `target_yr` over `yr_target`, to read it like in English?

pat-schmitt:
Yes I agree, and I have renamed it

fmaussion:
Thanks!

PR #1604, opened 2023-06-30
Add possibility of using MultiPolygon Outlines for Elevation Band Flowline

pat-schmitt:
Here I added the possibility of using MultiPolygon Outlines within Elevation Band Flowlines. For this, I added a new `cfg.PARAMS` variable, which must be turned on. The default behaviour of OGGM is unchanged.

This is useful for example if you are working with regional glacier inventories, where you have multiple outlines for a glacier and one glacier splits into multiple parts.

Not sure if we want to add this into OGGM. But I needed it so maybe it is also useful for others in the future...

- [ ] Tests added/passed
- [ ] Fully documented

### DIFF
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -40,6 +40,11 @@ Enhancements
   (``flux_divergence``). All variables are in units meter of ice per year
   (:pull:`1595`).
   By `Patrick Schmitt <https://github.com/pat-schmitt>`_
+- Added possibility to use MultiPolygon outlines together with elevation bands.
+  That can be useful when working with local glacier inventories with multiple
+  outlines (e.g. older outline single polygon but newer outline multi polygon for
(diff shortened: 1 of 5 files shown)


pat-schmitt:
As discussed I checked how the downstream line is handled for MultiPolygons. 

For elevation bands, the downstream line starts always from the lowest point along the outline, so it makes no difference if we compare MultiPolygons with Polygons. Here two plots for comparison:

Single Polygon Outline:
Polygon_downline

MultiPolygon Outline:
Multipolygon_downline

I will add whats-new and for the documentation, I think I could at a sentence or two to the 'Using your own inventory'-tutorial https://oggm.org/tutorials/master/notebooks/advanced/use_your_own_inventory.html#.

I plan to add no tests put make clear in the tutorial explanation that it is an untested feature and everyone who uses it must check the outcome carefully.

PR #1622, opened 2023-08-23
Update shop doc

anoukvlug:
This PR contains some suggestions for updating the shop documentation. 

I think we might need to split the shop documentation into two pages, one just on the pre-processed directories and one for all the other info. I didn't split it yet. Maybe we can shortly discuss this during the meeting.


fmaussion:
Thanks! Good edits. Sounds like a good idea to split the two.

PR #1623, opened 2023-08-23
Add the possibility to concatenate outputs in distribute_2d and small…

fmaussion:
@pat-schmitt FYI

### DIFF
--- a/oggm/sandbox/distribute_2d.py
+++ b/oggm/sandbox/distribute_2d.py
@@ -390,8 +390,8 @@ def distribute_thickness_from_simulation(gdir,
         out_thick[i, :] = new_thick
 
     with xr.open_dataset(gdir.get_filepath('gridded_data')) as ds:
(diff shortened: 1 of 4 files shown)


fmaussion:
This changes the API a bit (for the better)

oggm/sandbox/distribute_2d.py
```diff
@@ -206,14 +207,18 @@ def assign_points_to_band(gdir, topo_variable='glacier_topo_smoothed',
 
 
 @entity_task(log, writes=['gridded_data'])
```
pat-schmitt, line 209: I think this should be `'gridded_simulation'` instead of `'gridded_data'` now

fmaussion: Good catch! Fixed now. Also added more variables to the file so that it can be used by the 3dviz tool (to no success so far)

PR #1624, opened 2023-08-24
Changes docs

anoukvlug:
This PR splits the shop documentation over more pages.

### DIFF
--- a/docs/climate-data.rst
+++ b/docs/climate-data.rst
@@ -1,0 +1,149 @@
+Climate data
+============
+
+Here are the various climate datasets that OGGM can handle automatically, using the for instance
+one of the following functions to pre-process the climate data:
+
+.. code-block:: python
+
+    from oggm.tasks import process_w5e5_data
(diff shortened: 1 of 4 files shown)


docs/climate-data.rst
```diff
@@ -0,0 +1,149 @@
...
+
+.. _climate-w5e5:
+
+W5E5
+~~~~
+
+As of v1.6, W5E5 is the standard dataset used by OGGM as reference. All
+preprocessed directories use it.
+
+TODO: explain what W5E5 is and why it is the default.
```
fmaussion, line 24: It would be good to add a little bit of text here. Can you add? The main reason it is the new default is that it is bias corrected for impact models and we can use ISIMIP forcing: https://www.isimip.org/

You can write a bit of text and ask be and Lily to review it maybe? Would be super helpful!

anoukvlug: I can make a first attempt and will ask Lily to review it :)

fmaussion:
Thanks Anouk!

PR #1630, opened 2023-08-25
Standard projections doc

anoukvlug:
Here is a suggestion for some edits to the standard projection documentation.


fmaussion:
Thanks!

Issue #1632, opened 2023-08-27
filter inversion output creates quite a lot of errors (I believe on small glaciers)


fmaussion:
Analyzing the errors using Lily's tutorial here: https://oggm.org/tutorials/master/notebooks/advanced/preprocessing_errors.html

Applied to:

url = 'https://cluster.klima.uni-bremen.de/~oggm/gdirs/oggm_v1.6/L3-L5_files/2023.3/elev_bands/W5E5_spinup/RGI62/b_160/L5/summary/'

will show the glaciers where the task creates problems.

PR #1635, opened 2023-08-28
Bug in filter_inversion_output

pat-schmitt:
filter_inversion_output through an error if the number of grid points is smaller than the number of grid points that should be used for filtering (`n_smoothing = 5` by default), reported in https://github.com/OGGM/oggm/issues/1632. I included a check for this.

- If the number of grid points is smaller `n_smoothing` and at least 3: `n_smoothing = nr_of_grid_points - 1`
- If the number of grid points is smaller 3: log a warning and do not apply the filter

We also could raise an error for flowlines with one or two grid points only...

Closes https://github.com/OGGM/oggm/issues/1632

- [ ] Tests added/passed
- [ ] Fully documented

### DIFF
--- a/oggm/core/inversion.py
+++ b/oggm/core/inversion.py
@@ -543,6 +543,18 @@ def filter_inversion_output(gdir, n_smoothing=5, min_ice_thick=1.,
     cls = gdir.read_pickle('inversion_output')
     cl = cls[-1]
 
+    # check that their are enough grid points for smoothing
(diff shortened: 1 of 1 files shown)


fmaussion:
Thanks @pat-schmitt 

@lilianschuster @anoukvlug this bug was responsible for quite a few (very small) failing glaciers in the 1.6 directories. In theory we should reprocess all of them, but we will ignore this now.

Issue #1631, opened 2023-08-25
floatyear_to_date is horribly coded


fmaussion:
Yes it is. There should be a vectorized way to do it? it just needs some time to think about it.

PR #1636, opened 2023-08-29
Vectorize floatyear_to_date

pat-schmitt:
Here is a suggestion for a vectorized form of `floatyear_to_date`.

A quick performance check with the following code, for around 12 million values to convert:
```python
def test_time(yrs):
    start_time = time.time()
    year_orig, month_orig = utils.floatyear_to_date(yrs)
    time_orig = time.time() - start_time
    
    start_time = time.time()
    year_new, month_new = new_floatyear_to_date(yrs)
    time_new = time.time() - start_time
    
    assert np.allclose(year_orig, year_new)
    assert np.allclose(month_orig, month_new)
    
    print(f'orig time {time_orig:.2f} s')
    print(f'new time {time_new:.2f} s')

monthly_time = utils.monthly_timeseries(0, 1000000)
test_time(monthly_time)
```
`orig time 8.34 s`
`new time 0.16 s`

Closes https://github.com/OGGM/oggm/issues/1631

- [ ] Tests added/passed
- [ ] Fully documented

### DIFF
--- a/oggm/utils/_funcs.py
+++ b/oggm/utils/_funcs.py
@@ -610,6 +610,11 @@ def floatyear_to_date(yr):
 
     out_y = np.floor(yr).astype(int)
     out_m = np.minimum(12, np.round(((yr - out_y) * 12 + 1)).astype(int))
+
+    if len(yr) == 1:
+        out_y = out_y.item()
+        out_m = out_m.item()
+
     return out_y, out_m
 
 
--- a/oggm/utils/_funcs.py
+++ b/oggm/utils/_funcs.py
@@ -668,25 +668,16 @@ def hydrodate_to_calendardate(y, m, start_month=None):
                                  'callers of this function to specify the '
                                  'hydrological convention they are using.')
 
+    # nothing to do if start_month is 1
+    if start_month == 1:
+        return y, m
+
+    y = np.array(y)
+    m = np.array(m)
+
     e = 13 - start_month
-    try:
-        if m <= e:
-            if start_month == 1:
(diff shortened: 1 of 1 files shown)


pat-schmitt:
I added the lines
```python
  if isinstance(yr, float):
      out_y = out_y.item()
      out_m = out_m.item()
  elif (isinstance(yr, list) or isinstance(yr, np.ndarray)) and len(yr) == 1:
      out_y = out_y.item()
      out_m = out_m.item()
  elif isinstance(yr, xr.DataArray):
      out_y = np.array(out_y)
      out_m = np.array(out_m)
```
to be consistent with the output of `calendardate_to_hydrodate` (maybe this function also could be vectorized...). Otherwise, you get an error in  `run_until_and_store` for a single year, due to this line https://github.com/OGGM/oggm/blob/master/oggm/core/flowline.py#L1031.

Further, I included a check for floating point precision now, to have the same behaviour as the previous function. This slowed down the performance and is now (for the same check a above):
`orig time 8.38 s`
`new time 0.99 s`

pat-schmitt:
Also included vectorized versions for `hydrodate_to_calendardate` and `calendardate_to_hydrodate` now.

`hydrodate_to_calendardate`:
 - `orig time 4.26 s`
 - `new time 0.15 s`

`calendardate_to_hydrodate`:
 - `orig time 4.90 s`
 - `new time 0.25 s`

fmaussion:
OK, this is scary because very low level in OGGM, but I trust that if the tests pass, we should be good 😅

PR #1640, opened 2023-09-04
Delete required_model_steps

pat-schmitt:
Here is a suggestion for deleting `required_model_steps` from `FlowlineModel` as a potential optimisation as discussed on slack. I kept some of the logic to make sure the mass balance model updates correctly depending on the defined feedback frequency.

Further, I included `max_dt` to `FluxBasedModel`, as in the original comment it states that `required_model_steps` is used "to prevent 'too large' steps". But not sure if this is really needed.

- [ ] Tests added/passed
- [ ] Fully documented


fmaussion:
Something to discuss over a drink 😅 Intuitively I would think this is a good idea, but solving all the tests seems hard?

pat-schmitt:
I will close this for now, as this idea won't be developed further in the near future due to the need to check many things for only a small reward.

Issue #1660, opened 2023-11-03
Bug in preserving totals for binning variables in elevation band flowlines


pat-schmitt:
There is currently a bug when variables are binned in the elevation bands.

The error is that if we try to conserve the total value the glacier mask is not used (in this line https://github.com/OGGM/oggm/blob/master/oggm/core/centerlines.py#L2227). To make it clear here a small example, starting from the w_data gdirs:

```python
from oggm import utils, workflow, tasks, cfg
import xarray as xr

cfg.initialize(logging_level='WARNING')
cfg.PATHS['working_dir'] = utils.gettempdir('OGGM_tmp', reset=True)
rgi_ids = ['RGI60-11.00897']
base_url = 'https://cluster.klima.uni-bremen.de/~oggm/gdirs/oggm_v1.6/L3-L5_files/2023.3/elev_bands/W5E5_spinup_w_data/'
gdirs = workflow.init_glacier_directories(rgi_ids, prepro_base_url=base_url, from_prepro_level=4, prepro_border=80)
with xr.open_dataset(gdirs[0].get_filepath('gridded_data')) as ds:
    ds = ds.load()
```

If we now plot `ds.consensus_ice_thickness.plot()`
![image](https://github.com/OGGM/oggm/assets/50843444/14d9499c-b04d-4e79-8305-bf92d06adf10)

vs `ds.millan_ice_thickness.plot()`
![image](https://github.com/OGGM/oggm/assets/50843444/1f97ed42-609a-4e59-9eae-53aa247c1b17)

You see the consensus thickness is cut out, whereas the millan thickness is not. This results in too large thicknesses in the binned flowlines for millan (you can check by calling `pd.read_csv(gdirs[0].get_filepath('elevation_band_flowline'), index_col=0)`).

Two options to solve this issue:

1. add the glacier mask in the calculation of the totals (change line https://github.com/OGGM/oggm/blob/master/oggm/core/centerlines.py#L2227 to `out_totals.append(np.nansum(data[glacier_mask]) * gdir.grid.dx ** 2)`)
2. We use that glacier mask when we add data to gridded_data in the shop

Option 1 is the easier/faster way to deal with this, but is there a reason to keep data in gridded_data outside of the outline?

fmaussion:
> but is there a reason to keep data in gridded_data outside of the outline?

In the case of the Farinotti data, the regridded data might be outside of the OGGM mask (because they used another mask / grid method as OGGM). That was the main reason.

[...]

PR #1661, opened 2023-11-03
add glacier mask for preserving totals when binning variables

pat-schmitt:
Here I add the use of the glacier mask for preserving totals when binning variables to the elevation bins. The consensus thickness is treated differently.

Closes https://github.com/OGGM/oggm/issues/1660

- [ ] Tests added/passed
- [ ] Fully documented

### DIFF
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -3,6 +3,16 @@
 Version history
 ===============
 
+v1.6.2 (unreleased)
+-------------------
+
+Bug fixes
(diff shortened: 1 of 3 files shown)


fmaussion:
This bug was bad. It is unfortunate that the 1.6.1 folders have this issue...

fmaussion:
I don't know yet how to document this issue. I think that it wouldn't be a problem for many users, but we don't know who is using these data.

PR #1667, opened 2023-12-01
Adaptations in dynamic calibration functions

pat-schmitt:
Here I am suggesting some changes in the dynamic calibration functions:

- I changed the default min ice thickness from 10 to 2. I found this in a recent study that with the old default of 10 you are always on the higher end of the target area error boundary. This makes a larger difference for smaller (thinner) glaciers, on the large scale I do not expect to see large differences. Open for discussion! (if we change this should be added to whats-new)
- I slightly changed the order in the code to avoid wrong diagnostic messages in `gdir.get_diagnostics()` if something fails. But all functionalities stay the same.
- I renamed `t_bias` to `t_spinup` in the dynamic calibration procedure, to avoid confusion with the temperature bias of the mass balance model (suggestion from @rodaguayo)

- [ ] Tests added/passed
- [ ] Fully documented

### DIFF
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -3,6 +3,29 @@
 Version history
 ===============
 
+v1.6.2 (unreleased)
+-------------------
+
+Enhancements
+~~~~~~~~~~~~
+
+- There is now a possibility for initializing a elevation-band flowline using
+  external thickness data and conduct a dynamic run with it (:pull:`1658`).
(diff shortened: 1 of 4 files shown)


fmaussion (approved):
Thanks @pat-schmitt ! I agree with the change in defaults - please add whats new and merge

pat-schmitt:
It seems this got stock with 'Processing updates'. I will close and reopen again.

Issue #1679, opened 2024-01-23
Shift in coordinates of gridded_data and dem.tif


pat-schmitt:
From my understanding, when you do `tasks.gridded_data_var_to_geotiff(gdir, 'topo')` the resulting file `topo.tif` should be exactly the same as `dem.tif`. However, when you look at the two files (e.g. in QGIS) you see that they are shifted by half a pixel.

I will add a test for this and try to find the problem in the next few days. This probably also needs to be considered in PR https://github.com/OGGM/oggm/pull/1674.

fmaussion:
the culprit is probably here: 

https://github.com/OGGM/oggm/blob/6e0fb56d92f02aa525e71daeee9be4848c81b5e5/oggm/core/gis.py#L1686 

We have to ask Salem to provide the corner coordinates, with grid.corner_grid: https://salem.readthedocs.io/en/stable/generated/salem.Grid.corner_grid.html#salem.Grid.corner_grid

PR #1682, opened 2024-01-29
Dem gridded_data shifted coordinates

pat-schmitt:
This fixes a coordinate shift between `dem.tif` and the `gridded_data.nc`.

However, there seems to be a bug in salem when opening a .tif-file using `ds = salem.open_xr_dataset(tiff_path)` and calling `ds.salem.grid` afterwards. I will open a PR for salem and link it here.

Due to this bug, one test is currently not working and I commented it out. After salem is fixed the test can be uncommented again (and maybe adapted due to floating point problems...).

Closes https://github.com/OGGM/oggm/issues/1679

- [ ] Tests added/passed
- [ ] Fully documented

### DIFF
--- a/oggm/tests/test_prepro.py
+++ b/oggm/tests/test_prepro.py
@@ -9,6 +9,7 @@ import shapely.geometry as shpg
 import numpy as np
 import pandas as pd
 import xarray as xr
+import rioxarray as rioxr
 
 salem = pytest.importorskip('salem')
 rasterio = pytest.importorskip('rasterio')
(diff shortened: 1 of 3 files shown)


pat-schmitt:
Here the PR for salem https://github.com/fmaussion/salem/pull/234

oggm/tests/test_prepro.py
```diff
@@ -515,9 +515,20 @@ def test_gridded_data_var_to_geotiff(self):
...
             gridded_topo = ds[target_var]
             gtiff_ds = salem.open_xr_dataset(gtiff_path)
-            assert ds.salem.grid == gtiff_ds.salem.grid
+            # this line is uncommented due to a bug in salem, maybe also change
+            # it to check coordinates like below (floating point issues)
+            # assert ds.salem.grid == gtiff_ds.salem.grid
             assert np.allclose(gridded_topo.data, gtiff_ds.data)
 
+        # compare coordinates of topo.tif with dem.tif
+        demtiff_ds = salem.open_xr_dataset(gdir.get_filepath('dem'))
```
fmaussion, line 524: Can we try to avoid using salem here, but use rioxarray instead? It might be possible to use rioxarray to open but still use salem's accessor.

pat-schmitt: Yes, I can use rioxarray to open the tiff files and compare the coordinates, I will change this. But I can not use salem's accessor because their is no variable `.pyproj_srs` added to the dataset when opening with rioxarray.

fmaussion: OK yes I hoped that the coords would be enough, but indeed the proj info is missing - no worries, that's enough as is for now.

pat-schmitt:
Ok, now the tiff-files are opened with rioxarray and this PR is now independent of the bug in salem and can be merged if all tests pass.
I will add this to whats-new now.

pat-schmitt:
It seems that for the minimal test `rioxarray` is not available. How can I solve this?

fmaussion:
Sorry for the mess @pat-schmitt - just ignore and keep it simple, as long as it worked locally we should be good to go

pat-schmitt:
Yes, it also worked with the other containers, so it is ready to merge.

However, we should add `rioxarray` to the container used in the minimal test or exclude this test. Otherwise, it will never turn green...

fmaussion:
Yes sorry I was not clear: adding rioxarray to minimal defies the purpose of the minimal environment (which is to have no GIS library in it). To this test should rather be moved to tests_prepro, or altered/commented for simplicity.

pat-schmitt:
I used now `rioxr = pytest.importorskip('rioxarray')` in test_prepro, so all tests using rioxarray are skipped if it can not be imported. The same is done for the other gis packages like salem, rasterio and geopandas.

fmaussion:
Thanks @pat-schmitt ! And sorry for the confusion

PR #1684, opened 2024-02-01
Change the COPDEM server to prism-dem-open.copernicus.eu

dcodrut:
I replaced the server used for downloading the COPDEM30/90 (from ftps://cdsdata.copernicus.eu to  https://prism-dem-open.copernicus.eu).

A few comments:
- I replaced `download_with_authentication`  with `file_downloader` because no credentials are needed anymore;
- I kept the old parameter `cppfile` for `_download_copdem_file_unlocked` although it's not used anymore (didn't want to change too many things);
- I updated the corresponding two tests from `test_utils`  (i.e. `test_copdem90a` and `test_copdem30`). Please check them, I got a bit lost there.

- [ ] Tests added/passed
- [ ] Fully documented

### DIFF
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -43,6 +43,9 @@ Bug fixes
   different CRS. If they do, all files are converted to a single, common CRS
   to make sure they can be merged without issues (:pull:`1698`).
   By `Patrick Schmitt <https://github.com/pat-schmitt>`_
+- COPDEM data is now downloaded from public http server (no credentials
+  required) (:pull:`1684`).
+  By `Codruț Diaconu <https://github.com/dcodrut>`_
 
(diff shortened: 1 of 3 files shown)


fmaussion:
Thank you! Yes I'll check - bare with me though as this is a more consequential change than the code seems to indicate (we will have to check that the files are truly identical and mirror the server locally)

fmaussion:
Thanks @dcodrut 

This is now the "official" way to retrieve COPDEM data. It will imply that I redownload everything on cluster once more...

fmaussion:
Yes I checked one tile and its strictly the same. The filenames have changed though, and the product has a new timestamp (the update only seems to have added a few tiles...)

Issue #1694, opened 2024-03-08
merge_simulated_thickness produces all NaN for all glaciers as soon as one disappears


bpelto:
@pat-schmitt 
Using merge_simulated_thickness (a great tool!!), I found that I didn't get any errors for the first output years, and then eventually got an error for every output year:

Good output for a while: 
Applying global task merge_gridded_data on 24 glaciers
Execute entity tasks [reproject_gridded_data_variable_to_grid] on 24 glaciers

Then only bad:
/home/bpelto/oggm/oggm/sandbox/distribute_2d.py:398: RuntimeWarning: invalid value encountered in scalar divide
  new_thick *= this_vol / tmp_vol

The error stems from when one of the glaciers disappears, whenever that date is, then all values go to NaN as we're trying to divide zero by zero.

It looks like the error occurs here:   
https://github.com/OGGM/oggm/blob/c45194be38c90ef725906172c8ddacf27b0cb2c8/oggm/sandbox/distribute_2d.py#L398C7-L398C40
`new_thick *= this_vol / tmp_vol`

So once I turned preserve_totals to False, my issue was resolved. (e.g. all my glaicers now disapper in 2085 :( instead of 2035 or 2050 when the first glacier disappears amongst the group of 24 I'm running.

It seems like using preserve_totals=True is beneficial to ensure ice volume is conserved. So should I just run with preserve_totals = false, or is that something that needs fixing?

PR #1697, opened 2024-03-12
Bugfix in merge_gridded_data

pat-schmitt:
This is a bugfix for the problem reported here https://github.com/OGGM/oggm/issues/1694. The problem was a division through zero when we try to preserve totals but the glacier has completely disappeared (as described in the issue). I adapted one test and could reproduce and solve the problem. Thanks @bpelto for investigating!

Further, I included here an option for smoothing when merging simulated thickness. This is useful to make the ice thicknesses less pixelated for 3D visualisations. I will add this to the tutorials.

Closes https://github.com/OGGM/oggm/issues/1694

- [ ] Tests added/passed
- [ ] Fully documented


oggm/core/gis.py
```diff
@@ -1958,6 +1963,7 @@ def reproject_gridded_data_variable_to_grid(gdir,
     if preserve_totals:
         # only preserve for float variables
         if not np.issubdtype(data, np.integer):
+            # do we want to do smoothing?
```
VascoSch92, line 1966:
```suggestion
            # TODO: do we want to do smoothing?
```
because is an open question and not a comment ;-) (also it is also a comment)

fmaussion:
Thanks @pat-schmitt ! From the doctext I'm not sure why smoothing can only be done when preserving totals is true? Otherwise looks good to me.

pat-schmitt:
I made this choice because the smoothing effect is quite strong. If we don't preserve totals, the resulting smoothed volume will deviate even further from the actual modeled value. However, I could also adjust the approach so that smoothing is applied without preserving totals.

fmaussion:
I don't mind too much - best is to keep it easy, so let's stick to this for now. Merge when you are ready!

PR #1702, opened 2024-04-02
Add utilities for glacier entities in complexes

fmaussion:


fmaussion:
This cannot be tested easily - will merge for now and think about it later (second PR needed to add this the the prepro_cli)

PR #1709, opened 2024-05-29
added option to use gridded 2D topo data for geodetic_mb_calibration

afisc:
Until now the function `massbalance.mb_calibration_from_geodetic_mb()` did use the height/elevation data of a glaciers flowline(s) to calibrate the massbalance parameters. This is now also possible by using gridded/distributed data via the additional boolean argument `mb_2d`, which is useful for the usage of distributed models like IGM. 

- [x] Tests added/passed
- [x] Fully documented

### DIFF
--- a/oggm/core/massbalance.py
+++ b/oggm/core/massbalance.py
@@ -1433,7 +1433,7 @@ def mb_calibration_from_geodetic_mb(gdir, *,
                                     overwrite_gdir=False,
                                     use_regional_avg=False,
                                     override_missing=None,
-                                    mb_2d=False,
+                                    use_2d_mb=False,
                                     informed_threestep=False,
                                     calibrate_param1='melt_f',
                                     calibrate_param2=None,
@@ -1480,8 +1480,9 @@ def mb_calibration_from_geodetic_mb(gdir, *,
         if the reference geodetic data is not available, use this value instead
         (mostly for testing with exotic datasets, but could be used to open
         the door to using other datasets).
-    mb_2d : bool
-        2D data is used for the mass balance calibration
+    use_2d_mb : bool
+        Set to True if the mass balance calibration has to be done of the 2D mask
+        of the glacier (for fully distributed runs only).
     informed_threestep : bool
         the magic method Fabi found out one day before release.
         Overrides the calibrate_param order below.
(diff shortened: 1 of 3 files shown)


afisc:
added a test case which compares the calibrated 'melt_f' (Melt Factor) for the 1D(flowline) and 2D(gridded_data) calibration method

oggm/core/massbalance.py
```diff
@@ -1433,6 +1433,7 @@ def mb_calibration_from_geodetic_mb(gdir, *,
                                     overwrite_gdir=False,
                                     use_regional_avg=False,
                                     override_missing=None,
+                                    mb_2d=False,
```
fmaussion, line 1436: replace `mb_2d` with `use_2d_mb` (to clarify that this is a yes/no kwarg)

oggm/core/massbalance.py
```diff
@@ -1479,6 +1480,8 @@ def mb_calibration_from_geodetic_mb(gdir, *,
         if the reference geodetic data is not available, use this value instead
         (mostly for testing with exotic datasets, but could be used to open
         the door to using other datasets).
+    mb_2d : bool
+        2D data is used for the mass balance calibration
```
fmaussion, line 1484:
```suggestion
        Set to True if the mass balance calibration has to be done of the 2D mask 
        of the glacier (for fully distributed runs only).
```

oggm/core/massbalance.py
```diff
@@ -1673,6 +1679,8 @@ def mb_calibration_from_scalar_mb(gdir, *,
         if a `mb_calib.json` exists, this task won't overwrite it per default.
         Set this to True to enforce overwriting (i.e. with consequences for the
         future workflow).
+    mb_2d : bool
+        2D data is used for the mass balance calibration
```
fmaussion, line 1683: same as above

oggm/core/massbalance.py
```diff
@@ -1741,7 +1749,18 @@ def mb_calibration_from_scalar_mb(gdir, *,
...
+        fls = gdir.read_pickle('inversion_flowlines')
+    else:
+        # if the 2D data is used, the flowline is not needed.
+        fls = None
+        # get the 2D data
+        fp = gdir.get_filepath('gridded_data')
+        with xr.open_dataset(fp) as ds:
+            # 'topo' instead of 'topo_smoothed'?
+            _heights = np.where(ds.glacier_mask, ds.topo_smoothed, np.nan).flatten()
+            heights = _heights[~np.isnan(_heights)]
```
fmaussion, line 1762:
```      
_heights = np.where(ds.glacier_mask, ds.topo_smoothed, np.nan).flatten()
heights = _heights[~np.isnan(_heights)]
```

Can be replaced with:

```
heights = ds.topo_smoothed.data[ds.glacier_mask.data == 1]
```

fmaussion (commented):
Thanks @afisc ! Great work.

A few comments attached.

Agreed with the problem "smoothed" vs not smoothed. Keep it as is for now (including the inline comment)

PR #1724, opened 2024-08-16
Add Cook23 to the shop

fmaussion:


fmaussion:
This uses a lookup table similar to the millan datasets:

image

PR #1730, opened 2024-08-22
Update flowline.py

dngoldberg:
add the flowline object as a kwarg that can be used by get_annual_mb()

Currently, FluxBasedModel.step() in its call to get_mb() passes fls=self.fls as a kwarg. This change is simply to bring SemiImplicitModel up to parity, so the flowline object can be used by certain types of mass balance classes.

No issue was raised.

### DIFF
--- a/oggm/core/flowline.py
+++ b/oggm/core/flowline.py
@@ -2415,7 +2415,7 @@ class SemiImplicitModel(FlowlineModel):
         b_corr = - d_stag * self.dbed_h_exp_dx
 
         # prepare rhs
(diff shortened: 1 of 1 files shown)


dngoldberg:
@fmaussion i didn't see that!! I had to go home in a rush and didn't know how long the CI would take.

Seems strange.. it's a hydro test which presumably uses SemiImplicit. It may not be benign, at the least the targets should be updated. I can try and look into it if you want to unmerge..?

fmaussion:
@dngoldberg no its unrelated. It started to appear on all our tests and is related to a change in the image upstream. No worries!

PR #1742, opened 2024-09-10
Updating to NumPy > 2.0

pat-schmitt:
In this PR, I address test failures and warnings related to NumPy > 2.0.

Two tests were failing due to a silent conversion from float to int during the [fixed geometry spinup](https://github.com/OGGM/oggm/blob/master/oggm/core/flowline.py#L1274). Here's a small example illustrating the issue:

```python
test_array = np.array([1, 2, 3])
test_nr = np.float64(1.2)
test_array[0] = test_nr
print(test_array)
```
will print `array([1, 2, 3])`, whereas

```python
test_array = np.array([1., 2., 3.])
test_nr = np.float64(1.2)
test_array[0] = test_nr
print(test_array)
```
will print `array([1.2, 2. , 3. ])`. 

This behaviour is tricky to catch and highlights the importance of being mindful of the `dtype` when working with NumPy > 2.0.

Most of the warnings I fixed are related to xarray (likely to be resolved in future updates to xarray), and there are still some warnings from pandas that need to be addressed within pandas itself.

- [ ] Tests added/passed
- [ ] Fully documented

### DIFF
--- a/oggm/core/dynamic_spinup.py
+++ b/oggm/core/dynamic_spinup.py
@@ -1548,8 +1548,8 @@ def dynamic_melt_f_run(
     with utils.DisableLogger():
         ds = utils.compile_run_output(gdir, input_filesuffix=output_filesuffix,
                                       path=False)
-    dmdtda_mdl = ((ds.volume.loc[yr1_ref_mb].values -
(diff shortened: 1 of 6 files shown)


fmaussion:
Thanks!

PR #1744, opened 2024-09-16
Add bin_area to the inversion flowlines

fmaussion:
@bearecinos @dngoldberg this adds the bin_area attribute to the inversion flowlines objects. It's just a handy diagnostic, its also possible to compute it from the previous objects attribues, as is done in the code here.


bearecinos (commented):
yes! that will be what we need ... this means  that we will have to start from a lower pre-processing level right? maybe L1-L2 ? 
and re-compute some prepro levels tasks
Cheers!

bearecinos:
awesome 🥳 thanks @fmaussion

PR #1746, opened 2024-09-26
Adaptations for 2D redistribution

pat-schmitt:
This PR is for discussing some possibilities for adapting the 2D redistribution.

It includes a new possibility for ranking the pixels in each elevation band (order in which pixels melt) and a function for calculating a median flowline (e.g. madian flowline from different gcms with the same scenario).

After some testing we can decide what we want to keep.

- [x] Tests added/passed
- [ ] Fully documented (opened issues for tutorials)

### DIFF
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -9,6 +9,18 @@ v1.6.x (not released)
 Enhancements
 ~~~~~~~~~~~~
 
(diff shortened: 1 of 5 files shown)


pat-schmitt:
When tests pass, this PR is ready for merging.
It was already used for creating the latest visualizations (e.g. see https://edu.oggm.org/en/latest/glacier_videos.html).

PR #1747, opened 2024-09-30
fix floating point precision in floatyear_to_date

pat-schmitt:
In this PR I fix a bug which was brought up by @Ruitangtang.

The problem was when calling `floatyear_to_date(1+1/12-1/12)` you expect to get `(1, 1)`, but actually `(0, 12)` was returned, due to a floating point precision issue. I now included a check if the given floatyear is inside machine precision to the next higher int, and if so us the next higher int instead.

- [ ] Tests added/passed
- [ ] Fully documented

### DIFF
--- a/oggm/tests/test_utils.py
+++ b/oggm/tests/test_utils.py
@@ -123,6 +123,10 @@ class TestFuncs(object):
         r = utils.floatyear_to_date(yr)
         assert r == (1998, 2)
 
+        yr = 1 + 1/12 - 1/12
+        r = utils.floatyear_to_date(yr)
+        assert r == (1, 1)
+
     def test_date_to_floatyear(self):
 
         r = utils.date_to_floatyear(0, 1)
--- a/oggm/tests/test_utils.py
+++ b/oggm/tests/test_utils.py
@@ -123,9 +123,15 @@ class TestFuncs(object):
         r = utils.floatyear_to_date(yr)
         assert r == (1998, 2)
 
+        # tests for floating point precision
         yr = 1 + 1/12 - 1/12
         r = utils.floatyear_to_date(yr)
         assert r == (1, 1)
+
+        for i in range(12):
+            yr = 2000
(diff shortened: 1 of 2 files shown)


fmaussion:
This is a bit of an edge case. Arguably the problem is *not* in OGGM, since OGGM is given a "wrong" number:

```
>>> 1 + 1/12 - 1/12
0.9999999999999999
```

I would normally recommend against doing this kind of arithmetic on "OGGM dates" and use `utils.monthly_timeseries`, but it's okay to round like this (I hope - the only problem I can foresee is when a tiny time step is needed for the dynamical model - but even then, it's tiny)

pat-schmitt:
@Ruitangtang is currently implementing a way to save the fl_diagnostics at monthly steps, and for the diagnostic variable `climatic_mb` we want to use the mb of the previous time-step and hence she is calling `self.get_mb(surface_h_previous[fl_id], self.yr - 1/12, fl_id=fl_id)`, here the current version for yearly timesteps https://github.com/OGGM/oggm/blob/master/oggm/core/flowline.py#L1325.

But probably their is a better way around? Maybe directly saving the previous mb instead of calling it again with -1/12.

And yes, the tiny timestep problem could happen with this method, but it is in the order of 1e-13 of a year (~1e-7 seconds).

Further it seems with numpy 2.0 `len(np.array(1))` or `len(np.int64(1))` is not working anymore. But `len(np.array([1]))` us still good.

fmaussion:
> But probably their is a better way around?

yes I see. I understand why the current behavior is surprising and I agree to the workaround in OGGM. A probable more robust way would be to not to arithmetics on time but rather to do an array of times and use this instead:

```python
monthly_time = utils.monthly_timeseries(y0, y1)
i = 12
print(monthly_time[i])
print(monthly_time[i-1])  # previous months
```

fmaussion (approved):

PR #1756, opened 2025-01-13
Add new possiblities for providing paramters.

pat-schmitt:
This PR introduces two new ways to define and provide parameters for the model (in addition to the current default of using `cfg.PARAMS`):

1. Minimal `params.cfg` File: You can now pass a minimal `params.cfg` file to `cfg.initialize`. If some parameters are missing in this file, the defaults from the main `params.cfg` (available [here](https://github.com/OGGM/oggm/blob/master/oggm/params.cfg)) will be used automatically.
2. New `run_settings.yml` File, which could contain:
  -  Almost all parameters from `params.cfg`
  -  Values of observations used during calibration or initialization
  -  If any parameter is not included in `run_settings.yml`, it falls back to the current defaults.

These additions required significant changes in sensitive parts of the model, particularly for the second point. However, tests appear to be passing. I have also included some tests for `run_settings.yml`. Not every aspect is tested yet, as adding more tests would overly expand the test suite.

Points for Discussion
- Each function currently opens the `run_settings.yml` file independently, instead of opening it once at the start of a workflow and passing it around. While this approach may slightly reduce performance (dynamic runs are not effected by this), it maintains the flexibility needed for setting up custom workflows.
- The new `run_settings.yml` works similarly to the existing `mb_calib` logic but supports calibration for only a single flowline. For now, I’ve kept `mb_calib` for backward compatibility and enforce a selection between the two possibilities to avoid potential confusions. However, we could consider removing `mb_calib` in the future.
- When using `run_settings.yml`, all diagnostics are saved both to `gdir.add_diagnostics` and the `run_settings.yml` file. Maybe one or the other is enough?

- [ ] Tests added/passed
- [ ] Fully documented

### DIFF
--- a/oggm/cfg.py
+++ b/oggm/cfg.py
@@ -301,6 +301,11 @@ BASENAMES['elevation_band_flowline'] = ('elevation_band_flowline.csv', _doc)
 _doc = "The outlines of this glacier complex sub-entities (for RGI7C only!)."
 BASENAMES['complex_sub_entities'] = ('complex_sub_entities.shp', _doc)
 
+_doc = ("A dict containing optional settings for a model run. Could contain "
+        "mass balance parameters, dynamic parameters or observations. If a "
+        "required parameter is not included, the default value will be used.")
+BASENAMES['run_settings'] = ('run_settings.yml', _doc)
+
 
 def set_logging_config(logging_level='INFO'):
     """Set the global logger parameters.
--- a/oggm/cfg.py
+++ b/oggm/cfg.py
@@ -378,11 +378,14 @@ def initialize_minimal(file=None, logging_level='INFO', params=None):
     Parameters
     ----------
     file : str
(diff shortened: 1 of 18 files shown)


pat-schmitt:
I managed to fix all tests, beside of `python:3.11-slim`. The problem is that the package `yaml` is not getting installed. @TimoRoth  or @fmaussion could you direct me, where I need to add the package to be installed? Alternatively, I can do all imports of `yaml` using `try`-statements.

Beside of this, the PR slowly gets into a merge-able state. However, as this touches almost all sensitive parts of OGGM, before merging we should have a meeting. I would like to discus if the way I implemented everything is a good way or if their are better ways of doing it. And we need to decide how all this is or could change the default way of handling the params in OGGM.

pat-schmitt:
Yes, I will prepare a notebook for discussion.

Issue #1755, opened 2025-01-01
Mass balance model with "memory", and issues for apparent_mb_from_any_mb (and others)


dngoldberg:
I am helping to develop the coupling of oggm with an energy balance/snow model, and this has been implemented by creating a new class that inherits from the MassBalanceModel class. 

The snow model has memory, such that in consecutive years the snow model and ice temperature state is preserved. An issue arises when get_specific_mb is meant to be called multiple times with a specific instance. As far as i can tell, this is resulting in `apparent_mb_from_any_mb` throwing an error in `_check_terminus_mass_flux`, where the exact residual has been found.. but then is no longer the correct residual since a subsequent call at [this line](https://github.com/OGGM/oggm/blob/5ac2eea127bb21d92d5c86c3851b73fc181d5507/oggm/core/massbalance.py#L2153). As I have not explored enough functionality, Im not sure but similar issues may arise elsewhere.

My question is -- what would be the best way to handle this. I could envision:

1. adding a kwarg, e.g. `reset_state=false` to `massbalancemodel.get_specific_mb` (and `get_annual_mb`) to "reset" the state, and using the kwarg where appropriate (there will likely be multiple locations). It would do nothing for all exisiting MB classes.
2. re-instantiating the mass balance model where a state reset is needed.

(1) seems better, though will look a little odd until (hopefully) our energy mb model is merged. if you are OK with this, i will create a branch and add the kwarg in all the places i think are appropriate, though this will likely not be exhaustive without further testing.

fmaussion:
Thanks for raising this issue. I don't see a problem in adding functionality for downstream projects, but a few things come to mind first:
- Having a memory for a MB Model should be no issue. I'm not sure to understand why a subsequent call to the same year should lead to different results, since ideally your MB model will have stored (cached) the result of previous computations a a new call to a given year should simply return the same value.
- It's a bit lazy of the apparent_mb_from_any_mb task to call the MB model twice here. Actually, we could change the code to call it only once ([this call](https://github.com/OGGM/oggm/blob/5ac2eea127bb21d92d5c86c3851b73fc181d5507/oggm/core/massbalance.py#L2153) is needed, but could be done first and then the specific MB could be computed from it. 

Regardless of the above, ideally a call to `get_mb` for a given year and geometry should always give the same result. I know that models with memory raise challenges that are not well handled by OGGM at the moment, and eventually some form of "reset" logic might be needed, but the reason here seems not yet justified to me? To discuss.

PR #1757, opened 2025-01-14
Add capability to reset the "state" of the mass balance model

dngoldberg:
Although this does not apply to any existing mass balance models, there may be mass balance models in the future that have a "state" associated. Now, for a given set of climate files for a given glacier in a given year or interval, the mass balance models return consistent values whenever `get_annual_mb` or `get_specific_mb` is called. But if an energy balance model is implemented, the mass balance will depend on properties such as snow state and glacier surface temperature, and this may not give consistent results unless there is a consistent "state". The issue arises, for instance, in `apparent_mb_from_any_mb`, **but there could be others as well**.

One possibility is to save the "state" in a given year in a glacier directory. However, this will not completely address the problem as  get_annual_mb may be called for a different year. 

Another is to give flexibility by allowing a "reset_state" kwarg. This does absolutely nothing with existing mass balance models, but allows developers of new MB models a way to address the issues above. That is what the current pull request does.

**NOTE** a far less-invasive option to address the issue in `apparent_mb_from_any_mb` might be to re-instantiate the mass balance model within the function. But this led to multiple testing errors.

Closes https://github.com/OGGM/oggm/issues/1755

- Passes All ci Tests
- [ ] Fully documented
- Entry in `whats-new.rst`: Added "reset_state" functionality to ``MassBalanceModel.get_specific_mb()`` to 
  signal any state-dependent mass balance model to "reset its state" at the start
  of the period. This is achieved through a "reset_state" kwarg passed by 
  ``MassBalanceModel.get_annual_mb()``. Done to address that 
  ``apparent_mb_from_any_mb()`` calls ``get_specific_mb()``  twice, leading to nonzero
  terminal flux (potentially) with a state-dependent mass balance model.
  `apparent_mb_from_any_mb()` also changed accordingly. Similar change also 
  made to ``MultipleFlowlineMassBalanceModel.get_specific_mb()``. Does not impact
  any current functionality within oggm.

**NOTE**: this PR is invasive, and i have decided on a less invasive change. so i am closing it.

### DIFF
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -21,15 +21,13 @@ Enhancements
   ``dis_from_border``, ...) to define this ranking, providing greater flexibility
   and control over how the melting sequence is visualized (:pull:`1746`).
   By `Patrick Schmitt <https://github.com/pat-schmitt>`_
-- Added "reset_state" functionality to ``MassBalanceModel.get_specific_mb()`` to 
-  signal any state-dependent mass balance model to "reset its state" at the start
-  of the period. This is achieved through a "reset_state" kwarg passed by 
-  ``MassBalanceModel.get_annual_mb()``. Done to address that 
-  ``apparent_mb_from_any_mb()`` calls ``get_specific_mb()``  twice, leading to nonzero
-  terminal flux (potentially) with a state-dependent mass balance model.
-  `apparent_mb_from_any_mb()` also changed accordingly. Similar change also 
-  made to ``MultipleFlowlineMassBalanceModel.get_specific_mb()``. Does not impact
-  any current functionality within oggm.
+- Added "reset_state()" function to ``MassBalanceModel`` to signal any state-dependent 
+  mass balance model to "reset its state" at the start of the period. This function
+  does nothing in the parent class and would only be implemented by a state-ful
+  mass balance model. Done mainly to address that ``apparent_mb_from_any_mb()`` calls 
+  ``get_specific_mb()``  twice, leading to nonzero terminal flux (potentially) with a 
+  state-dependent mass balance model. `apparent_mb_from_any_mb()` also changed 
+  accordingly. 
 
 Bug fixes
 ~~~~~~~~~
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -21,14 +21,14 @@ Enhancements
   ``dis_from_border``, ...) to define this ranking, providing greater flexibility
   and control over how the melting sequence is visualized (:pull:`1746`).
   By `Patrick Schmitt <https://github.com/pat-schmitt>`_
-- Added "reset_state" functionality to `MassBalanceModel.get_specific_mb()` to 
+- Added "reset_state" functionality to ``MassBalanceModel.get_specific_mb()`` to 
   signal any state-dependent mass balance model to "reset its state" at the start
   of the period. This is achieved through a "reset_state" kwarg passed by 
-  `MassBalanceModel.get_annual_mb()`. Done to address that 
-  `apparent_mb_from_any_mb()` calls `get_specific_mb()`  twice, leading to nonzero
+  ``MassBalanceModel.get_annual_mb()``. Done to address that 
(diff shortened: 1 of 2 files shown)


fmaussion:
@dngoldberg thanks for this. Here is an alternative suggestion which I think is a bit cleaner. Changes:
- reset_state() is now a method of any MB model. It does nothing per default.
- reset_state() is called each year if asked for. I still think it's a bit wild to do that, but I can see how this can be useful, if anything for testing at least

The current way that `apparent_mb_from_any_mb` computes the mass-balance twice is still not optimal. We really don't need to do that, and I'll look into this in a separate PR (should not be too much work).

Let me know if this works for you as suggested by my last commit.

dngoldberg:
hi @fmaussion thanks very much and i will look at your change soon. But i want to say, i was thinking about it further and maybe the issue that led to this PR could be addressed simply at the level of the "state"-ful MB model through implementing get_specific_mb() instead of using the parent class.

at the same time, though, you mentioned that a MB model should always be reproducible and the output of get_annual_mb() should not depend on previous calls. this PR is meant to address this which is why i decided to submit it after all. But if you think this is less important, i could close without merging. Please let me know.

i need to understand apparent_mb_from_any_mb() better, my suspicion is that it does NOT need to call get_specific_mb() but i will look carefully. (that would, also, address the issue that precipitated this PR -- but, of course, there might be other issues not yet encountered, such as in dynamic spinup..)

fmaussion:
@dngoldberg the current PR is totally fine by me. It should do exactly what you wanted it to do (resetting a state each year when calling multiple years), but a bit differently.

>my suspicion is that it does NOT need to call get_specific_mb()

I agree with that ;-) - I've just been lazy. The second call to MB on the flowlines is what's needed. It can be avoided.

dngoldberg:
hi @fmaussion have looked it over. while this does solve the apparent_mb_from_any_mb() issue, it is not quite what i wanted, because in the specific implementation that I am thinking of the state *is* important when modelling mb over a multi-year period. this is why i was only resetting the state at the *beginning* of get_specific_mb().

However -- i like your addition of a reset_state() function in the parent class. Therefore I propose this:

1. we keep the reset_state() that does nothing in the parent class
2. undo any changes to get_specific_mb()
3. **Just for now**, we add a reset_state() call in the middle of apparent_mb_from_any_mb(). this will make the two calculations of mean mass balance over the apparent_mb period equivalent, which was the main problem i was concerned with
4. create an issue to refactor apparent_mb_from_any_mb()
5. the reset_state() method can then be used for any state-dependent mb model whenever it is needed.

I don't know a clever way to (1) and (2) without just undoing the commits and adding the reset_state() function manually. but if you agree with the above, and you know a quick way, feel free!

dngoldberg:
hi @fmaussion what do you think of these changes? i like it because it's far less extensive. all tests pass, and it addresses the issue i was having with apparent mass balance. refactoring apparent_mb_from_any_mb() might be beyond the scope of this PR, but an issue could be created. meanwhile, this gives more flexibility.

fmaussion:
@dngoldberg sorry I'm still thinking about this - I think that the reset_state is a good thing and should be added, but I'll work this week on making the inversion *not* reliant on this. I don't think its good to add a kwarg that will be made redundant soon

dngoldberg:
thanks fabi. i haven't tried to make apparent_mb_from_any_mb so that it only effectively goes through the time range finding mass balance just once. but if it did so, i *think* this would be ok -- so long as mass balance for each year is calculated in temporal order. let me know!

fmaussion:
Apologies for the time it took me to get there @dngoldberg - this pr now introduces the reset_state() method, but it's used nowhere in OGGM core. It might be useful in the future though.

apparent_mb_from_any_mb *finally* does compute MB only once. This is probably much more useful.

I note that some compute intensive models like yours might even want to provide MB profiles to this function instead of letting it compute it. This could easily be added.

dngoldberg:
thanks @fmaussion -- we will give this suggestion some thought, but just see how things go performance wise for now..

Issue #1750, opened 2024-10-30
Daily mass balance in OGGM


fmaussion:
This is a key requirement of the digital twin project. This is a placeholder to keep track of progress on this front.

gampnico:
Links to #1574, #1703 (implementation strategy).

Issue #855, opened 2019-08-16
Speed-up the numerical core


fmaussion:
I think there is potential for improvement on some aspects of the numerical core:
- review all occurrences of np.clip (https://github.com/numpy/numpy/issues/14281)
- do the sliding computation only if fs != 0
- other array shenanigans which require some thinking

PR #1767, opened 2025-03-31
Refactor recursive functions

gampnico:
This PR refactors recursive methods into iterative ones.

The refactored methods have known bounds (either a single year, or a predetermined date range), and therefore do not benefit from recursion. My main motivation here is to simplify the daily MB implementation (#1750). The iterative methods are much easier to overload and test.

**Changes**
* Recursive functions are now iterative.
* Removes ``np.append`` in favour of concatenating over lists of numpy arrays, which provides ~2x performance increase. This is especially important for #1750 where arrays are larger and ragged.
* Adds ``utils.set_array_type`` to ensure single values are returned as floats, because numpy 1.25+ no longer supports converting arrays to scalars. This maintains backwards compatibility with other parts of OGGM that expect a float.

**Points for discussion**
* I have run my own tests to ensure the recursive and iterative variants produce an identical output. However, I would like to know if (and how) these tests should be included, as the recursive variants have been removed entirely. This PR only checks if a value is within bounds for the minimum and maximum flowlines.
* ``np.append`` is used liberally in OGGM - presumably to deal with ragged arrays/setting elements with a sequence? It's usually faster to use ``np.concatenate`` or ``np.stack`` over lists of numpy arrays as the axis is always known (this is what ``append`` does internally anyway), and the elements in the lists still benefit from numpy's memory efficiency. Perhaps this could be cleaned up in a round of optimisations (#855)?

Refs: #1750,  #855

- [x] Tests added/passed
- [x] Fully documented

### DIFF
--- a/oggm/core/massbalance.py
+++ b/oggm/core/massbalance.py
@@ -17,7 +17,7 @@ from oggm.utils import (SuperclassMeta, get_geodetic_mb_dataframe,
                         floatyear_to_date, date_to_floatyear, get_demo_file,
                         monthly_timeseries, ncDataset, get_temp_bias_dataframe,
                         clip_min, clip_max, clip_array, clip_scalar,
-                        weighted_average_1d, lazy_property)
+                        weighted_average_1d, lazy_property, set_array_type)
 from oggm.exceptions import (InvalidWorkflowError, InvalidParamsError,
                              MassBalanceCalibrationError)
 from oggm import entity_task
@@ -194,8 +194,7 @@ class MassBalanceModel(object, metaclass=SuperclassMeta):
                 mbs = weighted_average_1d(mbs, widths)
             out.append(mbs)
 
-        # Convert units here to maintain backwards compatibility
-        return np.asarray(out) * SEC_IN_YEAR * self.rho
+        return set_array_type(out) * SEC_IN_YEAR * self.rho
 
     def get_ela(self, year=None, **kwargs):
         """Compute the equilibrium line altitude for a given year.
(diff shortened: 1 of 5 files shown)


oggm/core/massbalance.py
```diff
@@ -125,88 +125,113 @@ def get_annual_mb(self, heights, year=None, fl_id=None, fls=None):
         """
         raise NotImplementedError()
 
-    def get_specific_mb(self, heights=None, widths=None, fls=None,
-                        year=None):
-        """Specific mb for this year and a specific glacier geometry.
-
-         Units: [mm w.e. yr-1], or millimeter water equivalent per year
+    def get_weighted_mb_from_flowlines(self, fls: list, year: float) -> tuple:
```
fmaussion, line 128: Return type should be float

fmaussion (commented):
Looks reasonable to me! Thanks!

The only doubt I have is about the name and visibility of the new methods. What they do is compute the specific MB, i.e. they should be called that way (otherwise its confusing). A few options:
- get_specific_mb_from_flowlines
- get_specific_mb_from_fls
- make it private
- keep the code within the old method (bit ugly)

Thoughts?

> I have run my own tests to ensure the recursive and iterative variants produce an identical output. However, I would like to know if (and how) these tests should be included

This is not necessary  - the current test suite should catch any problem for such an important method (famous last words...) 

> np.append is used liberally in OGGM / could be removed

Yes - to be tackled in a separate PR, or perhaps within this one if you see MB related optims that can be done quickly now.

Same holds true with type hints - feel free to add it to the lines you are touching like in this PR, but only if you have time / energy.

fmaussion:
Ah it could also be called "get_annual_specific_mb" and keep the same signature as get_specific_mb (i.e. also accept heights). I think that get_specific_mb accepting scalars and arrays was an early coder's mistake anyways

gampnico:
> Ah it could also be called "get_annual_specific_mb" and keep the same signature as get_specific_mb (i.e. also accept heights). I think that get_specific_mb accepting scalars and arrays was an early coder's mistake anyways

Thanks, I'll think about the naming conventions, as with #1750 I don't want to end up with some byzantine combination of get_specific_mb, get_specific_mb_daily, get_annual_specific_mb_daily, get_daily_specific_mb_daily etc.

PR #1776, opened 2025-05-06
Allowe to use a mini params file during initialization

pat-schmitt:
Previously, modifying parameters during initialization required supplying a complete `params.cfg` file. This PR introduces support for an optional `mini_params.cfg` file, usabel as `cfg.initialize(file=mini_params_filepath)`. This file can include only the parameters you wish to override, defaults will be filled in from the base `params.cfg`.

his PR extracts a subset of changes from PR https://github.com/OGGM/oggm/pull/1756, as they are unrelated to the new parameter handling approach envisioned in that PR.

- [x] Tests added/passed
- [ ] Fully documented


fmaussion:
You can merge when green!

PR #1777, opened 2025-05-12
add new settings to massalance and flowline

pat-schmitt:
This is a first package of implementing the new way of parameter handling in OGGM. It is an adapted version of parts from PR https://github.com/OGGM/oggm/pull/1756.

 Currently those new settings handling is only included in massbalance.py and flowline.py (including their dependencies from other parts of the model). Further work is needed in the future to make everything consistent again.

- [x] Tests added/passed
- [ ] Fully documented

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -4,9 +4,11 @@ on:
   push:
     branches:
       - master
+      - dev
   pull_request:
     branches:
       - master
+      - dev
 
 jobs:
   test:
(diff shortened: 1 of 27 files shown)


pat-schmitt:
Ok, this is ready for review.

I included now some parts which are not ideal. The main troublemaker is the implementation of the dynamic spinup (which I plan to refactor a lot during summer). Consequences are that I needed to include a read from the settings yml file, whenever a parameter is obtained. This ensures that the provided parameter is always in sync with the file, even if their are two objects open, which are relying on the same yml file.

Further, for ensuring backwards compatibility, I used `cfg.PARAMS.copy()` at some points. I used the copy() only because otherwise the log will be flooded with messages when a object is copied (you see this if you do `copy.deepcopy(cfg.PARAMS)`). This is again something which is used in the dynamic spinup (really needs some refactoring).

However, so far the mass balance and the dynamics are working with the new settings file and I think this are the most critical parts for implementing the new daily massbalance model.

PR #1778, opened 2025-05-13
Compile fl diagnostics

fmaussion:
Closes a request in slack

### DIFF
--- a/oggm/tests/test_models.py
+++ b/oggm/tests/test_models.py
@@ -2713,6 +2713,7 @@ class TestHEFNonPolluted:
             assert 'volume_bsl_m3' not in ds_fl
 
         past_run_file = os.path.join(cfg.PATHS['working_dir'], 'compiled.nc')
+        diag_file = os.path.join(cfg.PATHS['working_dir'], 'compiled_diags.nc')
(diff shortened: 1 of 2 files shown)


fmaussion:
Users were asking to get diagnostics along the glacier flowlines for use in workflows external to OGGM. I think this is a fair request and an "asset" we should add to this page: https://docs.oggm.org/en/stable/assets.html 

This code is almost ready but it may need clarification from the users about what they want. Currently, the code simply opens the glacier directories, reads the `fl_diagnostics` file, and writes it in another file.

Here is a repr of what's in a diagnostics file for one glacier:

image

This is only one flowline (i.e. if we want to extract multiple flowlines the data structure would be even more nested). 

Each glacier has a different length, so we can only write each file either individually or as part of a nested (grouped) netcdf, either all glaciers in a region or in groups of 100. I note that if we write it individually the only difference to the user is that the data is extracted from a larger tar file, but the user still has to download each file manually.

Then there is the question of what to store:
- `_historical` is a run from the OGGM inverstion at the RGI date up to 2020
- `_spinup` is a run including dynamical spinup (i.e. better from an OGGM perspective, but less close to the original inversion)
- do the years outside of the RGI outline matter?

etc. Discussion to be continued

Issue #1771, opened 2025-04-04
Bug: custom calibration files are ignored


gampnico:
**Issue**

Some functions appear to silently use the default ``mb_calib.json`` even if the user has specified their own file suffix. Examples include merging tributary glaciers or initialising a merged glacier, dynamic spinup, dynamic melt calibration, glacier statistics etc.

**Possible fixes**

In most cases this can be solved either by passing the file suffix through cfg.PARAM, or adding args/kwargs/attrs, but it would be good to know how users approach calibration, especially in sensitivity tests - are they generating lots of custom calibration files in one go, or just defining one custom path for an entire workflow?

Refs: #1677, #1733, #1750

pat-schmitt:
Hi, yes this is true. I have already covered this issue in this PR https://github.com/OGGM/oggm/pull/1756. We can discuss once I have finished the 'discussion notebook'.

Issue #1574, opened 2023-05-03
Move (parts) of MB Sandbox to OGGM proper


fmaussion:
Current status:
- https://github.com/OGGM/massbalance-sandbox works well but is not fully compatible with 1.6.1
- there is a definite need from the community to use these models
- it seems inefficient to make the sandbox compatible with 1.6.1 if we move parts of it to OGGM anyway
- we are waiting for the reviews of the paper (will they change the models?) 
- @lilianschuster is busy with other things.

As often, resources is the issue. Implementing stuff in OGGM is hard (see Jan's open PR). Leaving this open for now.

Issue #1349, opened 2021-12-08
Hydro year definition for mass-balance calibration to compute pre-processed directories 


nchampollion:
The parameters used here **(https://cluster.klima.uni-bremen.de/~oggm/gdirs/oggm_v1.4/L3-L5_files/CRU/elev_bands/qc3/pcp2.5/match_geod_pergla/params.cfg)** to compute the glacier directories (level 3, elevation bands, match geodetic mass balance per glacier) are the following:

```
   # Hydrological year definition
   hydro_month_nh = 1
   hydro_month_sh = 1
```

What is the reason not using the usual ones ?

```
   # Hydrological year definition
   hydro_month_nh = 10
   hydro_month_sh = 4
```

Thanks !

fmaussion:
> What is the reason not using the usual ones ?

The main reason is that the geodetic reference data is given in calendar years - so we match that as well. 

The other reason is that it's much easier, we can finally get rid of hydro years ;-)

nchampollion:
> The main reason is that the geodetic reference data is given in calendar years - so we match that as well.
That was what I was thinking. But in that case, why not choosing calendar years for calibration at regional level ? Because the geodetic reference data at regional level are also in calendar years ?

> The other reason is that it's much easier, we can finally get rid of hydro years ;-)
I prefer this reason ;-)

Issue #950, opened 2020-02-10
Revise the MB Model get_annual_mb API


fmaussion:
After some back and forth with @drounce , I think that the current API (passing `fl_id` and `fls` could be replaced by a cleaner `flowline`. I have to think about it more.

PR #1797, opened 2025-08-19
Daily TI model and W5E5 daily data

gampnico:
This PR adds support for data at daily resolution and a daily temperature index model (#1750).

**Changes**

Adds:
- Daily temperature model which uses W5E5 data at daily resolution.
- Shop function to download and process W5E5 data at daily resolution via ``process_w5e5_data`` or ``process_gswp3_w5e5_data_daily``. This creates ``climate_historical_daily``.
- Support for leap years.
- Hydro/calendar date conversions with cftime - this is faster and handles arrays.

Refactors:
- ``MassBalanceModel``, ``MonthlyTIModel``. My strategy has been to refactor their methods into smaller ones, and overload with DailyTI's methods. This may need to change when implementing surface tracking, but it'll reduce merge conflicts downstream.
- Construction of valid URLs in ``conftest.patch_data_urls``.
- Minor optimisations (duplicate code, performance).
- Tests (#791)

**Points for discussion**

There is a negligible discrepancy O(-15) between the annual smb calculated from ``get_specific_mb`` and ``get_specific_mb_daily``. I think this is just because of when means/sums are taken, which cause different floating point errors.

I've added some type aliases to the autodoc configuration in ``docs/conf.py``. This imports the module under the hood, so it may be faster/safer to build docs without this.

Refs: #844, #950, #1349, #1574
Closes: #1750, #1771

- [ ] Tests added/passed - **one test marked as xfail**
- [ ] Fully documented

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -380,16 +380,11 @@ Interface
     :nosignatures:
 
     MassBalanceModel
+    MassBalanceModel.get_daily_mb
     MassBalanceModel.get_monthly_mb
     MassBalanceModel.get_annual_mb
     MassBalanceModel.get_specific_mb
     MassBalanceModel.get_ela
-    MonthlyTIModel
-    MonthlyTIModel.get_specific_mb
-    DailyTIModel
-    DailyTIModel.get_specific_mb_daily
-    DailySfcTIModel
-    TIModel_Sfc_Type
 
 Models
 ------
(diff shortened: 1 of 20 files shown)


pat-schmitt:
I close this, as the changes here were added to dev in https://github.com/OGGM/oggm/pull/1800 and https://github.com/OGGM/oggm/pull/1801

Issue #1787, opened 2025-07-17
More flexibility when defining ref_period in mb_calibration_from_scalar_mb


pat-schmitt:
Currently, `ref_period` must be specified in full years (from January 1st to January 1st). For more flexibility, it should be possible to specify arbitrary date ranges.

PR #1819, opened 2025-11-20
Adding more flexibility in defining period of reference massbalance

pat-schmitt:
This PR introduces more flexibility in defining `ref_mb_period` within `mb_calibration_from_scalar_mb `. The update allows calibration for hydrological years or between arbitrary dates.

A key change is that I added a new kwarg `ref_mb_unit` to `mb_calibration_from_scalar_mb` . This parameter allows to set the unit of the provided reference mass balance:

- kg m-2 yr-1 (default, used sofar)
- kg m-2

The new` ref_mb_unit` parameter ensures that when the period (`ref_mb_period`) does not correspond to full calendar years, the reference mass balance should be provided in kg m-2.

This approach simplifies usage for users, as it avoids the need to account for leap years when providing reference mass balance for arbitrary periods. While the effect of leap years may not be significant, the `DailyTIModel` internally accounts for them, so it is consistent to ensure observations also consider this.

Happy to discuss this further!

Closes https://github.com/OGGM/oggm/issues/1787

- [ ] Tests added/passed
- [ ] Fully documented

### DIFF
--- a/oggm/core/massbalance.py
+++ b/oggm/core/massbalance.py
@@ -3483,8 +3483,8 @@ def mb_calibration_from_scalar_mb(gdir, *,
                                   ):
     """Determine the mass balance parameters from a scalar mass-balance value.
 
-    This calibrates the mass balance parameters using a reference average
-    MB data over a given period (e.g. average in-situ SMB or geodetic MB).
+    This calibrates the mass balance parameters using reference MB data over a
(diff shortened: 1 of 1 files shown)


fmaussion:
if using `kg m-2`, is it the sum over the entire period then?

Issue #1817, opened 2025-11-12
Xarray merge warning in process_isimip_data


lilianschuster:
I am getting a xarray warning when running `process_isimip_data` that xarray will change its `xr.merge` behaviour. When running future simulations it results in a lot of warning which is cluttering the slurm files.  Probably we should add `join='outer'` to preserve the old behaviour on these two lines https://github.com/OGGM/oggm/blob/dfca928fa4339e87cb3f8293cc0e3e73b811a8de/oggm/shop/gcm_climate.py#L287 and https://github.com/OGGM/oggm/blob/dfca928fa4339e87cb3f8293cc0e3e73b811a8de/oggm/shop/gcm_climate.py#L306 ? 

For more info: Here is the warning that I am getting ... 
`FutureWarning: In a future version of xarray the default value for join will change from join='outer' to join='exact'. This change will result in the following ValueError: cannot be aligned with join='exact' because index/labels/sizes are not equal along these coordinates (dimensions): 'time' ('time',) The recommendation is to set join explicitly for this case.
  temp_a = xr.merge([temp_a_gcm, temp_a_hist], ...`

--> and here are the xarray changes described: https://docs.xarray.dev/en/v2025.08.0/user-guide/combining.html#merge

fmaussion:
Do you want to make the change @lilianschuster ?

PR #1821, opened 2025-11-25
Remove Xarray merge warning and add secondary ISIMIP3b GCMs

lilianschuster:
This pull request
- removes warning when doing `xr.merge` within `process_monthly_isimip_data` (due to  new xarray versions) --> resolves issue #1817 

- allows usage of other ISIMIP3b SSPs and GCMs


lilianschuster:
The new ISIMIP3b climate test files are missing, I will update that.

PR #1835, opened 2026-01-14
updated era5 to include 2025

lilianschuster:
Downloaded and postprocessed ERA5 monthly and daily to also include 2025 completely. Here are the updated paths. However, I am not sure if this should only be included for OGGM v1.7  as the ERA5 1.6.3 gdirs used the ERA5 files until end of 2024.


fmaussion:
Thanks Lily! I think it wouldnt be too much work to update the gdirs - can I do that?

Issue #1774, opened 2025-04-25
Enhancement: rework dependencies


gampnico:
General users are expected to install OGGM's dependencies using a yaml file listed in the docs. Only then can users install OGGM.

Contributors are expected to update those requirements themselves (e.g. unfreezing package versions), which has created a discrepancy between outdated documentation, undocumented internal standards shared between "core contributors", newer contributors, and environments used in the CI pipeline.

An example of this is the ``numpy`` dependency for the main branch: in [docs](https://docs.oggm.org/en/latest/installing-oggm.html#the-simplest-way-with-an-environment-file) and the [dependency lists](https://github.com/OGGM/OGGM-dependency-list/blob/master/Linux-64/oggmdev-latest_py311.yml) it is frozen at <2.0, but internally and within the [CI pipeline](https://github.com/OGGM/oggm/actions/runs/14200427836/job/39785793295) we are already using 2.x.

Additionally, maintainers for packages that depend on OGGM are either expected to track and list OGGM's dependencies within their own package's requirements (eg. [DTCG](https://github.com/DTC-Glaciers/dtcg/issues/9#issue-3013588840)), or link their users to follow OGGM's installation tutorial.

This presents a two- or more step installation process which may discourage contributions from those less familiar with OGGM, and shifts responsibility for dependency resolution onto users.

A possible solution is to explicitly declare OGGM's current dependencies in ``setup.cfg``.

**Points for Discussion**

@TimoRoth, would these changes have unwanted side effects on the CI pipeline, especially for Docker containerisation?

@fmaussion also noted there are legacy reasons for using yaml files related to dependency resolution in earlier versions of package management systems. Is this still an issue with modern package managers, and with newer versions of Python?

TimoRoth:
I don't really understand what changes you're asking for.
The repo you linked is a set of pre-tested conda environment, which are automatically updated.
Github just keeps turning off the workflow maintaining them in shorter and shorter intervals, but that's an unrelated issue.
They purely exist to deal with the incredibly flakey and unreliable conda/conda-forge packaging situation.

And non-optional dependencies are already listed in setup.cfg:
https://github.com/OGGM/oggm/blob/master/setup.cfg#L32

gampnico:
Thanks @TimoRoth,

Not all of the required dependencies are in ``setup.cfg`` (e.g. ``salem``, and even the listed python requirement is outdated) - my issue is there is no consistent list of dependencies between docs, users, and testing environments.

[...]

PR #1841, opened 2026-02-12
Fix distutils str2bool deprecation

gampnico:
Reimplements ``distutils.strtobool`` natively, as ``distutils`` is no longer part of the standard library from Python 3.12+.

I have added the test for this into ``test_utils``, as I couldn't find a dedicated ``test_cfg`` module.

Refs: https://github.com/OGGM/oggm/issues/1774

- [x] Tests added/passed
- [x] Fully documented


fmaussion:
@gampnico can we prioritize this (and the switch to uv) this week?

PR #1843, opened 2026-02-18
corrected longitude selection near 0

lilianschuster:
Removes bug in climate selection for glaciers near 0 longitude:

- this bug only appeared for ERA5, where all glaciers with longitudes (-0.125, 0.125) should use the longitude 0 data of ERA5. However, due to the +360 transformation, the neighboring gridpoint 359.75 was used for two small glaciers in previous versions --> see more in #1825 
- still quite some code changes necessary, as the glacier longitude transformation ("lon"), depends on the resolution and minimum longitude of each climate dataset 
- no direct "testing" possible, as there is only a hef_gdir as test available, but indirect test inside of `test_glacier_gridpoint_selection`. Also tested, that these changes do not influence W5E5 processing. 

--> I am not sure if the Histalp, CRU selection procedures have the same issues (I am not sure if I understand the salem code correctly)

- [x] Tests added/passed

### DIFF
--- a/oggm/shop/ecmwf.py
+++ b/oggm/shop/ecmwf.py
@@ -101,7 +101,7 @@ def _check_ds_validity(ds):
                                              'month': ds['time.month'],
                                              'day': 1})
     info = utils.climate_file_info(ds)
-    assert info['lon_range'] >= 0
+    assert info['lon_range'][0] >= 0
     return info
 
 


lilianschuster:
Do we also need to correct the longitudes for the gcm selection? I think yes, eventually ( 
- for process_isimip_data it does not matter, as it has the same resolution as W5E5, and no "0"-longitude cross-overs: https://github.com/OGGM/oggm/blob/0dd524ec650a937f1c4bb8122db0d8b17ed126f7/oggm/shop/gcm_climate.py#L274
- but probably important for the cesm, or process_cmip_data?: https://github.com/OGGM/oggm/blob/0dd524ec650a937f1c4bb8122db0d8b17ed126f7/oggm/shop/gcm_climate.py#L492

PR #1848, opened 2026-02-23
Bump Test-Images

TimoRoth:


fmaussion:
I'll try to see if I can reproduce locally today, otherwise yes we will remove 3.14

Issue #1482, opened 2022-09-20
Missing pytables dependence after pip install


juseg:
Is pytables a required dependence missing from `setup.cfg`? After install with pip:

```python
$ python -c "import oggm; oggm.cfg.initialize()"
2022-09-20 15:43:49: oggm.cfg: Reading default parameters from the OGGM `params.cfg` configuration file.
2022-09-20 15:43:49: oggm.cfg: Multiprocessing switched OFF according to the parameter file.
2022-09-20 15:43:49: oggm.cfg: Multiprocessing: using all available processors (N=4)
Traceback (most recent call last):
  File "/usr/lib/python3.10/site-packages/pandas/compat/_optional.py", line 138, in import_optional_dependency
    module = importlib.import_module(name)
  File "/usr/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1004, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'tables'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/julien/.local/lib/python3.10/site-packages/oggm/cfg.py", line 664, in initialize
    get_dl_verify_data('dummy_section')
  File "/home/julien/.local/lib/python3.10/site-packages/oggm/utils/_downloads.py", line 263, in get_dl_verify_data
    data = pd.read_hdf(verify_file_path, key=section)
  File "/usr/lib/python3.10/site-packages/pandas/io/pytables.py", line 429, in read_hdf
    store = HDFStore(path_or_buf, mode=mode, errors=errors, **kwargs)
  File "/usr/lib/python3.10/site-packages/pandas/io/pytables.py", line 572, in __init__
    tables = import_optional_dependency("tables")
  File "/usr/lib/python3.10/site-packages/pandas/compat/_optional.py", line 141, in import_optional_dependency
    raise ImportError(msg)
ImportError: Missing optional dependency 'pytables'.  Use pip or conda to install pytables.
```

fmaussion:
Thanks! Yes indeed. Note that setup.cfg does not list all OGGM requirements, only the necessary ones. PyTables should definitely be added though

fmaussion:
[...]

Issue #1429, opened 2022-06-25
A problem during installing OGGM using the yml file


Keeptg:
Hello! I'm not sure if this is a widespread problem, so I report it here just in case.
I encountered an error (see the detailed error in the end) during create an oggm environment by using the yml file in mamba. 
While when I add `- python=3.9` in the [dependencies ](https://github.com/OGGM/oggm/blob/157c44b5b63c2a848ffb318afe77da4a18223a0d/docs/environment.yml#L4), the OGGM environment will be created successfully.

The error I encountered is as following:
```
Pip subprocess error:
  Running command git clone --filter=blob:none --quiet https://github.com/OGGM/pytest-mpl /tmp/pip-req-build-9fec9bdl
  fatal: unable to access 'https://github.com/OGGM/pytest-mpl/': gnutls_handshake() failed: The TLS connection was non-properly terminated.
  error: subprocess-exited-with-error
  
  × git clone --filter=blob:none --quiet https://github.com/OGGM/pytest-mpl /tmp/pip-req-build-9fec9bdl did not run successfully.
  │ exit code: 128
  ╰─> See above for output.
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: subprocess-exited-with-error

× git clone --filter=blob:none --quiet https://github.com/OGGM/pytest-mpl /tmp/pip-req-build-9fec9bdl did not run successfully.
│ exit code: 128
╰─> See above for output.

note: This error originates from a subprocess, and is likely not a problem with pip.

failed

CondaEnvException: Pip failed
```

fmaussion:
Let's address this as part of the "Install OGGM documentation" overhaul.

Issue #1416, opened 2022-05-06
Difficulties to pass the test after installation - 2022-05-06


nchampollion:
I installed OGGM on an linux environment, and lot of tests failed. I had to make several changes especially about the tolerance in order that the tests success. I do not understand the reason why (changes in some data for example) and if I am the only one in that case.

[Here is the test output](https://drive.google.com/file/d/1lLDaOWomFSxZSkfMZ0rUO7qw5lDFRu1G/view?usp=sharing)

List of changes I had to do.

In test_models.py:
- Line 98 `assert_allclose([lons[0], lats[0]], [10.7470, 46.8048], atol=1e-3)`
- Line 99 `assert_allclose([lons[-1], lats[-1]], [10.8551, 46.8376], atol=1e-3)`
- Line 412 `mu_star: 199.92`

In test_prepro.py:
- Line 389 `np.testing.assert_allclose(dfh['Zmed'], entity.Zmed, atol=25)`
- Line 390 `np.testing.assert_allclose(dfh['Zmax'], entity.Zmax, atol=25)`
- Line 391 `np.testing.assert_allclose(dfh['Zmin'], entity.Zmin, atol=25)`
- Line 636 `np.testing.assert_allclose(d['downstream_line'].length, 12, atol=2.5)`
- Line 835 `np.testing.assert_allclose(df.dx.sum(), entity['Lmax'], rtol=0.25)`
- Line 1129 `self.assertTrue(utils.rmsd(h1*100*50, h2*100*50) < 2)`
- Line 1725 `cfg.PARAMS['min_mu_star'] = 4`
- Line 3176 `assert df['calving_rate_myr'] > 300`
- Line 3199 - 3120 `assert df['calving_mu_star'] >= mu_bef * frac
                               np.testing.assert_allclose(df['calving_flux'], df['calving_law_flux'], rtol=0.5, atol=0.5)`

In test_shop.py:
- Line 571 `assert utils.rmsd(ref, mine) < 5`

I did not do a pull request because I think some of my changes are not good but let me know.

TimoRoth:
I've been observing those issues on Anaconda for a while now, and it seems like everytime I check if they are resolved, more tests fail.
Meanwhile, an environment that was manually installed via pip, with all the same latest versions, works perfectly fine.
As do our Docker images.

I'm once again at the end of my wits with getting Anaconda to behave, I have no idea what's wrong. It must be something Anaconda-Specific. But what specifically? I don't know.

[...]

Issue #1395, opened 2022-03-20
tests are failing in a fresh conda envornment as of 2022.03.22


fmaussion:
Installing a fresh conda env leads to failing tests which (unfortunately) hint at difference in numerical results and / or projections, which is possibly hifing larger changes in the OGGM model results.

I can't reproduce this in a py3.8 + pip envorinomenent, but will continue to investigate... 

```
===================================================================== short test summary info =====================================================================
FAILED oggm/tests/test_models.py::TestInitPresentDayFlowline::test_init_present_time_glacier - AssertionError: 
FAILED oggm/tests/test_models.py::TestMassBalanceModels::test_repr - AssertionError: assert '<oggm.MassBa... - ye: 2003\n' == '<oggm.MassBa... - ye: 2003\n'
FAILED oggm/tests/test_prepro.py::TestGIS::test_glacier_masks_other_glacier - AssertionError: 
FAILED oggm/tests/test_prepro.py::TestCenterlines::test_downstream - AssertionError: 
FAILED oggm/tests/test_prepro.py::TestElevationBandFlowlines::test_irregular_grid - AssertionError: 
FAILED oggm/tests/test_prepro.py::TestGeometry::test_width - AssertionError: False is not true
FAILED oggm/tests/test_prepro.py::TestClimate::test_climate_qc - oggm.exceptions.MassBalanceCalibrationError: RGI50-11.00897: mu* out of specified bounds: 4.88
FAILED oggm/tests/test_prepro.py::TestColumbiaCalving::test_find_calving_eb - assert 346.5359448086526 > 400
FAILED oggm/tests/test_shop.py::Test_bedtopo::test_add_consensus - AssertionError: assert <xarray.DataArray 'consensus_ice_thickness' ()>\narray(3.39654, dtype=...
============================================== 9 failed, 204 passed, 140 skipped, 979 warnings in 212.31s (0:03:32) ===============================================
```

<details>

```
(oggm_env_formlist) mowglie@flappi-top:~/Documents/git/oggm-fork$ pytest .
================================================== test session starts ==================================================
platform linux -- Python 3.9.10, pytest-7.1.1, pluggy-1.0.0
Matplotlib: 3.5.1
Freetype: 2.10.4
rootdir: /home/mowglie/disk/Dropbox/HomeDocs/git/oggm-fork, configfile: pytest.ini
plugins: mpl-0.130, anyio-3.5.0
collected 353 items                                                                                                     

[...]

PR #1849, opened 2026-02-25
build: move to pyproject

### LINKED ISSUE #1774: Enhancement: rework dependencies

gampnico:
New features:

- Refactors installation to use pyproject instead of setuptools.
  - Install OGGM and its relevant dependencies with pip, conda, or uv (no more .yaml files).
- Adds support for `uv`.
<del>- Refactors installation documentation to reflect the new installation procedure.</del> - separate PR

Todo:
- [x] ensure dynamic versioning correctly reads from .git_archival when necessary.
- [x] verify conda packaging.
<del>- [ ] document uv features new users may find useful.</del> - separate PR

Closes #1774, https://github.com/OGGM/oggm/issues/990, https://github.com/OGGM/oggm/issues/1482, https://github.com/OGGM/oggm/issues/1429, https://github.com/OGGM/oggm/issues/1395, https://github.com/OGGM/oggm/issues/620, https://github.com/OGGM/oggm/issues/1416, https://github.com/OGGM/oggm/issues/883.

- [x] Tests added/passed
<del>- [ ] Fully documented</del> - separate PR


pyproject.toml
```diff
@@ -1,9 +1,153 @@
...
-requires = [
-    "setuptools>=61",
-    "setuptools_scm[toml]>=7.0",
-    "wheel",
+requires = ["uv_build >= 0.9.28, <0.11.0"]
+build-backend = "uv_build"
+
+[project]
+name = "oggm"
+version = "1.6.2"
```
fmaussion: This needs a change - it wont work with how we release OGGM and how setuptools decides on the version name when installing with pip.

pyproject.toml
```diff
@@ -1,9 +1,153 @@
 [build-system]
```
fmaussion: do we need to remove setuptools? See comment below related to versioning

TimoRoth:
What I still have trouble finding info on is if hatch-vcs uses .git_archival.txt to get version info.
Its documentation does not say anything about it. But if I `pip install https://github.com/gampnico/oggm/archive/refs/heads/refactor-uv-installation.zip` in a test venv, it versions it as `oggm-1.6.3.dev64+gada4b03f8`, which is strange, since it's the correct commit hash, and the correct offset, but the version number is wrong, since it should be 1.6.2, as git archival says `v1.6.2-64-gada4b03f8`.
Given the commit hash is nowhere else to be found in that archive except in the archival file, it must at least be looking at it.

Otherwise, this looks good to me.

pyproject.toml
```diff
@@ -1,9 +1,184 @@
...
+docs = [
+    "numpydoc>=1.10.0",
+    "pydata-sphinx-theme>=0.16.1",
+    "sphinx>=9.0.4",
+    "sphinx-book-theme>=0.3.2",
+    "sphinx-intl>=2.3.2",
+    "sphinx-reredirects>=1.1.0",
+    "sphinx-togglebutton",
+]
+full = [
```
fmaussion: Why is this repeted from the above, this time with pins?

TimoRoth:
> On the subject of pins, @TimoRoth would you like me to use the pins you've set in [the Dockerfile](https://github.com/OGGM/OGGM-Anaconda/blob/84fe467bd04830004d6aa2cd1bccfa21d74df43e/build/oggm/meta.yaml#L24)?

That's not a Dockerfile? Is this the right link, or did you mean the Anaconda-Recipe?
Neither the Dockerfile in https://github.com/OGGM/OGGM-Docker/ nor the Anaconda-Recipes contain any pins though, only minimum versions for a few packages, that should be mostly irrelevant nowadays since they're years old and should always be fulfilled.

Since those repos run automatic CI we notice breakages as they happen, and try to fix them.

TimoRoth:
There is no pypi release of our pytest-mpl fork, so probably needs to have the direct github URL there.

Issue #1799, opened 2025-08-29
Error while downloading RGI file from glims website


albangossard:
Hi, since yesterday a new bug appeared:
```python
import oggm

rgi_version = "60"
region = "06"
shp_path = oggm.utils.get_rgi_region_file(region, version=rgi_version)
```
which results in an error
```
Traceback (most recent call last):
  File "/home/gossarda/Documents/glacio/MassBalanceMachine/bug_download.py", line 5, in <module>
    shp_path = oggm.utils.get_rgi_region_file(region, version=rgi_version)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gossarda/anaconda3/envs/MassBalanceMachine/lib/python3.11/site-packages/oggm/utils/_downloads.py", line 1975, in get_rgi_region_file
    rgi_dir = get_rgi_dir(version=version, reset=reset)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gossarda/anaconda3/envs/MassBalanceMachine/lib/python3.11/site-packages/oggm/utils/_downloads.py", line 1884, in get_rgi_dir
    return _get_rgi_dir_unlocked(version=version, reset=reset)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gossarda/anaconda3/envs/MassBalanceMachine/lib/python3.11/site-packages/oggm/utils/_downloads.py", line 1932, in _get_rgi_dir_unlocked
    raise RuntimeError(f'Could not download RGI file: {dfile}')
RuntimeError: Could not download RGI file: http://www.glims.org/RGI/rgi60_files/00_rgi60.zip
```

It looks like the whole `rgi60_files` folder is no longer accessible on GLIMS website. I don't know if it's a temporary bug that will eventually be fixed on GLIMS website in the coming days or if the paths in the OGGM code should be updated.
I'm working with OGGM 1.6.2.
Do you have an idea of the timeline for a bugfix?
Thanks!

albangossard:
Thank you for your quick reply! I'll just change the `version` as you suggest.

PR #1886, opened 2026-04-08
fix rgi_entities version

fmaussion:


fmaussion:
Also closes https://github.com/OGGM/oggm/issues/1799

Issue #1857, opened 2026-03-03
Getting rid of pytables dependency


fmaussion:
Should not be a huge amount of work, but still considerable. Should be part of 1.7

gampnico:
- I've decided to use ZSTD for parquet compression, rather than the default snappy (benchmarks below). I think it's the best all-rounder.
- I'd like to add parquet equivalents for the CSV files, parquet is considerably more efficient for larger CSVs. This is a batch conversion, so little extra effort needed.
- Even though smaller CSVs converted to parquet take longer to read, a 0.5 microsecond delay is not really something a user would notice...
- Different file sizes between different compressions are relatively meaningless for anything below a system's page size since they occupy the same , so my priority for very small files (<4KiB) is speed.

I will handle this as three PRs:
- `oggm-sample-data`, minimal `rgi62_itmix_df.parquet` to get the tests up and running
- `oggm-sample-data`, full conversion of all hdf and csv
- `oggm`, with the new parquet functionality. This will temporarily replace `SAMPLE_DATA_GH_REPO` to my own branch until `oggm-sample-data` is fully updated, and the tests for `oggm` pass. Otherwise it will try to pull demo files that don't exist.
# Benchmarks

## HDF O(MB) `rgi62_itmix_df.h5`
|  | Mean (ms) | Std Dev (ms) | Size (KB) |
|---|---|---|---|
| HDF | 86.6 | 3.0 | 1398 |
| Snappy | 6.35 | 0.21 | 2973 |
| GZIP | 9.83 | 0.23 | 2084 |
| Brotli | 10.9 | 0.2 | 1612 |
| ZSTD | 6.09 | 0.19 | 1831 |

## CSV O(MB) `copdem30_2022_1.csv`
|  | Mean (ms) | Std Dev (ms) | Size (KB) |
|---|---|---|---|
| CSV | 16.7 | 0.5 | 2236 |
| Snappy | 3.84 | 0.11 | 1007 |
| GZIP | 6.97 | 0.30 | 608 |
| Brotli | 5.49 | 0.17 | 495 |
| ZSTD | 3.38 | 0.15 | 585 |

## CSV O(KB) `hef_lengths.csv`
|  | Mean (µs) | Std Dev (µs) | Size (KB) |
|---|---|---|---|
| CSV | 347 | 9 | 1.057 |
| Snappy | 928 | 33 | 2.492 |
| GZIP | 939 | 25 | 2.153 |
[...]

PR #1887, opened 2026-04-10
refactor: migrate compile_* output from HDF to Parquet

sebastianbreguel:
Replaces pytables/HDF5 output with Parquet in `compile_fixed_geometry_mass_balance` and `compile_ela` (`utils/_workflow.py`), plus the corresponding test in `test_utils.py`.

Scope is intentionally small — just these two functions. Happy to follow up with the remaining `read_hdf` calls (`_downloads.py`, `glathida.py`, `workflow.py`) in a separate PR once this lands.

Partial fix for #1857

### DIFF
--- a/oggm/tests/test_utils.py
+++ b/oggm/tests/test_utils.py
@@ -3083,6 +3083,13 @@ class TestELAComputation(unittest.TestCase):
 
         fpath = os.path.join(cfg.PATHS['working_dir'], 'ELA.csv')
         ela1 = pd.read_csv(fpath, index_col=0)
-        ela2 = pd.read_parquet(fpath.replace('.csv', '.parquet'))
+        # Read non-csv format (parquet if pyarrow available, else hdf)
+        non_csv_path = fpath.replace('.csv', '.parquet')
+        if not os.path.exists(non_csv_path):
+            non_csv_path = fpath.replace('.csv', '.hdf')
+        if non_csv_path.endswith('.parquet'):
+            ela2 = pd.read_parquet(non_csv_path)
+        else:
+            ela2 = pd.read_hdf(non_csv_path)
 
         assert_allclose(ela1, ela2, rtol=1e-3)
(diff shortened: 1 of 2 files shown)


fmaussion:
Thanks for this. Given the scale of the change (a new dependency added to OGGM, several files outside of the OGGM codebase need to change, etc.) I don't think the change can be tackled this way. Leaving this open, and perhaps we can free up some resources soon to make this happen.

sebastianbreguel:
Thanks for the feedback! I understand the concern about scope — happy to coordinate this more broadly.

A couple of questions:
- Is there a preferred way to track the full migration plan? I can open an issue listing all the HDF callsites across OGGM (and downstream) so we have a complete picture before writing more code.
- Regarding the dependency: `pyarrow` is already pulled in transitively by pandas in the CI environments (the Parquet read/write tests passed). Would it help if I verified whether it needs to be an explicit dependency or if the transitive one is sufficient?

Let me know how you'd like to approach this — happy to adapt to whatever workflow works best for the team.

fmaussion:
> Is there a preferred way to track the full migration plan? I can open an issue listing all the HDF callsites across OGGM (and downstream) so we have a complete picture before writing more code.

Yes that would be useful. But you'll see that there are calls to open files such as https://cluster.klima.uni-bremen.de/~oggm/geodetic_ref_mb/hugonnet_2021_ds_rgi60_pergla_rates_10_20_worldwide_filled.hdf which means we have to (manually) convert files ourselves on the cluster beyond trivial code changes.

>Regarding the dependency: pyarrow is already pulled in transitively by pandas in the CI environments (the Parquet read/write tests passed). Would it help if I verified whether it needs to be an explicit dependency or if the transitive one is sufficient?

I don't understand the question. The tests are red on CI now? What does "pulled in transitively" mean?

sebastianbreguel:
Fixed the CI failures — the issue was that `pyarrow` is not installed in the CI Docker images, so `pd.to_parquet()` / `pd.read_parquet()` raised `ImportError`.

**Fix:** Added a `_save_df_non_csv` helper that tries parquet first and falls back to HDF when pyarrow is unavailable. Updated the test to handle both formats. This way the migration works where pyarrow exists and doesn't break where it doesn't.

Regarding your earlier question about "pulled in transitively" — I meant that some pandas installations bring pyarrow as an optional dependency, but that's clearly not the case in OGGM's CI images. The fallback to HDF handles this cleanly now.

PR #1888, opened 2026-04-10
feat: save partial output when glacier exceeds domain boundaries

### LINKED ISSUE #1822: Save OGGM projection output even if glacier grows out of boundary
Currently a glacier fails completely if it grows out of the domain at any given time when doing projections (error message: "Glacier exceeeds boundaries"). That means, we do not get any information on the glacier projections, even not for those years where the glacier has not yet grown out of the domain. 

Maybe it would be nice to output the projections until the year where the glacier fails, and then add  NaNs after the time period where the glacier "grew" out of the boundary. Eventually one could also tag the output by some additional information to clarify the "partial failure" of that glacier. 

This would be specifically interesting for very long "stabilisation" or overshoot scenarios at relatively cold final temperatures. 

I am not 100% sure where to best start to tackle this issue, but I guess, running a projection where a glacier grows out of the boundary, as in this test here: https://github.com/OGGM/oggm/blob/1c68657c43397efe7a1bb8e1f917854156dce8e2/oggm/tests/test_numerics.py#L1204 and then debugging could be a good start ;-).  

Eventually it would mean to output the data just before this line here : https://github.com/OGGM/oggm/blob/1c68657c43397efe7a1bb8e1f917854156dce8e2/oggm/core/sia2d.py#L187 (just by doing sth. similar as in the last lines of `run_until_and_store` )
or adapt the "stop_if_border=False" within https://github.com/OGGM/oggm/blob/1c68657c43397efe7a1bb8e1f917854156dce8e2/oggm/core/sia2d.py#L176.

sebastianbreguel:
## Problem

When a glacier grows out of the domain during projections, `run_until` raises `RuntimeError('Glacier exceeds domain boundaries')` and **all** output is lost — including valid data from years before the failure. This is particularly painful for long stabilisation or overshoot scenarios at cold temperatures (#1822).

## Solution

Catch the boundary error inside `run_until_and_store`'s main loop (line 1299), save whatever has been computed up to that point, and clean up NaN-filled trailing timesteps using the same `dropna` pattern already used by `stop_criterion`.

Changes (16 lines, single file):

1. **Try/except** around `self.run_until(yr)` — catches only `'domain boundaries'` errors, re-raises everything else
2. **`exceeded_boundaries` flag** — triggers `dropna` on `diag_ds`, `geom_ds`, and `fl_diag_dss` (same pattern as `stop_criterion`)
3. **`exceeded_domain_boundaries` attribute** on the diagnostics dataset — so downstream code can detect partial runs

## What stays the same

- Other `RuntimeError`s (CFL, nan, etc.) still propagate normally
- `run_until` itself is unchanged — the 2D model (`sia2d.py`) is not touched
- No new dependencies

Closes #1822

### DIFF
--- a/oggm/core/dynamic_spinup.py
+++ b/oggm/core/dynamic_spinup.py
@@ -13,7 +13,7 @@ from scipy import interpolate
 import oggm.cfg as cfg
 from oggm import utils
 from oggm import entity_task
-from oggm.exceptions import InvalidParamsError, InvalidWorkflowError
+from oggm.exceptions import InvalidParamsError, InvalidWorkflowError, DomainBoundariesError
 from oggm.core.massbalance import (MultipleFlowlineMassBalance,
                                    ConstantMassBalance,
                                    MonthlyTIModel,
@@ -627,9 +627,9 @@ def run_dynamic_spinup(gdir, init_model_filesuffix=None, init_model_yr=None,
                     else:
                         is_ice_free_end = False
 
-                except RuntimeError as e:
+                except (RuntimeError, DomainBoundariesError) as e:
                     # check if glacier grow to large
-                    if 'Glacier exceeds domain boundaries, at year:' in f'{e}':
+                    if isinstance(e, DomainBoundariesError) or 'Glacier exceeds domain boundaries, at year:' in f'{e}':
                         # ok we where outside the domain, therefore we search
                         # for a new lower limit for t_spinup in 0.1 °C steps
                         is_out_of_domain = True
--- a/oggm/core/flowline.py
+++ b/oggm/core/flowline.py
@@ -1291,12 +1291,27 @@ class FlowlineModel(object):
 
         # Run
         prev_state = None  # for the stopping criterion
+        exceeded_boundaries = False
+        # Always check boundaries so we can catch and handle gracefully
+        original_check = self.check_for_boundaries
+        self.check_for_boundaries = True
         for i, (yr, mo) in enumerate(zip(monthly_time, months)):
 
             if yr > self.yr:
                 # Here we model run - otherwise (for spinup) we
                 # constantly store the same data
-                self.run_until(yr)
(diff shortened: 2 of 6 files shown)


fmaussion:
Thanks for this attempt - this looks interesting but needs testing. Last time @lilianschuster had a look the fix looked more involved than this, esp. to create output files which have the expected length but with NaNs in them. Also, there should be an option in cfg.PARAMS to control whether to error or not when this happens. The tests should cover if the compile_* functions still work when one of several glaciers fail, etc. 

I definitely think this is possible but the fix requires quite more work than the current PR suggests. I'm happy to support this, but do you have the time and commitment to dig into this @sebastianbreguel ?

sebastianbreguel:
Thanks for the feedback @fmaussion! I've reworked the PR to address your points:

**Changes:**

- **Full-length NaN output**: When `error_when_glacier_reaches_boundaries=False`, `run_until_and_store` now keeps the full time dimension with NaN for remaining timesteps (arrays are pre-allocated as NaN, so I just removed the `dropna` calls for the boundary-exceeded case).
- **Uses existing param**: Reuses `cfg.PARAMS['error_when_glacier_reaches_boundaries']` — when `True` (default), behavior is unchanged (raises). When `False`, catches the error and saves partial output with NaN.
- **Internal boundary check**: `run_until_and_store` temporarily forces `check_for_boundaries=True` so it can always detect the boundary exceeded event, then decides whether to re-raise or save based on the original param value.
- **Tests**:
  - `test_save_output_on_boundary_exceeded` in `test_numerics.py` — verifies single glacier produces full-length output with NaN after failure point
  - `test_compile_run_output_with_boundary_exceeded` in `test_utils.py` — runs two glaciers (one succeeds, one exceeds boundaries), verifies `compile_run_output` merges them correctly with NaN for the failed glacier

Happy to iterate further if needed!

PR #1894, opened 2026-04-13
Round DEM elevations for better compression

### LINKED ISSUE #1864: New gdirs still larger in size because of less efficient DEM compression
After https://github.com/OGGM/oggm/issues/1856 was fixed, I re-processed a big chunk of the gdirs I did for v1.6.3 (not all of them, but the default ones at least). We did save a chunk of data (+44% instead of +57% for the offending L4 border 80 file `RGI60-11.01.tar`) but still I got intrigued by the difference in size.

The bigger file in the old and new gdirs was always `gridded_data`, but it went from 336K to 436K (+30%) for the largest glacier in the file (`RGI60-11.01450`). Turns out, the difference stems exclusively from how well the compression works in both cases. For the "old" gdirs, data came from NASADEM which stored the DEM as int16, i.e. the elevations where given in rounded meters. COPDEM stores data as floats.

To test this, I rounded the `topo` variable in the netCDF file than stored it again, and indeed I was able to "rewin" the 30% with this simple change (note that I could win another 100K by rounding the smoothed_topography as well).

Given that `dem.tif` and `gridded_data.nc` are stored an unbelievable amount of times across many experiments, it's really worth thinking if it's really "worth it" to have such precisions in the DEM, when data used to be rounded before... Changing this however will lead to non-trivial changes to everything we've done in 1.6.3 so far, but it's worth wondering if we can't round many of the variables we are creating with OGGM in `gridded_data.nc`.

Pinging @pat-schmitt @lilianschuster @gampnico for attention

sebastianbreguel:
## Summary
- COPDEM stores elevations as floats, causing ~30% larger `gridded_data.nc` files vs NASADEM (int16)
- Round DEM data to nearest meter before writing to netCDF and GeoTIFF
- Applied in `reproject_dem`, `process_dem`, and `simple_glacier_masks`
- Meter precision is sufficient for glacier modeling and restores compression efficiency

Closes #1864

## Test plan
- [ ] Run `test_prepro` to verify DEM processing still works correctly
- [ ] Compare `gridded_data.nc` file sizes before/after on a sample glacier
- [ ] Verify float DEM sources (COPDEM) produce rounded values
- [ ] Verify integer DEM sources are unaffected

### DIFF
--- a/oggm/core/gis.py
+++ b/oggm/core/gis.py
@@ -447,6 +447,9 @@ def reproject_dem(dem_list, dem_source, dst_grid_prop, output_path):
             dst_nodata=nodata,
             # Configuration
             resampling=resampling)
+        # Round float DEMs to nearest meter for better compression (#1864)
+        if np.issubdtype(dst_array.dtype, np.floating):
+            np.round(dst_array, out=dst_array)
         dest.write(dst_array, 1)
 
     for dem_ds in dem_dss:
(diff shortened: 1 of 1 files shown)


gampnico:
@sebastianbreguel @fmaussion maybe in future linting should be done by CI after the PR is approved? Currently this change appears to be just five lines (474-476, 917-918, 2873), and 1367 additions/deletions for linting which is tricky to sort through, especially when the functions indicated in the summary aren't the functions the fix was applied to.

@sebastianbreguel please note the contributing guideline: "Keep style fixes to a separate commit to make your pull request more readable."

sebastianbreguel:
Fair point — force-pushed a clean commit with only the fix (+11/-0). Thanks!

PR #1895, opened 2026-04-13
Replace get_lock() wrappers with @locked_func decorator

### LINKED ISSUE #1845: Review occurences of get_lock() and replace with lock_func decorator

sebastianbreguel:
## Summary
- Replace 14 `wrapper + _unlocked` function pairs with `@locked_func` decorator
- Move `locked_func` definition right after `get_lock()` to avoid forward references
- Remove all `_unlocked` suffixes — decorated functions keep the public API names
- Inline `with get_lock():` usages in other files (loops, partial locking) left as-is

Closes #1845

## Test plan
- [ ] Verify imports work: `from oggm.utils._downloads import get_lock, locked_func`
- [ ] Run download-related tests to confirm locking behavior preserved
- [ ] Grep for `_unlocked` — should return zero matches
- [ ] Verify no circular import or initialization-order issues

### DIFF
--- a/oggm/utils/_downloads.py
+++ b/oggm/utils/_downloads.py
@@ -50,11 +50,16 @@ except ImportError:
 
 # Locals
 import oggm.cfg as cfg
-from oggm.exceptions import (InvalidParamsError, NoInternetException,
-                             DownloadVerificationFailedException,
-                             DownloadCredentialsMissingException,
-                             HttpDownloadError, HttpContentTooShortError,
-                             InvalidDEMError, FTPSDownloadError)
+from oggm.exceptions import (
+    InvalidParamsError,
+    NoInternetException,
+    DownloadVerificationFailedException,
+    DownloadCredentialsMissingException,
+    HttpDownloadError,
+    HttpContentTooShortError,
+    InvalidDEMError,
+    FTPSDownloadError,
+)
 
 # Python 3.12+ gives a deprecation warning if TarFile.extraction_filter is None.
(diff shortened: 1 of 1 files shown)


TimoRoth:
Why all the unrelated and random changes to indentation, " vs. ' and whitespaces?
Something like that should be in its own commit and ideally PR, so the actual change can be reviewed more easily.

Also, I don't think python really has the concept of a forward-declaration like C/C++ do.

This is also potentially very risky. The multiprocessing-Lock is not recursive.
The whole reason that pattern of locked and unlocked function emerged because if you need to call the function from within a context that already holds the lock, calling the locked one would deadlock.
With this change, that ability is removed, and if one of the functions get called from inside the other, all processes will lock up.

sebastianbreguel:
Cerrando esta PR — los puntos del review son válidos:

1. El diff mezcla el refactor con reformateo masivo (quotes, whitespace, rewrapping) que hace imposible revisar el cambio real.
2. La justificación de 'forward references' no aplica en Python.
3. Más importante: colapsar los pares `foo`/`_foo_unlocked` en un solo `@locked_func` es riesgoso porque `multiprocessing.Lock` no es reentrante — si alguna función decorada llama a otra decorada, se produce deadlock. El patrón original existía precisamente para evitar eso.

Gracias por el review. Cierro esto; si vuelvo al tema voy a abrir una PR separada solo con el cambio de locking (usando `RLock` o manteniendo el patrón locked/unlocked para los call sites que lo necesiten).

PR #1899, opened 2026-04-22
Incorporate SfcTypeTIModel in FlowlineModel and run_with_climate_data

pat-schmitt:
In this PR I add support for `SfcTypeTIModel` and `DailyTIModel` within the dynamic module (no `run_with_hydro` included yet).

In particular, I adapted `FlowlineModel` to provide separate ice and firn outputs and to incorporate firn heights from `SfcTypeTIModel` in the elevation feedback of the mass-balance calculation (for this I added a very experimental firn densification model). Furthermore, I added an option in `run_from_climate_data` to store the final firn buckets of `SfcTypeTIModel` to a file, and to continue a run by loading the exact state from that file (useful for projection runs).

Here is a more detailed list of changes introduced in this PR:

- General changes:
  - renamed `rho` to `ice_density` everywhere to make it more explicit, now that `snow_density` is involved
  - added `mass` as a new potential diagnostic (especially interesting when incorporating firn output from `SfcTypeTIModel` with varying density)

- `SfcTypeModel`:
  - added experimental densification model: currently it assumes a linear relation between firn `melt_f` and `density`: both evolve following the same exponential function as described in Schuster et al. (2023). The fresh snow density is set to 300 kg m-3. **This is currently very experimental and not validated**.
  - with density we have new diagnostics `buckets_thickness_m` (for each bucket, useful for plotting density profiles for testing and validation later) and `columns_thickness_m` (the total thickness of all buckets of one column).
  - new kwarg `include_mb_model_heights` in `get_{annual,monthly,daily}_mb`: if `True`, we add the current bucket thickness on top of the provided heights to include the elevation feedback of the bucket heights as well
  - new methods `save_to_file` (saving the complete current state, including firn buckets) and `load_from_file` (returns an initialized `SfcTypeTIModel` at the same state as when it was saved). The same methods were also added to `MultipleFlowlineMassBalance` for supporting multiple flowlines. To save the actual state I introduced new `mb_diagnostics` files (if `MultipleFlowlineMassBalance` creates it, the dataset is organized with groups, similar to `fl_diagnostics`)

- `FlowlineModel`:
  - new diagnostic `mass` in kg and Mt: important when including firn output, because the density conversion between volume and mass is no longer fixed at 900 kg m-3 (see above under `SfcTypeModel`)
  - new kwarg `include_mb_model_heights`: includes snow bucket heights for elevation feedback (see above under `SfcTypeModel`)
  - new kwarg `include_firn_outputs`: if `True`, the output provided for `volume`, `mass`, and `thickness` is total = ice + firn. Firn outputs are available in glacier-integrated output and `fl_diagnostics`
  - added support for daily time resolution and leap years:
    - This means the actual number of days in each month is used. Previously, `SEC_IN_MONTH` was used for converting the model output from s-1 to month-1 (especially important for `DailyTIModel`).
    - It is debatable whether the added code complexity for leap year support is worth the small impact it has; I still added it for completeness. Leap years are used if the provided `mb_model` has `use_leap_years=True`, otherwise we use the old default `SEC_IN_YEAR` for yearly seconds. (months always use number of days per month to get to seconds in months, see bullet point above. Feb. always 28 days when `use_leap_years=False`)

- `run_from_climate_data`:
  - new kwarg `mb_diagnostics_filesuffix`: if provided, a `MultipleFlowlineMassBalance` is initialized by loading a past `mb_model` state (only supported for `SfcTypeTIModel` at the moment)
  - new kwarg `save_mb_diagnostics_filesuffix`: if provided, at the end of the run the final `mb_model` state is saved to an `mb_diagnostics` file for later use (only supported for `SfcTypeTIModel`)

- [x] Tests added/passed
- [ ] Fully documented

### DIFF
--- a/oggm/cfg.py
+++ b/oggm/cfg.py
@@ -262,6 +262,12 @@ BASENAMES['climate_historical_daily'] = ('climate_historical_daily.nc', _doc)
 _doc = "A dict containing the glacier's mass balance calibration parameters."
 BASENAMES['mb_calib'] = ('mb_calib.json', _doc)
 
+_doc = ('A group netcdf file containing the complete state of a SfcTypeTIModel '
+        'instance. This is useful for projeciton runs to start from historical '
+        'snow bucket state. Netcdf groups = fl_{i}, with i between 0 and '
+        'n_flowlines - 1')
+BASENAMES['mb_diagnostics'] = ('mb_diagnostics.nc', _doc)
+
 _doc = 'The monthly GCM climate timeseries stored in a netCDF file.'
 BASENAMES['gcm_data'] = ('gcm_data.nc', _doc)
 
--- a/oggm/core/dynamic_spinup.py
+++ b/oggm/core/dynamic_spinup.py
@@ -1021,7 +1021,10 @@ def define_new_melt_f_in_gdir(gdir, new_melt_f):
     -------
 
     """
-    if 'melt_f' not in gdir.settings:
+    # this is for backwards-compatiblity, it would be cleaner to call
+    # if 'melt_f' not in gdir.settings (only possible when mb was calibrated
+    # using the new settings system)
+    if gdir.settings['melt_f'] is None:
         raise InvalidWorkflowError('No `melt_f` in gdir.settings. '
                                    'You first need to calibrate the whole '
                                    'MassBalanceModel before changing melt_f '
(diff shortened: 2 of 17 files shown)


pat-schmitt:
Remaining test failures are related to some 'Connection broken' errors.

PR #1907, opened 2026-04-23
Improve tests runtime

pat-schmitt:
With recent developments the overall time to run the tests has increased a lot. This PR is for exploring how we could make it faster again.

- [ ] Tests added/passed
- [ ] Fully documented


pat-schmitt:
I added two new test environments: `models_dynamics` (~20 min) and `models_mb` (~15 min). This does not solve the problem of having too many slow tests, but it distributes the load a bit.

I also came across a potentially useful approach: using `pytest-xdist` together with `pytest -n auto ...` in `run_tests.sh`. The idea is to further distribute tests across all available CPUs. On GitHub Actions there are two, meaning we could potentially halve all test run times. However, for this to work, we need to be very careful that all files use correct locking.

I tested this in the current PR (commits https://github.com/OGGM/oggm/pull/1907/commits/4d9c0d34cf01aca7dbfef7be596d0b5ce0f7bb90 and https://github.com/OGGM/oggm/pull/1907/commits/41328e325ff9e9ff39228e038dc6714d7791549c), but it did not work out of the box. It may be worth revisiting in the future.

PR #1908, opened 2026-04-30
feat: read and write npz methods to replace read_pickle

### LINKED ISSUE #1903: Get rid of pickles, one pickle at a time
OGGM currently writes pickles and reads pickles. This is bad (https://github.com/OGGM/oggm/issues/1524).

We can use this issue to list each of these pickle files and discuss one by one. Some pickles are quite related (the flowlines ones) but we need an overview first.

A good step would be to look at [BSENAMES](https://github.com/OGGM/oggm/blob/76f44230b6ecd2bd3161d1e04ff426e5947a0dc5/oggm/cfg.py#L176-L301) and see which ones end in pkl.

The hope is that each pickle can be replaced by its own format. For some (outlines?), I hope there is a straightforward replacement. For most, pickle has been chosen to store OGGM objects (flowlines). But the flowlines are nothing else than a bunch of data, so it should be possible to store this data in json and then serialize the objects to json and from json. Perhaps Copilot can help with this, to a certain extent.

@gampnico your task is to list the pickles here, discuss and tackle them one by one, ideally in the order they are created in the OGGM workflow if possible.

gampnico:
This PR currently adds:
  - `read_store`, `read_npz` and `write_npz` methods which fall back to pickle if no npz files are available
  - Warnings that read_pickle methods will be deprecated for future GlacierDirectories.
  - A transcoding module `transcoder.py` that handles transcoding between existing pickles and the new npz files. These could

**NB: This PR's design has changed substantially and has moved to using npz instead of zarr for performance.**

Refs: #1903

**Points for discussion**

All pickles are now stored as npz files in a glacier directory's subdirectory, which I'm currently naming "data_store". Zarr was too inefficient for such small data.

The default replacement for `read_pickle` is now `read_store`, which handles both pickle (`read_pickle`) and numpy (`read_npz`). `read_store` falls back to `read_pickle` if no npz file exists. This preserves backwards compatibility with existing gdirs if a user selects an older URL.

I've designed this as a type registry, rather than mapping to basenames, as it's less brittle, and I don't think it would prevent basename mapping later on. This way we only need to make changes when Python/numpy introduces or deprecates a type (so basically never), rather than when we change or add a new OGGM object.

Zarr does not support all the data structures used by certain pickles (e.g. shapely objects). Previous zarr commits converted these to a compatible type. When using `read_store`, objects are converted back into the type expected by OGGM. This should minimise rewrites across the rest of the codebase. However, zarr-dependent packages (e.g. DTCG) will require compatibility wrappers.

Closes https://github.com/OGGM/oggm/issues/1903

- [x] Tests added/passed
- [x] Fully documented

---
**Keeping this section for posterity, to help explain how the design decisions evolved.**

All pickles are now stored as `xr.DataTrees` in a single zarr file, which I'm currently naming "data_store". The filename in `read_zarr(filename)` therefore reads the group, rather than entire zarr file. This avoids having multiple small zarr files, and we can maintain cross-compatibility with DTCG.

The default replacement for `read_pickle` is now `read_store`, which handles both pickle (`read_pickle`) and zarr (`read_zarr`). `read_store` falls back to `read_pickle` if no zarr store exists. This preserves backwards compatibility with existing gdirs if a user selects an older URL.

Zarr does not support all the data structures used by certain pickles (e.g. shapely objects). [oggm-zarr](https://github.com/gampnico/oggm-zarr) converts these to a compatible type. When using `read_store`, these objects are converted back into the type expected by OGGM. This should minimise rewrites across the rest of the codebase.

I'd like to upload a small sample zarr to oggm-sample-data, so this can be included in `init_hef`, and I can then add tests for reading a zarr from a gdir directly.

I'm currently keeping code for converting from pickle to zarr as a [dual-licensed package](https://github.com/gampnico/oggm-zarr), as some of this includes code from DTCG which is not compatible with OGGM's license. I'll see if I can rework this, but it may be simpler to handle conversion via `dtcg`, since this is already set up and we can build directly on the existing `GeoZarrHandler`.

You can view a [conversion workflow here](https://github.com/gampnico/oggm-zarr/blob/main/notebooks/convert_pickles.ipynb).


fmaussion:
Thanks! Looking promising

A few quick thoughts:
- I don't understand the PR yet - it does not touch the code that needs to be changed, i.e. the places where the pickles are actually written? 
- the `read_zarr`, `read_store` and `_validate_store` are written with the assumption that they are just replacing read_pickle - I think these may become more generic in the future. Maybe we need to rename them according to what they do. Alternatively they could be external functions (not methods of gdir) because they are dealing only with some specific data formats within the oggm workflow - to discuss.
- I do not want a new oggm-zarr package unless strictly necessary / there are good arguments for it - we can deal with the license like usual, by including the license in OGGM. This could be a module in OGGM if needed
- It should not be necessary to resort to oggm-sample-data for tests, as OGGM writes and reads pickles (there are no pickles in oggm-sample-data either). In theory, since the change will be mostly internal, the existing tests should "just work" after the conversion to zarr (except those where read_pickle is used), and we need to add a test for backwards compatibility, although I think https://github.com/OGGM/oggm/blob/76f44230b6ecd2bd3161d1e04ff426e5947a0dc5/oggm/tests/test_utils.py#L786 will cover a little bit of this

gampnico:
Notes based on internal meeting:

> I don't understand the PR yet - it does not touch the code that needs to be changed, i.e. the places where the pickles are actually written?

Next on agenda.

> the read_zarr, read_store and _validate_store are written with the assumption that they are just replacing read_pickle - I think these may become more generic in the future. Maybe we need to rename them according to what they do. Alternatively they could be external functions (not methods of gdir) because they are dealing only with some specific data formats within the oggm workflow - to discuss.

Agreed. `_validate_store` is for converting to and from zarr so that data is structured in a way that OGGM recognises. As discussed we may need separate helper functions for writing specific files (e.g. `write_model_flowlines`). In pseudo-code:

```python
model_flowlines: Centerline = Centerline(...)
write_store(model_flowlines, "model_flowlines")
model_flowlines: Centerline = read_store("model_flowlines")
```
where internally `read_store` and `write_store` both do something like:
```python
def write_store(self, obj, filename, filesuffix, **kwargs):
    data: xr.DataTree = _validate_store(obj)  # converts shapely objects etc. into zarr-compatible types
    write_zarr(data: xr.DataTree=data, group:str=filename, **kwargs)
```

> I do not want a new oggm-zarr package unless strictly necessary / there are good arguments for it - we can deal with the license like usual, by including the license in OGGM. This could be a module in OGGM if needed

Zarr-related code will be placed under `utils` as a public module. I'd prefer to keep `read_store` etc under `_workflow` as this is where `read_pickle` is located.

gampnico:
Failing tests:
- Discussion for hanging tests in test_utils in #1908 
- pytables dependency error is linked to #1857

gampnico:
Image diffs for the two failing tests when run locally:

result-failed-diff
result-failed-diff

I'll look into forcing formatting again like I did for salem.

The other failing tests for `untested_base:20231004` are linked to #1857.

But the good news it doesn't seem to be a numerical or logic issue.

gampnico:
I'm not sure what's happening with the dependency resolution in `untested_base:20231004`. I can't reproduce these failures locally, even when matching the package versions (numpy 1.25, xarray 2026.4, pandas 3.03, zarr 3.1.5 etc.). The error message in the logs `Cannot interpret '<StringDtype(na_value=nan)>' as a data type` is related to xarray and pandas 3.0 (see [here](https://pandas.pydata.org/docs/user_guide/migration-3-strings.html)), but running these package versions on tests locally do not produce this error.

oggm/workflow.py
```diff
@@ -132,11 +144,36 @@ def reset_multiprocessing():
...
+        if _loop is not None and not _loop.is_closed():
+            _exc = getattr(_loop, "_default_executor", None)
+            if _exc is not None:
+                _exc.shutdown(wait=True)
+                _loop._default_executor = None
+        _zs.cleanup_resources()
+        # cleanup_resources() waits only 0.2 s for iothread; ensure it is
+        # truly dead before any subsequent fork (Pool/Manager creation).
+        if _iothread is not None and _iothread.is_alive():
+            _iothread.join(timeout=5.0)  # TODO: Play around with timeout
```
gampnico, line 174: @fmaussion I introduced a timeout here which is likely the cause of the bottleneck.

fmaussion: Perhaps yes, but I'm pretty confident there is another layer of overhead. And if the timeout is reached this should raise I guess?

gampnico:
Interim benchmarks for zarr vs npz:

| | zarr | npz | |
|---|---|---|---|
| `init_hef` full build | 100.4 s | 3.3 s | 30x faster |
| `write_store` x60 | 40.76 s | 0.04 s | 1000x faster |
| `read_store` x60 | 5.34 s | 0.13 s | 41x faster |
| store on disk | 0.37 MB | 0.21 MB | 43% smaller |

gampnico:
Using `store_bench.py` you provided, the current changes on `fix-pickle-doomsday` are mostly faster than on `refactor_suggestion`.

I highlight the fastest times and smallest size for non-master branches, although in many cases `fix-pickle-doomsday` is faster.

## Task timings

| task | master | refactor_suggestion | fix-pickle-doomsday | npz_compressed | zarr |
|---|---|---|---|---|---|
| L2_compute_downstream_bedshape | 0.94 s | 1.00 s | **0.92 s** | 0.95 s | 1.84 s |
| L2_compute_downstream_line | 2.66 s | 2.79 s | **2.64 s** | **2.64 s** | 3.06 s |
| L2_elevation_band_flowline | 0.46 s | 0.47 s | **0.44 s** | 0.46 s | **0.44 s** |
| L2_fixed_dx_elevation_band_flowline | 0.12 s | **0.13 s** | **0.13 s** | **0.13 s** | 1.09 s |
| L2_simple_glacier_masks | 1.07 s | 1.10 s | 1.06 s | 1.05 s | **1.03 s** |
| L3_apparent_mb | 0.16 s | 0.23 s | 0.22 s | **0.21 s** | 1.49 s |
| L3_filter_inversion_output | 0.02 s | **0.03 s** | **0.03 s** | **0.03 s** | 1.51 s |
| L3_init_present_time_glacier | 0.08 s | **0.06 s** | **0.06 s** | **0.06 s** | 2.31 s |
| L3_mass_conservation_inversion | 0.12 s | 0.07 s | **0.06 s** | 0.07 s | 1.64 s |
| L3_mb_calibration | 0.84 s | 0.89 s | **0.83 s** | 0.84 s | 0.98 s |
| L3_prepare_for_inversion | 0.02 s | **0.02 s** | **0.02 s** | **0.02 s** | 1.33 s |
| L3_process_climate_data | 0.54 s | 0.54 s | 0.55 s | 0.53 s | **0.44 s** |
| L4_run_from_climate_data | 0.87 s | 0.95 s | 0.91 s | **0.89 s** | 1.41 s |
| READ_downstream_line | 0.55ms | 2.86ms | **2.50ms** | 2.81ms | 132.93ms |
| READ_inversion_flowlines | 0.65ms | 6.80ms | 4.27ms | **4.20ms** | 181.45ms |
| READ_inversion_input | 0.62ms | 5.71ms | **4.15ms** | 4.17ms | 155.79ms |
| READ_inversion_output | 0.74ms | 6.72ms | 5.18ms | **5.05ms** | 177.21ms |
| READ_model_flowlines | 5.05ms | 34.36ms | **26.50ms** | 27.18ms | 259.94ms |
| TAR_GZ | 0.59 s | **0.65 s** | 0.67 s | 0.68 s | 0.88 s |
| TOTAL | 9.37 s | 10.72 s | 9.94 s | **9.93 s** | 38.46 s |
| TOTAL_no_download | 8.36 s | 9.78 s | **9.04 s** | 9.08 s | 37.59 s |
| compile_glacier_statistics | 0.29 s | 0.32 s | **0.30 s** | **0.30 s** | 0.72 s |
| init_from_prepro_L1 | 1.00 s | 0.93 s | 0.88 s | **0.85 s** | 0.87 s |

## Stage sizes

On disk (MB):

| stage | master | refactor_suggestion | fix-pickle-doomsday | npz_compressed | zarr |
|---|---|---|---|---|---|
| L1 | 3.69 | 3.69 | 3.69 | 3.69 | 3.69 |
| L2 | 8.40 | **8.43** | 8.54 | 8.54 | 11.91 |
| L3 | 9.14 | **9.44** | 9.60 | 9.60 | 21.27 |
| L4 | 10.25 | **10.55** | 10.71 | 10.71 | 22.38 |

Files:

| stage | master | refactor_suggestion | fix-pickle-doomsday | npz_compressed | zarr |
|---|---|---|---|---|---|
| L1 | 94 | 94 | 94 | 94 | 94 |
| L2 | 140 | **140** | **140** | **140** | 754 |
| L3 | 200 | **200** | **200** | **200** | 2328 |
| L4 | 224 | **224** | **224** | **224** | 2352 |

gampnico:
@fmaussion this is ready for deployment.

Issue #1837, opened 2026-01-20
Optimize spinup time?


fmaussion:
I see many reasons why this might make our life much more complicated, but Brandon writes in their paper:

_The duration of the spinup period required such that modeled elevation change aligns with observations was found to vary for glaciers in this study, presumably due to differences in glacier response times. Thus, an optimization routine that identifies the best spinup duration ranging from 0-60 years was implemented by minimizing the mean absolute error between modeled and observed surface elevation change data._

PR #1914, opened 2026-05-27
Add option for longer dynamic spinups

pat-schmitt:
In this PR, I introduce a new methodology that allows for longer spinup periods during dynamic spinup, particularly when creating the preprocessed directories. When dynamic spinup was initially developed, historical data was only available from 1979, so the spinup in the prepro levels was limited to that year. However, historical data is now available from 1901, and recent studies have shown that some glaciers require a longer spinup period.

Main changes:

- New kwarg `spinup_periods_to_try` in `tasks.run_dynamic_spinup`: accepts a list of additional spinup periods, which are tried in order if both the initially defined spinup period (`spinup_period_initial`) and the minimum spinup period (`min_spinup_period`) fail. This means we first try the old default workflow, and only if this fails we try the additional spinup periods.
- New argument `--dynamic-spinup-periods-to-try` in `prepro_levels.py`, defaulting to `[30, 40, 50, 60, 70, 80, 90, 100]`. To disable this new behaviour, set `--dynamic-spinup-periods-to-try none`.
- No additional tests were added. The change was only tested by running RGI region 19 (Antarctica).
- The resulting model output `_historical_spinup` in the prepro levels still starts from 1979, which is achieved via `fixed_geometry_spinup_yr`. If `fixed_geometry_spinup_yr` (e.g. 1979) is later than `FlowlineModel.yr` (e.g. 1950), the model runs until `fixed_geometry_spinup_yr` before storing any output. This works for now, but should eventually be made more explicit. For example, by adding a new kwarg such as `start_yr_to_store_model_output`.

Related to https://github.com/OGGM/oggm/issues/1837: but no real optimization to find the best spinup time, we just look for any working spinup time, and try to have it as short as possible, up to a maximum of 100 years spinup

- [ ] Tests added/passed
- [ ] Fully documented

### DIFF
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -12,6 +12,11 @@ Enhancements
 - `run_prepro_levels` now allows to set a custom climate data processing
   module and custom geodetic data file to create glacier directories (:pull:`1910`).
   By `Fabien Maussion <https://github.com/fmaussion>`_
+- New kwarg `spinup_periods_to_try` in `run_dynamic_spinup` to be able to
+  provide a list of additional spinup periods, which are tried in order if both
+  the initially defined spinup period (`spinup_period_initial`) and the minimum
+  spinup period (`min_spinup_period`) fail (:pull:`1914`).
+  By `Patrick Schmitt <https://github.com/pat-schmitt>`_
 
 Bug fixes
 ~~~~~~~~~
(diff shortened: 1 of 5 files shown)


fmaussion:
Thanks @pat-schmitt ! Fantastic addition, can't wait to try it out

Issue #1921, opened 2026-06-09
Bug: Deprecation in multiprocessing can cause run_prepro_levels to hang


gampnico:
**Expected behaviour**

Methods calling `run_prepro_levels` should execute without hangs or errors when multiprocessing is enabled.

**Actual behaviour**

Running methods with multiprocessing enabled on Python 13 or lower invokes `os.fork()` and can randomly deadlock children. This causes tests in `test_utils` to hang indefinitely. This almost always occurs with `test_geodetic_per_glacier_and_massredis_run`, when running the full test suite with `pytest --run-slow oggm` or similarly with the entire `TestPreproCLI` class.

I cannot reproduce this issue when running the same test function individually e.g. `pytest --run-slow oggm/tests/test_utils.py::TestPreproCLI::test_geodetic_per_glacier_and_massredis_run` or similar. Maybe this is a concurrency issue with a file download/read or write?

From the Python documentation [here](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods):

> Changed in version 3.12: If Python is able to detect that your process has multiple threads, the [os.fork()](https://docs.python.org/3/library/os.html#os.fork) function that this start method calls internally will raise a [DeprecationWarning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning). Use a different start method. See the os.fork() documentation for further explanation.

From [os documentation](https://docs.python.org/3/library/os.html#os.fork):

> We chose to surface this as a warning, when detectable, to better inform developers of a design problem that the POSIX platform specifically notes as not supported. Even in code that appears to work, it has never been safe to mix threading with [os.fork()](https://docs.python.org/3/library/os.html#os.fork) on POSIX platforms. The CPython runtime itself has always made API calls that are not safe for use in the child process when threads existed in the parent (such as malloc and free).

fmaussion:
this looks important enough to me. We actually already have the infrastructure for this: the `use_mp_spawn` parameter (and the `OGGM_USE_MP_SPAWN` env variable) were added for this reason I guess (cc @TimoRoth). The issue is that it defaults to `False`, and `run_prepro_levels` doesn't activate it.

Given that Python is moving away from `fork` as the default (and will likely make `spawn` the default on Linux in 3.14), I think the cleanest fix is simply to flip the default in `params.cfg`:

```
use_mp_spawn = True
```

TimoRoth:
[...]

PR #1922, opened 2026-06-11
Set mp_spawn to True per default

fmaussion:
Closes or at least addresses https://github.com/OGGM/oggm/issues/1921

### DIFF
--- a/oggm/params.cfg
+++ b/oggm/params.cfg
@@ -31,9 +31,10 @@ use_multiprocessing = False
 mp_processes = -1
 # To avoid race issues when using GDAL, it might be necessary to set this
 # to "true". It makes the initialisation of the pool slower, which is
-# why it is false per default
+# why it was false per default until OGGM 1.7.
+# This the newer default as "False" was not a recommended setting anyways.
 # Can be overridden with the environment variable OGGM_USE_MP_SPAWN
-use_mp_spawn = False
+use_mp_spawn = True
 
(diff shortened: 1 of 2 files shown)


TimoRoth:
I'm not sure if this is a good idea.
Spawn-Mode is still not the default upstream, since it's quite troublesome at times. The new default there is forkserver.
And for us, using fork still drastically speeds things up.
Since in spawn mode, each new process will have to re-load everything from disk, and can't re-use any existing resources.
So it's a severe performance penalty, and I also think it still has hidden broken edge cases which rely on pre-loaded global state.

Using fork is only deprecated when using it combined with threads.
And we're not using threads anywhere.

fmaussion:
Thanks! That helps me understand the issue a bit more.

> Using fork is only deprecated when using it combined with threads.
> And we're not using threads anywhere.

this is where I'm fuzzy - I wonder if libraries that we are using don't use threads without us knowing, or we should make sure they don't.

TimoRoth:
That's impossible to tell, but sadly quite possible.
I guess ideally we'd just be using threads instead of entirely new processes, but python global interpreter lock makes those not really workable and effectively nearly fully synchronize.

We could try experimenting with a free-threaded python build, which are a thing now and eliminate that lock.
But I think the various binary dependencies we have, like GDAL, do not support that mode at all.

fmaussion:
Convinced!

PR #1925, opened 2026-06-18
refactor: modify glacier bundle limit

### LINKED ISSUE #1902: Changing the 1000 glaciers tarfiles to 100 glaciers tar files or 1 glacier tar file?
Related to https://github.com/OGGM/oggm/issues/1896 

Many, many years ago I took a few hasty decisions that have stuck until now.

Prehistoric glacier directories were stored directly in RGI region folders, which made linux *very* slow when navigating the directory tree of regions with well over tens of thousands of glaciers. Therefore I chose to have sub-folders of 1000 glaciers. This is ok and could be changed to 100, or a more incremental tree (e.g. RGIREG/1000/100/10) easily without much change in functionality. The only reason this exists is because many files in a folder is slow.

This choice of 1000 glaciers stuck when we started offering the gdirs for download. We create zipped tars per glaciers and then tar them again in 1000 batches. The decision process here was mostly driven by unsubstantiated worries about having "not too many files to download and unpack". This was a time where we had hash checks on files for example, to discover failed downloads. Increasing the number of files to download felt like increasing the chances of failed/partial downloads.

We should now revisit this choice. 1000 glaciers makes lots of sense for global runs, but exploratory runs or tutorials where you need only one glacier are clearly disadvantaged by this choice. 

It would be a very low hanging fruit to change this in a backwards compatible manner (OGGM could discover by itself if its a "1000" or "100"). But we need a decision on the "what is a right number" there. The most obvious answer is "1", but then a global run user would download 270k files multiplied by the level layers implemented in https://github.com/OGGM/oggm/issues/1896?issue=OGGM%7Coggm%7C1900 .

To discuss, but easily implementable (cc @gampnico @pat-schmitt )

gampnico:
Adds `bundle_size` argument to `base_dir_to_tar`, defaulting to 100.
Refactors `robust_tar_extract` and `gdir_from_tar` to be agnostic to bundle sizes, falling back to 1k bundles.
Refactors some tests into pytest-style.

Backwards compatible with existing `base_urls`, no breaking changes, and tested with the tutorial notebooks with both 1k and 100 glacier bundles for RGI6 & RGI7.

This means you can convert existing glacier directories like so:
```
gdirs = workflow.init_glacier_directories(rgi_ids, from_prepro_level=4, prepro_base_url=base_url)
workflow.execute_entity_task(utils.gdir_to_tar, gdirs, delete=True)
utils.base_dir_to_tar(WORKING_DIR, delete=True, bundle_size=100)
```
and upload the files. No need to regenerate the glacier directories.

Closes https://github.com/OGGM/oggm/issues/1902

- [x] Tests added/passed
- [x] Fully documented

### DIFF
--- a/oggm/utils/_workflow.py
+++ b/oggm/utils/_workflow.py
@@ -4268,12 +4268,16 @@ def base_dir_to_tar(
         bundles = {}
         for dirpath, _, filenames in os.walk(base_dir):
             for fname in sorted(filenames):
+                # TODO: Would this be easier/faster with pathlib?
                 if not fname.endswith(".tar.gz"):
                     continue
                 rgi_id = fname[:-7]
                 # TODO: RGI7 has a different ID length
                 # check for RGI, e.g. `centerlines_11` is also 14 chars long
-                if (len(rgi_id) != 14) or "RGI" not in rgi_id:
+                if not (
+                    (len(rgi_id) == 14 and "RGI" in rgi_id)
+                    or (len(rgi_id) == 23 and "RGI" in rgi_id)
+                ):
                     continue
                 # TODO: or maybe:
                 # bundle_name = f"{rgi_id[:-6]}.{int(rgi_id[-5:-2]) // bundle_size:03d}"
--- a/oggm/utils/_workflow.py
+++ b/oggm/utils/_workflow.py
@@ -2646,8 +2646,9 @@ def robust_tar_extract(
             # 100-glacier bundle: RGI60-11.006.tar in the region dir
             # TODO: Really ought to consider switching to pathlib!
             rgi_id = bname[:-7]  # strip .tar.gz
-            # TODO: Check support for RGI7 as ids have a different length
-            if len(rgi_id) == 14:
+            # The bundle name slices work for both RGI6 (14 char IDs) and
+            # RGI7 (23 char IDs).
+            if len(rgi_id) in (14, 23):
                 dirbname = f"{rgi_id[:-6]}.{rgi_id[-5:-2]}"  # e.g. RGI60-11.006
                 region_dir = os.path.dirname(os.path.dirname(from_tar))
                 base_tar = os.path.join(region_dir, dirbname + ".tar")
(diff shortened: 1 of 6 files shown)


fmaussion:
Looks reasonable to me - the last remaining tests seem related to the PR.

gampnico:
Thanks, just want to check today there's proper support for RGI7. I'll ping you when it's good to go

gampnico:
@fmaussion tested with RGI70, including the tutorial notebooks. Should be good to go on my end.

gampnico:
Current test failure is due to a timeout on the cluster. You can rerun the workflow if you like and it'll pass.

oggm/utils/_workflow.py
```diff
@@ -4247,46 +4248,50 @@ def base_dir_to_tar(
...
     """
 
     if base_dir is None:
         if not cfg.PATHS.get("working_dir", None):
             raise ValueError("Need a valid PATHS['working_dir']!")
         base_dir = os.path.join(cfg.PATHS["working_dir"], "per_glacier")
 
-    # TODO: get index based on the number of tens in repackage_index, e.g. 100 -> 2, 1000 -> 3, etc.
-    # index = int(np.log10(bundle_size))
+    # The read side (robust_tar_extract, gdir_from_tar, _get_prepro_gdir)
```
gampnico, line 4264: @fmaussion just to flag that I refactored ``robust_tar_extract`` to support arbitrary bundle sizes, so this edit may be a regression from the read side

fmaussion: Thanks for flagging it! I'd argue for keeping the explicit {100, 1000} check. The system only ever constructs and serves 100- and 1000-bundles. The only other possible extension might be 10, but I think 100 really is the best compromise now

fmaussion:
Thanks for this @gampnico - important and long overdue update to the gdirs

oggm/utils/_workflow.py
gampnico: Yeah, no argument from me, just noting it in relation to gdirs with daily mass balance data, with a very large `climate_historical_daily.nc` it may be worth dropping to a lower number. But I suppose this would be a temporary measure anyway if we switch away from tars completely.

PR #1929, opened 2026-06-23
Allow for runs to stop midway and still store

### LINKED ISSUE #1104: run_until_and_store still writes a model output file even if the run fails
It's not a big deal and there is probably a clever fix for that but I'm not clever at the moment.

fmaussion:
Closes https://github.com/OGGM/oggm/issues/1104

### DIFF
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -9,6 +9,18 @@ v1.x (unreleased)
 Enhancements
 ~~~~~~~~~~~~
 
+- New ``cfg.PARAMS['store_output_on_error']`` (default ``False``): when a run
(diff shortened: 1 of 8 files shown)


fmaussion:
Also closes https://github.com/OGGM/oggm/issues/1822

fmaussion:
Changed the default to False - this is more logical for most uses except projections.

This will need to be documented in the tutorials somewhere.

fmaussion:
@lilianschuster @pat-schmitt this is quite a nice addition - I've set the option to "False" per default (we don't want varying end-dates in the prepro folders) but for projections > 100 years most of the times you would set it to True

PR #1927, opened 2026-06-19
refactor: refactor ON-78 optimise rmsd tests

gampnico:
Closes #1906

- [ ] Tests added/passed
- [ ] Fully documented
- [ ] Entry in `whats-new.rst`

PR #1933, opened 2026-06-26
refactor: ON-78 optimise TestClimate and RMSD

### LINKED ISSUE #1906: Tests on RMSD are slow
@chloe-hancock , it seems that your test has a long runtime - we should discuss how to shorten it!

gampnico:
Closes https://github.com/OGGM/oggm/issues/1906

- [x] Tests added/passed
- [x] Fully documented

### DIFF
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -20,7 +20,7 @@ Enhancements
 
 Bug fixes
 ~~~~~~~~~
-- Test optimisations for climate and mass balance calibration e.g. (:pull:1933)
+- Test optimisations for climate and mass balance calibration e.g. (:pull:`1933`)
   By `Nicolas Gampierakis <https://github.com/gampnico>`_
 
 Breaking changes
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -20,7 +20,8 @@ Enhancements
 
 Bug fixes
 ~~~~~~~~~
-- Test optimisations for climate and mass balance calibration e.g. (:pull:`1933`)
+- Test optimisations for climate, mass balance calibration, and dynamic
+  spinup (:pull:`1933`).
   By `Nicolas Gampierakis <https://github.com/gampnico>`_
 
 Breaking changes
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -27,12 +27,12 @@ Enhancements
   By `Nicolas Gampierakis <https://github.com/gampnico>`_
 - Some tests have been refactored from unittest to pytest. (:pull:`1925`).
   By `Nicolas Gampierakis <https://github.com/gampnico>`_
-
-Bug fixes
-~~~~~~~~~
 - Test optimisations for climate, mass balance calibration, and dynamic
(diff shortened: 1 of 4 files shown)


gampnico:
I removed the optimisations to `dynamic_spinup` from #1927 like you asked, but actually the changes there had the greatest performance boost. Let me know if I should include these as well.

fmaussion:
Thanks! this looks OK to me, but I guess it doesn't change much, if anything, in the runtime of the tests? Or does it?
Are the removal of the deepcopy's safe? Or does @pat-schmitt need to confirm?

> I removed the optimisations to dynamic_spinup from https://github.com/OGGM/oggm/pull/1927 like you asked, but actually the changes there had the greatest performance boost. Let me know if I should include these as well.

I still think it would be cleaner to have one PR for dynamic spinup and one for RMSD. The RMSD tests probably need more substantial change: I think the point was that many tests are testing the same thing hence not all necessary, but I haven't looked in the details.

fmaussion:
> but actually the changes there had the greatest performance boost

Yes! If they impact performance this would be good! But just in their own PR.

pat-schmitt:
> Are the removal of the deepcopy's safe? Or does @pat-schmitt need to confirm?

Yes, I think those deepcopy's could be removed (if tests are passing).

> The RMSD tests probably need more substantial change: I think the point was that many tests are testing the same thing hence not all necessary, but I haven't looked in the details.

Yes, in the tests the test all possible combinations of calibration parameters. This is the correct way of doing things, but only if test are not taking too long. However, after the comment here https://github.com/OGGM/oggm/issues/1906#issuecomment-4825728760 I run the test locally again and it now only takes ~1min. I am not sure what happened (maybe also somehow related to download speeds on the cluster?), but with this I think we can leave it as it is.

docs/whats-new.rst
```diff
@@ -20,6 +20,8 @@ Enhancements
 
 Bug fixes
 ~~~~~~~~~
+- Test optimisations for climate and mass balance calibration e.g. (:pull:1933)
```
fmaussion:
```suggestion
- Test optimisations for climate and mass balance calibration e.g. (:pull:`1933`)
```

fmaussion:
the RMSD test is still the slowest of the test suite! See https://github.com/gampnico/oggm/pull/7 for a fix

gampnico:
One thing I picked up is that checking the mean for melt_f-temp_bias was omitted from the original test. Adding this causes the test to fail with a difference of 12.5 against a tolerance of 10. Same with correlation. @chloe-hancock @fmaussion is this expected, since it's prcp_fac that drives convergence?

The missing combination in the original test [was here](https://github.com/OGGM/oggm/blob/09b7b9b87a72cbc33622cc3701543de72ccbdb93/oggm/tests/test_prepro.py#L1835).

So far I've marked this as xfail for that specific test combo. I lean toward relaxing the tolerance for this.

fmaussion:
Good catch @gampnico - yes, please re-add the test and relax the tolerance to 15. Thanks!

PR #1934, opened 2026-06-27
Replace gdir urls with test_gdirs

fmaussion:
Partly addressed the issues discussed in https://github.com/OGGM/oggm/issues/1906 (RMSD tests still need a fix).

@pat-schmitt I noticed the test suite contains a few tests with `@pytest.mark.skip` - is it worth to unskip these?

### DIFF
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -21,6 +21,9 @@ Enhancements
 Bug fixes
 ~~~~~~~~~
 
+- Fixed test runtime due to unnecessary downloads (:pull:`1934`).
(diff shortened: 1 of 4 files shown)


pat-schmitt:
if we unskip them, this will add again some runtime to the tests. Those tests where written during the development of the dynamic spinup and test for special cases (which I discovered) and error managing within the dynamic spinup. As long as the currently executed tests pass and we do not see big changes in the error_tasks of the preprocessed directories, I think we can leave it as it is. 

However, I think it is good to still keep them as kind of a documentation and if at some point we plan to do a larger restructuring of the dynamic spinup (which is currently not planned).

Issue #1901, opened 2026-04-22
Review tests when possible


fmaussion:
As part of OGGM-Next we promised a streamlining of our tests. It's an extremely tedious a unrewarding task if done in one go.

However, we (the core team) should always be mindful of test duplication when we add functionalities and test them. ideally, when touching tests, take some notes on potential improvements and open an issue. Ideally, streamline tests at the moment you touch them, e.g. by moving to pytest and away from unittest

PR #1936, opened 2026-06-30
refactor(tests): refactor tests in test_prepro

gampnico:
Migrates tests in `test_prepro.py` from unittests to pytest.

Note this does not touch TestClimate or instances of `read_pickle`, as these are addressed in other PRs.

`TestGIS` and `TestInversion` can't really be parametrised without losing readability, e.g. TestInversion doesn't share a lot of logic across tests, so only a few mini-blocks can actually be merged. That means passing parameters as dicts to pytest.mark.parametrize().

Refs https://github.com/OGGM/oggm/issues/1901, ON-86

- [x] Tests added/passed
- [x] Fully documented


fmaussion:
Thanks!

Issue #1947, opened 2026-07-06
[Bug]: OGGM can't detect if cached downloads are truncated


gampnico:
**Overview**

One interesting thing I've run into when working on #1900 is that OGGM doesn't seem to check if a cached file was truncated by an interrupted download. This leads to non-obvious errors because re-running a test or using `reset=True` doesn't refresh the download cache.

**Expected Behaviour**

A cache hit should only return for a file that is verified as complete. If the cached file is incomplete, it should re-download this.

**Actual Behaviour**

`_cached_download_helper` returns the cached path as soon as `os.path.isfile(cache_path)` is true, with no checks. Downloads are written directly to `cache_path`, and cleanup only removes a partial file if the download function itself raises an exception. Rerunning a command does nothing, so the file sticks around until the user clears the cache manually (so never). This affects any download (prepro bundles, DEMs, climate data, RGI files), and is more likely when multiprocessing is enabled and a worker is torn down mid-write.

**Possible Solutions**

1. **Atomic writes**: download to `cache_path + '.tmp'` then `os.replace()` into place once a transfer completes.
2. **Size validation on cache hit**: check the expected size against `Content-Length`, which is read in `_requests_urlretrieve`. I'm still mulling over how this will work with streaming.
3. **Optional integrity check**: an opt-in checksum or `HEAD` size check for stricter verification. Maybe this is fine for smaller files, but checksums for each gdir would just be a waste of compute time.

Maybe a mix of 1 and 2? I'll probably get back to this once I start work on streaming for #1896.

fmaussion:
good one. In the early days of OGGM we mainained a hash database of all teh files OGGM would serve. Users would first download the hash list and teh file_downloader would check hashes. It was a maintainance hell, esp. when the number of files we are servring exploded. this bit has been removed recently: https://github.com/OGGM/oggm/pull/1844

 If size-validation is a possibility that might be good.

Issue #1896, opened 2026-04-15
Develop gdir-compatible GeoZarr data store


gampnico:
Part of OGGM-Next. Opening issue so we can track progress outside of JIRA and get some external engagement.

Points to consider:

- What would a Geozarr version of a gdir look like?
- Can we store a glacier directory and everything OGGM needs within Geozarr?
- How is the data layer discoverable?
- How can attributes be discovered by users (e.g. STAC catalogue)?

More technical points:

- How does this affect existing functions for import/export of existing tars, e.g. `gdirs = workflow.init_glacier_directories(rgi_ids, from_tar=True, delete_tar=True)`
- How does data streaming play a role? Should larger datasets always be streamed on an ad hoc basis, or should we support presets for downloading and caching data (similar to specifying prepro levels)?

gampnico:
See also [this discussion here](https://github.com/radiantearth/stac-spec/discussions/1344#discussioncomment-13279347) for ideas on efficiently structuring data.

fmaussion:
A quick reply on some points:

> What would a Geozarr version of a gdir look like?
> Can we store a glacier directory and everything OGGM needs within Geozarr?

They are interlinked. The ideal scenario would be that we can replace `workflow.init_glacier_directories(rgi_ids, from_tar)` with `workflow.init_glacier_directories(rgi_ids, from_zarr)` with all that `init_glacier_directories` needs is stored in a single or family of geozarr. If this is not possible, then fall back to a hybrid version where minimal data is available as standard files in tar (e.g. the outlines, flowlines, etc.), and the rest via zarr (I think this would be really acceptable).

>  How does this affect existing functions for import/export of existing tars, e.g. gdirs = workflow.init_glacier_directories(rgi_ids, from_tar=True, delete_tar=True)

Is this a backwards compatibility question? The two use cases are completely different, and except for code bloating, I don't see a reason with both couldn't co-exist. However, there is going to be a considerable cost to have code to open both formats.

fmaussion:
A few other thoughts. We need to break this down into smaller pieces if we ever want it to happen. There are also smaller hanging fruits with big returns that can be achieved with smaller steps.

I can think of three tasks that will help to think about the issue in certain ways:
- ~~Get rid of climate data in glacier directories~~ https://github.com/OGGM/oggm/issues/1900
- https://github.com/OGGM/oggm/issues/1902
- Get rid of pickles, one pickle at a time

I need to write up a more comprehensive issue for each of those

[...]

PR #1951, opened 2026-07-07
Feat: incremental glacier directory creation

### LINKED ISSUE #1900: Incremental glacier dir creation
Related to https://github.com/OGGM/oggm/issues/1896

The current glacier directory generation and fetch is very lazy (in the sense that not much thought has been given in optimizing any of it). Each level is built on top of the next, and data from each level is stored together with data from the previous one.

A classic example:
- L2 directories contain gridded data, outlines, and flowlines
- L3 adds climate data and calibration
- L4 adds spinup (in terms of code, this is agnostic of what kind of climate data was added to L3 btw, its the same code).

L2 data is stored again and again and again. Add two options in L3 (e.g. ERA5 and W5E5), and data duplication is multiplied.

Now that I think of it, this is so silly. Or, better said, I knew this was silly, but laziness and simple of use made it worthwhile. The pros of this strategy is that all what users need to describe a setup they use is the `base_url`, and all what oggm needs is a `base_url`. 

I think we need to change this paradigm towards a "shopping list" of "slices" or "levels" that users have to add to their gdirs to have what they want. I can easily envision how that would work. The ideal setup would be a "shopping list of everything" (this is what https://github.com/OGGM/oggm/issues/1896 would have as end goal), but this is a bit scary to think about.

A much easier fruit to pick would be a slightly more layered approach to glacier dir creation:

- L0: outlines
- L1: DEM
- L2: Additional gridded shop datasets
- L3: flowlines (with or without binned data) - which is currently L2
- L4: climate data + calibration (currently L3)
- L5: spinup files (currently L4)

Each level would store *only* the data it created, and a record of what urls / levels it used to create this data (whether this needs to be strict or not depends on the decisions below). 

A user would then say: "I take the L5 spinup files which have been created with ERA5 and COPDEM." 
- *If the separation of files is strict*, the user would then have to also add at least L4 files in order to do a glacier directory in the simplest setup, that would give enough data to do a projection, like the current L5 files. 
- *If we choose to duplicate a few files* (which I am tending towards right now), L4 and L5 would already be enough data to start a run (like the current L5 files)

The user could also say: "I want to add L2 files (preprocessed shop datasets) to my gdir", which is one python line in their code and works regardless of the rest, because this is upstream and independent of the choices you make after. Some internal checks would need to happen because there is a risk that users use L5 files that have been calibrated with, say, COPDEM and then want to add gridded DEMS from another product. This can be checked inside OGGM with the internal record of what has happened.

In terms of changes needed to OGGM, I think this is very doable. The current mechanic for downloading and unpacking tar-ed gdirs would be slightly adjusted to *add* data to a pre-existing gdir instead of overwriting it. In terms of server access, the current "one access per setup" would become something like 2 or 3 accesses, which is very manageable.

As part of this process, we could revisit some old ways OGGM is doing things:
- the very archaic [BASENAMES](https://github.com/OGGM/oggm/blob/76f44230b6ecd2bd3161d1e04ff426e5947a0dc5/oggm/cfg.py#L176-L301) should be replaced by one or more proper yaml or json file, indicating a "data dependency tree" required by OGGM to run (to be refined)
- the [task logging mechanism](https://github.com/OGGM/oggm/blob/76f44230b6ecd2bd3161d1e04ff426e5947a0dc5/oggm/utils/_workflow.py#L508) could be improved to be a more integral part of the workflow.

It's going to be quite hard to make this entirely backwards compatible for old urls, but perhaps not impossible. I think it's a really manageable change which should be discussed internally first (@gampnico @pat-schmitt).

gampnico:
It's finally here!

For an easier-to-read demonstration of the new system, [download and try the notebook](https://github.com/gampnico/oggm-scripts/blob/main/notebooks/incremental_gdirs.ipynb): 

**Features**

- Backwards compatible
- 30% less data
- Append your own files
- Manifests for each glacier directory, tracking OGGM version, user-added files, dataset tags, unique hashed IDs, timestamps and more...
- Manifest peeking so you can check what's in a gdir before loading it into memory
- Wrappers to batch convert existing glacier directories in `utils/compat.py`, with support for multiple base URLs and checks to make sure datasets are compatible. You can convert any region, number or combination of RGI IDs.
- Comes with ~zarr~ npz.

**Overview**

This changes the default glacier directory system from cumulative levels to incremental ones. Most levels on the cluster are now deltas, which contain only the files that are different or added relative to the previous levels. Clients layer levels they need into a single glacier directory, which now contain manifests to track changes and identify datasets.

L3 and L5 datasets are standalone. L3 because there are differences in certain files between L1 and L3 e.g. inversion_flowlines, and L5 because it's a "minimal" preset. I'm calling a combination of levels a "materialisation".

Every prepro level comes with a manifest, and creating a glacier directory now works by layering the levels they need into one glacier directory. Legacy datasets keep working via a fallback that treats them as cumulative materialisations.

**Points for Discussion**

- This is a really big change that includes changes from #1908.
- This is incremental in the sense that users pick from preset deltas available on the cluster, rather than "get this one file". I've decided to do this for multiple reasons:
  - This issue is solved with streaming, because we can use `fsspec` and `xarray` to pick and mix the data we want without downloading it first!
  - Zarr increases the number of files stored in each glacier directory. This means if a user downloads data for the globe, this is potentially millions of small files, which is going to strain the server and makes #1947 impossible to solve.
  - Compatibility checks between datasets become significantly harder when checking multiple single files.
  - Most OGGM workflows require a standard set of files anyway. If we want to give users access to extra files, we could introduce this with a higher preprocessing level (since these are now deltas the storage impact is minimal).

TODO:
- [x] Incorporate @pat-schmitt's changes after dev is merged to master
- [ ] Replace the test base URL with an actual URL once data is available.
- [x] Decide on terminology for documentation. So far I'm calling merged deltas or standalone prepros in a single gdir "materialisations" (e.g. L3, L5). but I think this is a bit awkward.
- [x] Maybe add manifest schema version to dataset_id hash to prevent future manifests clashing?

Closes #1900 
Refs: #1896 

- [x] Tests added/passed
- [x] Fully documented


hilljessica8787-dot (approved):

gampnico:
Review notes from today's meeting @fmaussion, @pat-schmitt.

- It should be possible to build different deltas from the same parent, i.e. deltas should be "parent-agnostic".
- Add a new `kind="materialisation"` to schema for clarity, especially since materialisations are going to be a very niche case.
- Deltas should use chained downloads: e.g. read L2 manifest -> download L1 -> read L1 manifest -> download L0.
- Need to review nomenclature both generally and in schema e.g. "artifact", "dataset" etc. This must be documented.
- Conversion scripts in `compat` are more useful for development or a small handful of users.
- More discussion needed on `dataset_tag` vs `base_url`: some compatibility checks can be relaxed since deltas should be parent-agnostic. My personal concern is having some kind of identifier for reproducibility, and it's a faster way of checking compatibility (when needed). With v0 this is handled with `base_url` and `log.txt`, but in practice a user can have two directories with the same base URL but different files.

@fmaussion your suggestions for expanding scope:
- Maybe have a watcher or some similar mechanism to see what exactly has changed between two glacier directories (or rather `log.txt`)?
- Maybe add a check to see if a `settings.yml` has changed? (Note this will already be marked in the manifest under `files.updated`, so maybe which specific parameters, e.g. via a diff).

fmaussion:
thanks for the meeting @gampnico ! Quick notes:

> It should be possible to build different deltas from the same parent, i.e. deltas should be "parent-agnostic".

Mh, more the other way around? Artifacts should be "child-agnostic". Some deltas need a very specific parent, while some deltas (briefly mentioned this morning) should be parent-fixed and others might be "parent-flexible".

> Add a new kind="materialisation" to schema for clarity, especially since materialisations are going to be a very niche case.

To me, materialisations aren't niche but the _main_ case from the user perspective: all of the gdirs they will build locally are materialisations. A materialisations hence needs to know all the parents it was built from, which parameter settings they used, which tasks were run, etc. not only incremental anymore.

> Deltas should use chained downloads: e.g. read L2 manifest -> download L1 -> read L1 manifest -> download L0.

yes

> Need to review nomenclature both generally and in schema e.g. "artifact", "dataset" etc. This must be documented.

I grew quite accustomed to "artifact" today!

> More discussion needed on dataset_tag vs base_url: some compatibility checks can be relaxed since deltas should be parent-agnostic. My personal concern is having some kind of identifier for reproducibility, and it's a faster way of checking compatibility (when needed). With v0 this is handled with base_url and log.txt, but in practice a user can have two directories with the same base URL but different files.

Agreed that some id would be good. `parent_base_url` is a "must" I think, together with perhaps some hash version of params.yaml + base_url or something for the id. If the id check fails, the params can be checked. TBD

PR #1960, opened 2026-07-17
Audit and fix test suite

gampnico:
This also contains test fixes relevant to @pat-schmitt's DMB changes.

Fixes:
  - `test_mb_daily_model` wasn't actually testing `mb_annual_oggm_v16` because it was in the `do_plot` block and set to False.
  - `rtol` on smoothed segments in `test_shapefile_output` are now positive assertions and tightened. `rtol=2` was always going to pass.
  - Bare pytest skips now have reasons.
  - `test_geom_width` didn't have any assertions.
  - `test_compile_calving` also didn't have any assertions.
  - measured max rel diffs to `test_process_monthly_ismip_data`.
  - inverted assertions in `test_fl_diag_quantiles` are now direct comparisons of computed thresholds.   

Removes:
  - Duplicate `test_mb_calibration_to_rmsd`, kept newly refactored version.
  - Some redundant assert gdirs, since the following assertions will fail if it's an empty list

Refactors:
  - `test_sfc_type_mb_model` had a setup block wrapped in `pytest.raises`, so the test could pass if the setup raised an InvalidWorkflowError. Removed unnecessary trailing regex.
  
Discussion:

- `do_plot` in `test_models.py` is set to `False`. @pat-schmitt is this intentional? I could move these tests to `test_graphics.py` and compare them to baseline images.
- I added some assertions to `test_geom_width`. I do these per flowline, and just assume they should roughly sum to the glacier area. I also figured NaNs were OK, since these are removed in `catchment_width_correction`.
- What are we supposed to test in `test_compile_calving`? Since it's a copy of hef_gdir and doesn't seem to calve, I added tests to check it's finite, non-negative, and not decreasing.

- [x] Tests added/passed
- [x] Fully documented

### DIFF
--- a/oggm/tests/test_models.py
+++ b/oggm/tests/test_models.py
@@ -1368,6 +1368,38 @@ class TestMassBalanceModels:
         np.testing.assert_allclose(ds_daily.volume[-1], ds_monthly.volume[-1],
                                    atol=3e7)
 
+    def test_temp_melt_not_persisted_by_ctor(self, hef_gdir):
+        """Constructing model with explicit default temp_melt shouldn't
+        write to gdir settings file, and calib params check should catch
+        the mismatch.
+        """
+
+        tasks.process_gswp3_w5e5_data(hef_gdir, daily=True)
+        gdir = hef_gdir
+
+        stored_before = gdir.get_stored_settings()
+        assert "temp_melt" not in stored_before
+
+        # DailyTIModel defaults temp_melt=0.0 while gdir calibrated with
+        # monthly model (temp_melt=-1)
+        with pytest.raises(InvalidWorkflowError, match="temp_melt"):
+            massbalance.DailyTIModel(gdir)
+
+        # explicit kwarg mismatching the calibration also raises
+        with pytest.raises(InvalidWorkflowError, match="temp_melt"):
+            massbalance.MonthlyTIModel(gdir, temp_melt=5.0)
+
+        # opting out of the check works, uses value for instance
+        mb_mod = massbalance.DailyTIModel(gdir, check_calib_params=False)
+        assert mb_mod.temp_melt == 0.0
+
+        # and in no case is anything written to settings file
+        assert gdir.get_stored_settings() == stored_before
+
+        # monthly model still fine
+        mb_mod = massbalance.MonthlyTIModel(gdir)
+        assert mb_mod.temp_melt == gdir.settings["temp_melt"]
+
     @pytest.mark.slow
     def test_sfc_type_mb_model(self, hef_gdir):
 
--- a/oggm/tests/test_models.py
+++ b/oggm/tests/test_models.py
@@ -1031,36 +1031,35 @@ class TestMassBalanceModels:
         np.testing.assert_allclose(prcp_d, prcp_m)
         np.testing.assert_allclose(t_d, t_m, atol=0.02)
 
+        # for checking how much has changed to OGGM v1.6 here a hard coded
+        mb_annual_oggm_v16 = [
+            -1.42297448e-07, -1.41160282e-07, -1.39199219e-07, -1.36113197e-07,
+            -1.31840437e-07, -1.26940610e-07, -1.21949801e-07, -1.17948533e-07,
+            -1.14388734e-07, -1.11166865e-07, -1.08178550e-07, -1.05308084e-07,
+            -1.02515762e-07, -9.96907271e-08, -9.67172896e-08, -9.36011930e-08,
+            -9.03592687e-08, -8.69923123e-08, -8.29093380e-08, -7.90813507e-08,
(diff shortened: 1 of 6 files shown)


pat-schmitt:
> - test_mb_daily_model wasn't actually testing mb_annual_oggm_v16 because it was in the do_plot block and set to False.

I used mb_annual_oggm_v16 more for testing for myself, to make sure the adaptations introduced in dev do not change the monthly mass balance model much. At some point we can get rid of this again (maybe with the new release...)

> - do_plot in test_models.py is set to False. @pat-schmitt is this intentional? I could move these tests to test_graphics.py and compare them to baseline images.

This is intentional and mainly for developing (i copied this style from existing oggm code). I think the reason those are not checked regularly is that keeping baseline images up to date involves quite some work, and we only want to do this for images which we think are very important in showing nothing crazy is happening in the model.

oggm/tests/test_models.py
```diff
@@ -1032,36 +1032,35 @@ def test_daily_mb_model(self, hef_gdir):
...
+            6.39301260e-09,  1.01207768e-08,  1.35185100e-08,  1.59884941e-08,
+            1.70314225e-08,  2.04225150e-08,  2.35563242e-08,  2.53007150e-08,
+            2.65123374e-08,  2.91862845e-08,  3.14872294e-08,  3.31400347e-08,
+            3.39361834e-08,  3.73918160e-08,  3.92227052e-08,  4.10314532e-08,
+            4.13874876e-08,  4.39342346e-08,  4.43393872e-08,  4.58317424e-08,
+            4.86526460e-08,  4.93663560e-08,  5.36643971e-08,  5.41680453e-08,
+            5.84341486e-08,  6.00092412e-08,  6.30844863e-08,  6.58271495e-08,
+            6.76431998e-08,  7.15019273e-08,  7.21872267e-08,  7.61114775e-08,
+            7.71056592e-08,  7.76305573e-08,  8.12629938e-08,  8.45592440e-08,
+            8.70176740e-08,  8.83996244e-08]
```
fmaussion, line 1059: This is OK (there aren't many of such regression tests in OGGM, maybe we need more) but htis is a bit ugly - can we put them somewhere else?

oggm/tests/test_models.py
```diff
@@ -1098,35 +1097,36 @@ def test_daily_mb_model(self, hef_gdir):
...
+            -2.46389327e-07, -2.34266499e-07, -2.21147635e-07, -2.11610869e-07,
+            -2.07584057e-07, -1.94490833e-07, -1.82390996e-07, -1.75655792e-07,
+            -1.70977641e-07, -1.60653361e-07, -1.51769266e-07, -1.45387680e-07,
+            -1.42313699e-07, -1.28971282e-07, -1.21902102e-07, -1.10290002e-07,
+            -1.07780730e-07, -8.98316793e-08, -8.69762307e-08, -7.64583579e-08,
+            -6.02983722e-08, -5.64291333e-08, -3.31281440e-08, -3.03977137e-08,
+            -8.33160424e-09, -1.40054651e-09,  1.21318019e-08,  2.42006525e-08,
+            3.21920240e-08,  4.91720184e-08,  5.21876188e-08,  6.94559431e-08,
+            7.11095569e-08,  7.11095569e-08,  7.11095569e-08,  7.11095569e-08,
+            7.11095569e-08,  7.11095569e-08]
```
fmaussion, line 1124: same here

oggm/tests/test_prepro.py
```diff
@@ -1925,347 +1940,6 @@ def reset_observation_file(gdir):
         np.testing.assert_allclose(mb_calib_2d['melt_f'],
                                    mb_calib_1d['melt_f'], rtol=0.05)
 
-    @pytest.mark.slow
```
fmaussion, line 1928: omg I think my AI PR left this in??? Terrible.

I assumed you checked that the three tests that replaced this all make sense?

fmaussion (approved):
All good from me. the hardcoded values are a bit ugly but it;s ok

oggm/tests/test_models.py
gampnico: I think this was introduced by @pat-schmitt? I don't know if it's still necessary. I'd rather have this as a small file if needed.

fmaussion: yeah perhaps one json file as part of oggm-sample-data but it's a lot of hoops

pat-schmitt: those are actually not needed anymore and can be deleted. I used them to check that my changes in dev do not change the MonthlyTIModel too much, and all looked good. Maybe before deleting check one more time with the current version after merging

gampnico: OK, thanks for letting me know. @fmaussion shall I remove this in a separate patch?

fmaussion: yes thanks

PR #1961, opened 2026-07-17
fix: gdir.settings not correctly read

gampnico:
This patch provides two fixes:

1) The init constructor for MonthlyTIModel was writing `temp_melt` to `gdir.settings` in `mb_global_params`. When initialising `DailyTIModel`, the `check_calib_params` was running before this was written to `settings.yml`, which meant `temp_melt` was permanently set to zero, or ignored if `check_calib_params=False`.
I've removed the write and switched the order so the check occurs afterwards.

2) The `SemiImplicitModel` was silently ignoring the model settings and reading from cfg.PARAMS. It also allowed calving to occur for land-terminating glaciers. I've added a guardrail: `self.do_calving = do_calving and self.is_tidewater`

**Point for discussion**

- There was no check for `is_tidewater` before setting `do_calving` to True in the `SemiImplicitModel`, which meant in theory a land-terminating glacier could calve. The check does exist in `FluxBasedModel`. In both models, there is no check for `is_lake_terminating` which could also allow calving. Is this behaviour intentional? If so I will also need to modify `TestKCalving::test_water_level` before merging.

- [x] Tests added/passed
- [x] Fully documented

### DIFF
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -152,6 +152,10 @@ Bug fixes
 - Fixed a bug in ``compile_to_netcdf`` decorator, avoiding to raise an error if
   a single chunk failes (:pull:`1954`).
   By Copilot and `Patrick Schmitt <https://github.com/pat-schmitt>`_
+- Model constructors no longer silently persist a non-default ``temp_melt`` to
+  the gdir settings file. ``check_calib_params`` now validates the effective
+  model parameters instead of the settings file, and calibration tasks record ``mb_global_params`` from the model used.
+  By `Nicolas Gampierakis <https://github.com/gampnico>`_
 
 Breaking changes
 ~~~~~~~~~~~~~~~~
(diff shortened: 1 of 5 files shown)


pat-schmitt:
For the daily model `temp_melt` should usually be zero. But it is True that it should not silently be set to zero, but maybe at least a user warning should be shown...

pat-schmitt:
This looks all good to me, and I assume with the updated sample data all test should also pass now

fmaussion:
Thanks @gampnico - I'm leaving the whats-new entry for now, but we might have to reconsider some what's new entries later as this is all fixes to the big dev PR, i.e. the what's new is irrelevant for users coming in ("we added a bug and fixed it" kind of message).

pat-schmitt:
This is true, also some of my whats-new entries are related to 'dev-fixes'

PR #1963, opened 2026-07-17
fix: melt_f recovery in dynamic spinup now retries instead of falling…

gampnico:
… back

`get_mismatch` had an incorrectly parsed condition, so the function wouldn't retry with a new `melt_f` limit.

This fixes the loop logic and post-loop check and adds a regression test.

WIP, might stick another `melt_f` fix in here.

- [x] Tests added/passed
- [x] Fully documented


pat-schmitt:
Do you have an example where this is causing a problem? You also could check if `test_run_dynamic_melt_f_calibration_with_dynamic_spinup_special_cases` is passing. This should include special cases like parameter limits, but it is skipped by default due to very long run times. (At the moment i have not time to check if your special case is included)

PR #1967, opened 2026-07-29
Fix multiprocessing OOM with new settings

pat-schmitt:
Large `oggm_prepro` runs started hitting OOM errors on the cluster that didn't happen before. This could be traced back to the pickling logic of `GlacierDirectory` during multiprocessing. As a result  `settings`/`observations` gets serialized every time a gdir is sent to a worker process via `workflow.execute_entity_task()`'.

The fix is to add `__getstate__`/`__setstate__` to `GlacierDirectory`:                                                                                                                                                                                                                                                                                                                                                                                           
  - `__getstate__` drops `settings`/`observations` from the pickled state.                                                                                                                                                                                                                 
  - `__setstate__` rebuilds them via the existing `_get_settings_class()`/`_get_observations_class()` helpers, which read fresh from disk in the worker process.           

With this changes I was able to run the prepro levels for region 14 (around 37,000 glaciers) on the cluster again.

- [x] Tests added/passed
- [ ] Fully documented

### DIFF
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -182,7 +182,7 @@ Bug fixes
   (e.g. ``oggm_prepro`` on tens of thousands of glaciers): ``GlacierDirectory``
   no longer re-serializes its ``settings``/``observations`` on every task
   dispatched to a worker process. They are now dropped before pickling and
-  rebuilt from disk in the worker instead (:pull:`XXXX`).
+  rebuilt from disk in the worker instead (:pull:`1967`).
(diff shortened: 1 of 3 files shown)


fmaussion:
Are the failing tests related?

pat-schmitt:
The problem was that for one new test uses a very high `prcp_fac` and does not reset at the end. And as many tests use the same hef_gdir the failing tests used the very high `prcp_fac`, which caused them to fail. Should be fine now.

PR #1972, opened 2026-08-05
fix calibrate_inversion_from_consensus

### LINKED ISSUE #1962: [Bug]: `calibrate_inversion_from_consensus` doesn't calibrate against the consensus
## Overview

`calibrate_inversion_from_consensus` defaults to `overwrite_observations=False`. `calibrate_inversion_from_ref_table` therefore skips loading the reference table and instead reads a value from `gdir.observations` that may not exist or may not match the intended Farinotti reference.

## Expected Behaviour

Calling `calibrate_inversion_from_consensus` is deprecated, and should calibrate Glen A against the Farinotti 2019 consensus reference table (`ref_table='consensus'`), matching the docstring's stated default of `overwrite_observations=True`. The `else` branch in `calibrate_inversion_from_ref_table` should behave consistently.

## Actual Behaviour

`calibrate_inversion_from_consensus` passes `overwrite_observations=False` by default, so `calibrate_inversion_from_ref_table` never loads the reference table. Instead, the `else` branch ignores `ref_table='consensus'` entirely. This means:
- Fresh gdirs without `ref_volume_m3` stored in `observations.yml` raise a `KeyError` before any calibration happens.
- Gdirs with a stored `ref_volume_m3`** silently calibrate Glen A against their stored value instead of the Farinotti 2019 consensus, contradicting the deprecation warning of unchanged behaviour.

`test_calibrate_inversion_from_consensus` doesn't catch the bug because earlier calls in the test already provide the observation and it only tests for a FutureWarning

## Possible Solutions

1. Change the wrapper's default to `overwrite_observations=True`, matching the docstring.
2. Alternatively, fix `calibrate_inversion_from_ref_table`'s `else` branch to load the reference table when `ref_volume_m3` is missing from `gdir.observations`.
3. Add a regression test that calls the wrapper on a `gdir` without a pre-populated `ref_volume_m3` observation, to catch the `KeyError` directly.

Since this is deprecated, I may class this as lower priority.

pat-schmitt:
Now `calibrate_inversion_from_consensus` always overwrites potential existing observations, to be sure to always use the consensus estimates when calling this function. If you want to use other datasets you should use `calibrate_inversion_from_ref_table`. 

Closes https://github.com/OGGM/oggm/issues/1962

- [x] Tests added/passed
- [ ] Fully documented

### DIFF
--- a/oggm/tests/test_workflow.py
+++ b/oggm/tests/test_workflow.py
@@ -276,10 +276,38 @@ class TestFullRun(unittest.TestCase):
         assert (gdir_0_ref_vol['value'] !=
                 gdir_0_ref_vol_user_provided['value'])
 
-        # the deprecated alias still works and warns
+        # the deprecated alias still works, warns, and always calibrates
+        # against the correct (consensus) ref table - regardless of what
+        # was stored in the observations file before (the user provided
+        # volume from the previous calibration call above)
         with pytest.warns(FutureWarning, match='deprecated'):
-            workflow.calibrate_inversion_from_consensus(gdirs,
-                                                        ignore_missing=True)
+            df_consensus = workflow.calibrate_inversion_from_consensus(
+                gdirs, ignore_missing=True)
+        df_consensus = df_consensus.dropna()
+
+        # same ref table as a direct call with ref_table='consensus'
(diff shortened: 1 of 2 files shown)


fmaussion:
How come wasn't this discovered by a test? Might be good to add one if easy

fmaussion (approved):

PR #1973, opened 2026-08-05
Automate the temperature bias file creation

fmaussion:
The temperature bias prior file used by the `informed_threestep` calibration
was, until now, produced by copy-pasting an exploratory notebook
(`prepare_bias_map.ipynb` and its many descendants) for each new setup. The
notebook had long lost its exploratory purpose: out of ~50 cells, only the csv
file and one map were ever used. This replaces it with a command.

### New workflow

```
1. the temp_melt preprocessing (one job per region)
oggm_prepro --temp-bias-run <all your usual options> --rgi-reg 01

2. summarize all the regions into the bias file (+ diagnostic plots)
oggm_temp_bias --input RGI62/b_080/L3/summary --output-file temp_bias_v2026.1.csv

3. copy it somewhere reachable, then use it for the real run
oggm_prepro --temp-bias-file-path <path or url> <your options>
```

### What's in here

- **`utils.compute_temp_bias_dataframe()`** does the work: drops glaciers whose
  calibration failed, fills missing MB errors with a quantile of the
  distribution, computes the three medians (plain / area-weighted /
  MB-error-weighted) per climate grid point, and grows a square search radius
  (with longitude wrap-around) for grid points holding fewer than
  `min_glaciers` glaciers. Output columns are identical to the current
  reference files. All the notebook's magic numbers are kwargs, defaulting to
  the notebook values.

- **The climate netCDF is no longer needed.** `glacier_statistics_*.csv`
  already carries `baseline_climate_ref_pix_lon` / `_lat` (via
  `get_climate_info()`), which is *exactly* the coordinate
  `mb_calibration_from_geodetic_mb` matches against, so the grid is
  reconstructed from the statistics file alone. This is what makes the command
  work with an arbitrary `--custom-climate-task` without the user having to
  know where the forcing file lives. Side effect: `lon_id`/`lat_id` (and hence
  `unique_id`) are now relative to the extent of the data instead of absolute
  indices into the climate file. Nothing reads them — the calibration matches
  on `lon_val`/`lat_val` and selects positionally — they are only used
  internally for the grouping and the map.

- **`oggm_temp_bias`** (new console script) wraps it, and writes `<stem>_map.png`
  and `<stem>_hist.png` next to the csv plus a summary in the log (glaciers
  used/dropped, inferred grid spacing, grid points per search radius, bias
  quantiles). Plots are plain matplotlib — no cartopy, which is an optional
  dependency.

- **`oggm_prepro --temp-bias-run`** is a preset for step 1: it forces the
  `temp_melt` strategy, `max_level=3` and `skip_inversion` (the inversion has
  no effect on the MB calibration — this is most of the runtime), writes no
  glacier directory tar files (they are thrown away anyway), and computes the
  bias file for that region at the end of L3. It logs everything it overrides.

- **`utils.weighted_quantile_1d()`**, the notebook's `quantile_1D`, promoted to
  a public utility next to `weighted_average_1d`.

### Note on the region split

The neighbour grouping crosses RGI region borders and wraps in longitude, so it
is only correct when applied to all regions at once. `--temp-bias-run` therefore
emits a warning next to its per-region file: for anything but a single-region
setup, `oggm_temp_bias` should be re-run over all the regions' statistics. That
is why the bias computation is its own command rather than only a prepro option.

### Tests

- `TestFuncs.test_weighted_quantile_1d`
- `TestTempBiasCLI`: arg parsing, file format and grouping invariants against a
  synthetic statistics file, dateline wrap-around, and the CLI writing csv +
  plots and round-tripping through `get_temp_bias_dataframe` and the
  nearest-pixel lookup.
- `TestTempBiasCLI.test_prepro_temp_bias_run` (slow) is the end-to-end one: a
  `--temp-bias-run` prepro on 3 Oetztal glaciers produces the file with no tars
  written, and feeding that file back in as `--temp-bias-file-path` for an
  `informed_threestep` run reproduces the prior to 1e-3.

- [x] Tests added/passed
- [x] Fully documented

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -64,6 +64,7 @@ Input/Output
     utils.gdir_to_tar
     utils.base_dir_to_tar
     utils.cook_rgidf
+    utils.compute_temp_bias_dataframe
     global_tasks.write_centerlines_to_shape
     global_tasks.compile_glacier_statistics
     global_tasks.compile_run_output
@@ -227,6 +228,7 @@ These commands are available:
 
 - ``oggm_netrc_credentials``
 - ``oggm_prepro``
+- ``oggm_temp_bias``
 - ``oggm_benchmark``
 
 .. autosummary::
@@ -234,6 +236,7 @@ These commands are available:
     :nosignatures:
 
     cli.prepro_levels.run_prepro_levels
+    cli.temp_bias.run_temp_bias
 
 Classes
 =======
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -61,7 +61,8 @@ Enhancements
   statistics themselves, so this works with any (custom) climate dataset.
   ``oggm_prepro`` also gets a ``--temp-bias-run`` preset for the preprocessing
   step itself: it forces the `temp_melt` strategy, stops at level 3, skips the
-  ice thickness inversion and does not write the glacier directory tar files.
+  ice thickness inversion and writes nothing but the level 3 glacier statistics
+  file, which is the input of ``oggm_temp_bias``.
   New utility function ``utils.weighted_quantile_1d``.
   By `Fabien Maussion <https://github.com/fmaussion>`_
 - Test durations are now visible in Actions logs (:pull:`1920`).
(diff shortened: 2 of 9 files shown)


fmaussion:
Validated against the v2025.6 W5E5/RGI6 file from the same run — 4254/4641 grid points identical, area-weighted mean difference −0.0003 K; the residual comes from 0.1% of glaciers sitting exactly on a cell boundary, where the old notebook's digitize and OGGM's argmin break the tie in opposite directions.

PR #1979, opened 2026-08-09
fix gdir size with new params cross-ref

fmaussion:
Don't pickle the cfg.PARAMS snapshot with every flowline

Flowline.settings carried a shallow copy of cfg.PARAMS, which holds
intersects_gdf - a region-wide table. It was serialized into
model_flowlines.pkl once per glacier, up to five times per glacier
across levels 3 to 5, making L5 directories of dense regions such as
RGI60-19 up to 11x bigger than in v1.6.3.

Flowline, FlowlineModel and ModelSettings now drop that snapshot before
pickling and rebuild it from cfg.PARAMS on unpickling, as
GlacierDirectory already does since #1967

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,5 @@ docs/_images
 docs/.doctrees
 oggm/version.py
 oggm/ignore
+*_project*
+CLAUDE.md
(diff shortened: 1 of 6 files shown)


fmaussion:
I note that https://github.com/OGGM/oggm/pull/1908 would have probably prevented this from happening - probably worth prioritising this?

Issue #1978, opened 2026-08-09
Rethink `cfg.DATA` as a generic multiprocessing cache


fmaussion:
## Summary

`cfg.DATA` currently mixes two responsibilities:

1. a generic runtime cache for lazily loaded OGGM data
2. a container intended to support multiprocessing reuse of cached state

This was pragmatic historically, but it now creates unclear semantics and likely avoidable overhead.

I think OGGM should stop using `cfg.DATA` as a generic shared cache for heavy read-only objects and instead move toward typed, per-process lazy caches plus narrowly scoped shared state where needed.

## Background

`cfg.DATA` seems to have started as a general replacement for special-purpose globals such as demo metadata. Later, it was extended to support multiprocessing reuse for cached reference datasets.

Today it stores a mix of objects, including:

- pandas DataFrames
- pandas Series
- lists and sets
- nested dicts
- DEM grid objects

That makes it a broad mutable global with mixed semantics.

## Why this is still questionable on Linux/fork

Even on Linux, where major OGGM runs typically use `fork`, the current pattern is not obviously the right one for heavy read-only cached data like geodetic MB tables.

There are three different designs to compare:

1. preload the DataFrame in the parent, then fork workers and let them inherit it read-only via copy-on-write
2. let each worker maintain its own local lazy cache
3. store the DataFrame in a manager-backed shared dict such as `cfg.DATA`

For large read-only tables, option 3 is often the weakest of the three, even under fork:

- pre-fork loading avoids repeated file reads and avoids proxy overhead on each access
- per-worker local caching avoids proxy overhead and keeps semantics simple
- manager-backed caching adds IPC and pickle/unpickle overhead to every top-level read

So the main concern here is not just spawn. It is that a manager-backed global cache is a weak fit for large read-only scientific tables even in the fork case.

## Secondary note on spawn

This issue is not Mac-specific. However, the same design looks even less attractive under spawn-based multiprocessing, where worker startup and state reconstruction are already more expensive.

### 1. Manager/proxy-backed access is expensive for large cached objects

If `cfg.DATA` is manager-backed, each top-level read goes through proxy IPC plus pickle/unpickle of the stored object.

That is a poor fit for cached payloads like:

- geodetic mass-balance DataFrames
- temperature-bias DataFrames
- DEM grid bundles

A quick local benchmark on macOS gave roughly:

- manager dict read of a large DataFrame: about 5 ms per access
- manager dict read of a medium list: about 0.9 ms per access
[...]

PR #1980, opened 2026-08-10
Remove intersects from params

fmaussion:
PARAMS stored a copy of the large intersects dataframe - this was not a good design (see also: https://github.com/OGGM/oggm/issues/1978 )

### DIFF
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -219,6 +219,14 @@ Bug fixes
 Breaking changes
 ~~~~~~~~~~~~~~~~
 
+- The glacier intersects are no longer stored in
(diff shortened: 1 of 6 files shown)


fmaussion:
I think that a better fix, but more backwards incompatible and breaking, would be to treat intersects as an RGI file: it's an input, no need to have it as global always.

I'll do a simpler fix now (not read the file at each cfg.initialise()), but longer term this is still bad

fmaussion:
correcting myself: it's already lazy. Conceptually the global is still wrong, but changing cfg.set_intersects_db is simple enough but it's a public API change on init_glacier_directories plus tutorial changes, which I'm not ready to do right now.

PR #1984, opened 2026-08-10
Attempt stricter spinup starts

fmaussion:
@pat-schmitt to discuss when possible


pat-schmitt:
Looks good to me after a quick look. Only a comment regarding performance: Why not trying the shorter periods before the longer periods, like it was before? By default two shorter periods are tried vs. around ten longer periods. But to see what is faster we need to test which variant is computationally faster and use this as the default (maybe for a few regions). Or check the diagnostics after a run to see the distribution of used spinup periods. But not urgent at the moment

fmaussion:
> Why not trying the shorter periods before the longer periods, like it was before?

The main motivation is to "force" a start year for e.g. Glambie, and make sure that all spinups are either from start year or before, unless absolutely necessary (or not at all, if you set `allow_shorter_spinup`). I'm not entirely clear about what it means to be honest, or what it means for glambie when we have a mix of dyn spin and fixed geometry data - I'll investigate

fmaussion:
@copilot resolve the merge conflicts in this pull request

fmaussion:
I think this should now be closer to the previous behavior (testing shorter year first) except when `allow_shorter_spinup` is false.

PR #1986, opened 2026-08-11
fix: OGGM CLI arguments no longer accept arbitrary strings

gampnico:
A bunch of arguments e.g. `--temp-bias-run` were declared in a way that would allow an optional value that is never validated. So for example if someone passed something like: `--temp-bias-run whatever` this would evaluate to True and force a temp bias run. More seriously, if someone tried passing `--elev-bands False` this would evaluate to True.

`--store-monthly-hydro` was set to `const=True` and `default=True`, so it evaluated to `True` whether it was passed or omitted. This meant it could not be switched off from the command line. I personally think this should default to False, but in the meantime I've added an optional `--no-store-monthly-hydro`. @fmaussion this is quite a dirty fix from me and I'd appreciate your input here on what the default should be.

I also added type validation to `--dynamic-spinup-periods-to-try` so that it works.

## Discussion

This breaks any external script which passes a value, e.g. `--elev-bands True`. This will now raise an error, which I think is worth it as the same scripts passing e.g. `--elev-bands False` were silently running with the option set to True, and there was nothing in the output to say so.

- [ ] Tests added/passed
- [x] Fully documented
- [x]  Entry in whats-new.rst

### DIFF
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -180,12 +180,13 @@ Enhancements
 - New ``store_hydro_output`` kwarg in ``run_prepro_levels`` (and
   ``--store-hydro-output`` CLI flag) to also compute and store hydrological
   model output during preprocessing, via ``run_with_hydro``. The accompanying
-  ``store_monthly_hydro`` kwarg (and ``--store-monthly-hydro`` CLI flag,
-  default ``True``) additionally stores this hydrological output at monthly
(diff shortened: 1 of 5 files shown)


fmaussion:
thanks - I changed default to False to match default OGGM. Ic couldn't find the --no-store-monthly-hydro you were talking about?

This will clash with https://github.com/OGGM/oggm/pull/1984 - I'll see tomorrow waht to merge first

PR #1989, opened 2026-08-13
Fix numerical instability in trapezoid origin widths

Ruitangtang:
Related to (Fix) #1839 

## Summary

This PR fixes a floating-point instability in the initialization of trapezoidal beds.

For a sloping trapezoid, the section must satisfy:

section > lambda * thickness**2 / 2

When the section lies exactly at this physical boundary, floating-point cancellation can cause the calculated origin width to become zero or slightly negative. This triggers:

ValueError: Trapezoid beds need to have origin widths > 0.

The problem was reproduced for RGI60-07.00026 with recent NumPy versions, including NumPy 2.x.

## Changes
- Compute the trapezoid origin width using a numerically consistent expression.
- Detect positive-thickness sloping trapezoids at or slightly below their physical boundary.
- Reject sections that are materially below the physical minimum.
- Advance boundary sections to the next representable floating-point value using np.nextafter.
- Preserve zero-thickness downstream points and strict origin-width validation.
- Avoid using a fixed absolute epsilon or silently clipping invalid widths.

## Validation

For RGI60-07.00026, the corrected initialization produced:

minimum origin width: 1.2068248134370502e-14
invalid origin widths: none

The glacier initialization and subsequent model run completed without the trapezoid-width error.

The selected existing OGGM tests initially produced:

295 passed, 134 skipped, 1 xfailed, 1 failed

The sole failure occurred in an unrelated GIS test while deleting a nonempty temporary directory during teardown (`TestGIS::test_gridded_data_var_to_geotiff - OSError: [Errno 39] Directory not empty: 'RGI50-11.00897'`). That test passed when rerun independently.

PR #1995, opened 2026-08-20
revert previous formula in width compute

fmaussion:
not worth the change


fmaussion:
for context, this reverts an unnecessary change in https://github.com/OGGM/oggm/pull/1989

